PR #8, opened 2025-05-09
feat: explicit reverser for reversible accounts

dastansam:


pallets/reversible-transfers/src/lib.rs
```diff
@@ -398,8 +425,18 @@ pub mod pallet {
...
+            let ReversibleAccountData {
+                delay,
+                explicit_reverser,
+                policy: _,
+            } = Self::reversible_accounts(&who).ok_or(Error::<T>::AccountNotReversible)?;
+
+            match explicit_reverser {
+                Some(reverser) => {
+                    ensure!(who != reverser, Error::<T>::InvalidReverser);
+                }
```
n13, line 437: There is an error type for this called ExplicitReverserCanNotBeSelf - should probably be that error. 

Also ExplicitReverserCanNotBeSelf isn't used anywhere otherwise ;)

illuzen: I personally would have made the reverser not optional and just have the client set it as sender, but this works.

illuzen: It would reduce the number of conditionals in the function, tho, which makes it easier to audit.

PR #29, opened 2025-05-26
fix: make benchmarks compile

dastansam:
we can close the #28 @czareko

### DIFF
--- a/pallets/merkle-airdrop/src/benchmarking.rs
+++ b/pallets/merkle-airdrop/src/benchmarking.rs
@@ -8,7 +8,6 @@ use frame_benchmarking::v2::*;
 use frame_system::RawOrigin;
 extern crate alloc;
 use alloc::vec;
-use frame_support::traits::fungible::{Inspect, Mutate};
 use frame_support::BoundedVec;
 use sp_runtime::traits::Saturating;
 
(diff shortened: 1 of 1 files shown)


n13:
Getting these warnings no idea if they are related though. They might be in main. 

```
warning: non-local `impl` definition, `impl` blocks should be written at the same level as their item
   --> runtime/src/apis.rs:292:4
    |
283 | /         fn dispatch_benchmark(
284 | |             config: frame_benchmarking::BenchmarkConfig
285 | |         ) -> Result<Vec<frame_benchmarking::BenchmarkBatch>, alloc::string::String> {
    | |___________________________________________________________________________________- move the `impl` block outside of this associated function `dispatch_benchmark`
...
292 |               impl frame_system_benchmarking::Config for Runtime {}
    |               ^^^^^---------------------------------^^^^^-------
    |                    |                                     |
    |                    |                                     `Runtime` is not local
    |                    `Config` is not local
    |
    = note: an `impl` is never scoped, even when it is nested inside an item, as it may impact type checking outside of that item, which can be the case if neither the trait or the self type are at the same nesting level as the `impl`
    = note: `#[warn(non_local_definitions)]` on by default

warning: non-local `impl` definition, `impl` blocks should be written at the same level as their item
   --> runtime/src/apis.rs:293:4
    |
283 | /         fn dispatch_benchmark(
284 | |             config: frame_benchmarking::BenchmarkConfig
285 | |         ) -> Result<Vec<frame_benchmarking::BenchmarkBatch>, alloc::string::String> {
    | |___________________________________________________________________________________- move the `impl` block outside of this associated function `dispatch_benchmark`
...
293 |               impl baseline::Config for Runtime {}
    |               ^^^^^----------------^^^^^-------
    |                    |                    |
    |                    |                    `Runtime` is not local
    |                    `Config` is not local
    |
    = note: an `impl` is never scoped, even when it is nested inside an item, as it may impact type checking outside of that item, which can be the case if neither the trait or the self type are at the same nesting level as the `impl`
```

PR #48, opened 2025-06-04
add mining guide

n13:


MINING.md
```diff
@@ -0,0 +1,277 @@
...
+```bash
+quantus-node \
+  --node-key-file ~/.quantus/node_key.p2p \
+  --rewards-address ~/.quantus/rewards-address.txt \
+  --validator \
+  --chain live_resonance \
+  --port 30333 \
+  --prometheus-port 9616 \
+  --name "ResonanceLiveTestnetNode" \
+  --experimental-rpc-endpoint "listen-addr=127.0.0.1:9944,methods=unsafe,cors=all"
```
grenade, line 34: id be tempted to lose the last 4 args.
- port 30333 is the default, so setting it in docs will encourage people to change it without sleuthing through the `--help` to understand what they're doing
- prometheus-port is only useful to people who run a prometheus server and again we're explicitly setting to the default port
- name will automatically generate something unique when unset. putting a value in our docs will result in 9/10 of every node in telemetry having the same name, which defeats the purpose of having a name
- there's no good reason for the public to run an unsafe rpc. anyone doing so needs to read the docs and most of them will just copy/paste this

n13: Thank you - done, also changed in the installer script.

PR #51, opened 2025-06-04
Make scheduler work with timestamps

dastansam:


n13 (commented):
needs to be updated from main and the latest changes in the other PR

pallets/scheduler/src/lib.rs
```diff
@@ -1496,136 +1273,111 @@ impl<T: Config> Pallet<T> {
...
 {
-    type Address = TaskAddress<BlockNumberFor<T>>;
-    type Hash = T::Hash;
+    type Address = TaskAddressOf<T>;
+    type Hasher = T::Hashing;
 
     fn schedule(
-        when: DispatchTime<BlockNumberFor<T>>,
+        when: DispatchBlock<BlockNumberFor<T>>,
         maybe_periodic: Option<schedule::Period<BlockNumberFor<T>>>,
```
n13, line 1284: Here maybe_periodic is only a blocknumber, no moment...

is this because v3 is a legacy method?

Should we just remove legacy methods? Seems like we don't need old versions around.

n13 (commented):
Added some comments.

n13:
Also I had o3 review this PR. (my comments upfront)

**I don't think this is an issue...** 
Weight audit. on_timestamp_set() now calls service_agendas() with u32::MAX as the “per-block limit”. A malicious user could queue many tiny timestamp buckets and force an expensive loop every block.→ Respect T::MaximumWeight, or iterate only through the bucket that matches now / TimestampBucketSize.

**Maybe...**
Enum SCALE layout not documented. BlockNumberOrTimestamp is persisted; once on chain it cannot change easily. Add a comment that the enum order must stay (0 => BlockNumber, 1 => Timestamp) and ensure #[repr(u8)] to pin it.

illuzen (approved):
Looks good, this was much easier to read thank you for splitting it up

illuzen:
> Also I had o3 review this PR. (my comments upfront)
> 
> **I don't think this is an issue... as far as I can tell it respects max weight and there's tests for that too** Weight audit. on_timestamp_set() now calls service_agendas() with u32::MAX as the “per-block limit”. A malicious user could queue many tiny timestamp buckets and force an expensive loop every block.→ Respect T::MaximumWeight, or iterate only through the bucket that matches now / TimestampBucketSize.
> 
> **Maybe...** Enum SCALE layout not documented. BlockNumberOrTimestamp is persisted; once on chain it cannot change easily. Add a comment that the enum order must stay (0 => BlockNumber, 1 => Timestamp) and ensure #[repr(u8)] to pin it.

I think this shouldn't be an issue if the time bucket is similar to the target block time. Then the max frequency would be about once per block.

PR #73, opened 2025-06-10
Speed up governance tests

n13:
🚀 Performance Gains
* Speed: Tests still run ~200,000x faster (milliseconds vs 44+ seconds)
* Block Usage: 2-60 blocks instead of 1.6M+ blocks
* Fast Governance: All governance tracks use 2-block periods for testing
* All tests succeed
* Tests with original track periods (slow times = lots and lots of blocks) can be run with --features slow-governance-tests 

Run with original value (slow)
```
cargo test --workspace --features slow-governance-tests
```

==================
Test semantics were preserved!


n13:
I agree it's not a good name, let me find a better name

n13:
renamed production-governance-tests

PR #74, opened 2025-06-11
Format toml files: fix indendation, order dependencies, make workspace dependency declaration consistent

dastansam:
- make workspace toml file a single source of truth for dependency versions
- order dependencies alphabetically (ascending)
- sort dependencies by domain - this helps when bumping `polkadot-sdk` deps
- consistent workspace dependency usage, i.e `crateA.workspace = true` instead of `crateA = { workspace = true }`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,49 +28,86 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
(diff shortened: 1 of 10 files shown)


n13 (approved):
Nice

PR #79, opened 2025-06-17
Patch in libp2p-noise-pqc

illuzen:
Simple change, please see https://github.com/Quantus-Network/libp2p-noise-pqc# to understand what this will do. 

Essentially we use post-quantum cryptography to encrypt the peer connections. You can see it in the logs by looking for 

🔊 Using pqc (ml-kem) ⚛ for noise!


illuzen:
Tested this with three nodes locally `scripts/run_local_nodes.sh`

illuzen:
Please review the unsafe code in protocol.rs::ensure_handshake_initialized @dastansam

illuzen:
I used the version that appears in our vendor. Clatter is an existing library

n13 (approved):

PR #81, opened 2025-06-18
Disable Litep2p

n13:
Litep2p is not quantum secure
Libp2p is the only network backend we support now


illuzen:
I'm closing this bc I have to do it just to get my libp2p-identity branch to compile, so will roll into one

PR #86, opened 2025-06-20
libp2p-identity-pqc

illuzen:
- Cargo.toml patches in both libp2p-identity-pqc and sc-network-pqc 
- the rusty-crystals api changed slightly, so we handle that
- in node/src/command.rs we overwrite NodeKeyConfig to make it dilithium. We do this to avoid having to fork sc-cli


n13 (approved):
LGTM

PR #87, opened 2025-06-22
Integrate wormhole from zk-circuits

vrom911:
Things to consider:
- We need a way to release a new `verifier.bin` and `common.bin`. I think it should be done through CI releases of new versions. Then we can use exact artefacts on the `chain`, it would be tightly coupled with the version of the `zk-circuits` we will be using.
- Same on the `chain` side: we need a proper way to use the bins of the specific version
- How do we ensure the authenticity of verifier.bin and common.bin files in CI? Should we include checksums/hashes of the verifier files in the code for verification?
- Need to double check the mining fee strategy (@illuzen @dastansam ), cause I had to change the `mint_into` approach we had before and am using the `widthrow` function of Balance (to be discussed, I guess).
- For the mining fees, how will we set up the specific `Account used as the "from" account when creating transfer proofs for minted tokens`

Depends on https://github.com/Quantus-Network/zk-circuits/pull/25


pallets/wormhole/src/lib.rs
```diff
@@ -149,53 +75,103 @@ pub mod pallet {
...
             ensure_none(origin)?;
 
-            let proof = ProofWithPublicInputs::from_bytes(proof_bytes.clone(), &*CIRCUIT_DATA)
-                .map_err(|_e| {
-                    // log::error!("Proof deserialization failed: {:?}", e.to_string());
-                    Error::<T>::ProofDeserializationFailed
-                })?;
+            let verifier_bytes = VerifierData::<T>::get();
+            ensure!(!verifier_bytes.is_empty(), Error::<T>::NotInitialized);
+            let verifier = WormholeVerifier::from_bytes(&verifier_bytes)
```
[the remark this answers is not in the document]
vrom911: @illuzen , good question! I looked into this to see if we could optimise the gas cost.

My plan was to perform the expensive `from_bytes` deserialization once in `initialize_verifier`, then store the resulting `WormholeVerifier` object in a SCALE-encoded format, so in the verify call, we only do the scale decode. This requires `Encode`/`Decode` instances.

Unfortunately, this is not possible to do. The `WormholeVerifier` struct wraps plonky2's `VerifierCircuitData`, which does not implement Substrate's `Encode` and `Decode` traits. This prevents us from storing the processed verifier object in the pallet's storage.

So, for now, it seems we're constrained to deserialising the raw verifier bytes on each call.

pallets/wormhole/src/lib.rs
```diff
@@ -178,43 +116,106 @@ pub mod pallet {
...
+                <[F; 2]>::try_from(
+                    &proof.public_inputs[FUNDING_AMOUNT_START_INDEX..FUNDING_AMOUNT_END_INDEX],
+                )
+                .map_err(|_| Error::<T>::InvalidPublicInputs)?,
+            );
 
-            BalancesPallet::<T, ()>::mint_into(&public_inputs.exit_account, exit_balance)?;
+            // Check for overflow before converting to Balance type
+            let exit_balance: <T as BalancesConfig>::Balance =
+                if exit_balance_u128 > u64::MAX as u128 {
```
illuzen, line 178: we should just set Balances to be u128 in testing, this conditional shouldn't be here

illuzen:
> Things to consider:
> 
> * We need a way to release a new `verifier.bin` and `common.bin`. I think it should be done through CI releases of new versions. Then we can use exact artefacts on the `chain`, it would be tightly coupled with the version of the `zk-circuits` we will be using.

Maybe we can roll this into the system @czareko made for checking genesis integrity?

> * Same on the `chain` side: we need a proper way to use the bins of the specific version

Maybe we give them names from the commit they were produced in and a timestamp for sorting? It's a little awkward, but I don't have a better idea at the moment.

> * How do we ensure the authenticity of verifier.bin and common.bin files in CI? Should we include checksums/hashes of the verifier files in the code for verification?

Putting it in CI might be a little too much, maybe we can specify releases of zk-circuits in Cargo.toml and generate it on build?

> * Need to double check the mining fee strategy (@illuzen @dastansam ), cause I had to change the `mint_into` approach we had before and am using the `widthrow` function of Balance (to be discussed, I guess).

I think we want to call increase_balance instead of mint_into or deposit_creating, since that will not increase the total issuance stat or create an imbalance. Then we can use withdraw to create an imbalance that the miner will collect as fees. 

> * For the mining fees, how will we set up the specific `Account used as the "from" account when creating transfer proofs for minted tokens`
> 
> Depends on [Quantus-Network/zk-circuits#25](https://github.com/Quantus-Network/zk-circuits/pull/25)

Merge with main, it's a fixed "minting account"

            // Create a transfer proof for the minted tokens
            let mint_account = T::MintingAccount::get();
            BalancesPallet::<T, ()>::store_transfer_proof(
                &mint_account,
                &public_inputs.exit_account,
                exit_balance,
            );

pallets/wormhole/src/lib.rs
```diff
@@ -178,43 +116,106 @@ pub mod pallet {
...
+                Error::<T>::ProofDeserializationFailed
+            })?;
+
+            // Public inputs are ordered as follows:
+            // Nullifier.hash: 4 felts
+            // StorageProof.funding_amount: 2 felts
+            // StorageProof.root_hash: 4 felts
+            // ExitAccount.address: 4 felts
+            //
+            // TODO: These constants should be exposed from the common crate.
```
illuzen, line 136: or maybe just use the parser from that crate?

illuzen:
In general looks good, can we do the weight info for the wormhole as well? Maybe @dastansam can help?

pallets/wormhole/src/lib.rs
```diff
@@ -43,114 +64,27 @@ pub mod pallet {
...
+        type FeeReceiver: OnUnbalanced<
+            <BalancesPallet<Self> as Currency<Self::AccountId>>::NegativeImbalance,
+        >;
     }
 
     pub trait WeightInfo {
         fn verify_wormhole_proof() -> Weight;
-        fn verify_wormhole_proof_with_used_nullifier() -> Weight;
-        fn verify_wormhole_proof_deserialization_failure() -> Weight;
-        fn verify_wormhole_proof_empty_data() -> Weight;
```
vrom911, line 55: This is actually the error states, and Substrate benchmarking framework is designed to measure the cost only of successful extrinsic execution. When an extrinsic returns an error (like `NullifierAlreadyUsed` or `ProofDeserializationFailed`), which it does in the above cases, it treats this as a benchmark failure rather than a success

PR #88, opened 2025-06-22
Proper benchmarking and weights

dastansam:
this is a general cleanup and enhancement of the current state of benchmarks:

- adds missing benchmarks, including for hooks
- fixes compiling but not running benchmarks
- fixes benchmarks properly benchmark bounded extrinsics (any param can be bounded)
- converts old style extrinsics to latest v2
- refactoring of some pallets code

NOTE: commented out `pallet-wormhole` benchmarks since calling `verify_wormhole_proof()` panics on runtime because the static .hex files are empty. once it is working, we can remove comments and re-run the benchmarks


pallets/faucet/src/benchmarking.rs
```diff
@@ -0,0 +1,28 @@
+//! Benchmarking for pallet faucet
```
n13, line 1: Kinda overkill since faucet pallet will be removed ;) but OK!

n13 (approved):
Very nice. LGTM and nice to have this cleaned up!

illuzen (approved):

PR #97, opened 2025-06-26
Added blocknumber to TransferProof

illuzen:
Pretty straightforward change, but this will ensure no collisions in the nullifier or storage key in the case that a user deletes their account and uses it again, which resets the nonce back to zero.


illuzen:
this depends on https://github.com/Quantus-Network/poseidon-resonance/pull/6

PR #101, opened 2025-06-26
Refactoring and clean up of pallets

dastansam:
- convert info logs to trace with proper targets
- remove unnecessary tight pallet coupling to loose coupling, i.e `Currency` instead of `pallet_balances::Config`
- remove unnecessary comments, events
- use explicit aliased types 

and more


illuzen (approved):
looks good, thanks for cleaning this up

n13 (approved):
more log::infos gone - great ;)

Cargo.toml
```diff
@@ -144,6 +144,6 @@ substrate-wasm-builder = { version = "25.0.0", default-features = false }
 [patch.crates-io]
 sp-state-machine = { git = "https://github.com/Quantus-Network/zk-state-machine" }
 sp-trie = { git = "https://github.com/Quantus-Network/zk-trie" }
-sc-network = { git = "https://github.com/Quantus-Network/sc-network-pqc" }
```
illuzen, line 147: I don't think the version gets used here and this is the wrong branch at this point, it should be main after we merge 48.5 in

PR #102, opened 2025-06-27
Global Transfer Counter

illuzen:
This addresses some issues around TransferProofs
- now wormhole exits generate TransferProofs
- TransferProof hashes depend on a global TransferCount that increments on all transfers 

This removes the dependency on account nonce, as this has some corner cases like minting or recurring payments or account deletion that would make the nullifier inputs (blocknum, nonce, from) collide. Now all transfers have a globally unique id.


n13 (approved):
LGTM

The counter seems to be counting

Issue #95, opened 2025-06-25
Scheduler can skip timestamps / buckets


n13:
Sometimes, pending scheduled transactions remain pending forever. 

I think it's because our time buckets assume that they will always be hit by blocks, but if the block time happens to be a lot longer than a time bucket, a time bucket may be skipped

E.g 

block time = 10 seconds

Bucket 1 = 100s
Bucket 2 = 110s
Bucket 3 = 120s

Now we run the chain, and block 1 happens at 100s time, but then we take a lot longer to mine block 2, and it happens at 130s time

Now the scheduler is called with timestamp 130, and no incomplete agendas are known because at 100ms there were no agendas, and at 130 there are also no agendas. 

110 and 120 agendas were skipped. 

Fix would be to keep a last checked time bucket variable around

Blocks are not affected by this because no matter how long it takes to mine a block, they are in sequence and never skip. 

### Screenshot 

Listed transaction is hanging there forever, even as new blocks are minted. 

1. Start release node like this:
```
RUST_LOG=error,scheduler=debug ./target/release/quantus-node --dev --tmp
```

2. Run quantus-cli like this:

```sh
cargo run -- reversible schedule-transfer-with-delay --to 5DGBqCtK2EGeSwXKG5k9zT9zjpVLAdBzCeJmu6gvfmLKS4KW --amount 1.0 --delay 11 --from crystal_alice

cargo run -- reversible list-pending --from crystal_alice
```

<img width="1982" alt="Image" src="https://github.com/user-attachments/assets/123cfac0-f505-4283-8f50-dcb9aaa07006" />

PR #103, opened 2025-06-27
fix for skipped buckets and also incomplete blocks

n13:
# Two big + 1 small scheduler fixes

## Fix 1: Time based buckets are no longer skipped

#95 

## Fix 2: Conflict with time based on block based timestamps

In the edge case when both blocks and timestamps would store a IncompleteSince value, they would override each other causing horrible problems and stuck tasks. 

## Fix 3: Weight limit is observed across time based and block based angendas.

Prior, they were run in different contexts

### DIFF
--- a/scripts/pre_checkin_checks.sh
+++ b/scripts/pre_checkin_checks.sh
@@ -1,4 +1,5 @@
 cargo fix --all
+cargo fmt
 
 taplo format --check --config taplo.toml
 


pallets/scheduler/src/lib.rs
```diff
@@ -381,11 +393,25 @@ pub mod pallet {
...
             Self::service_agendas(
                 &mut weight_counter,
                 BlockNumberOrTimestamp::BlockNumber(now),
                 u32::max_value(),
             );
+
+            // Process timestamp-based agendas using current system time
+            // This ensures no buckets are skipped if block times are longer than bucket intervals
+            let current_timestamp = T::TimeProvider::now();
+            if current_timestamp > T::Moment::zero() {
```
illuzen, line 407: when would current_timestamp be zero?

illuzen:
Can we just split service_agendas into two functions, one that gets called for timestamps and one for block numbers?

pallets/scheduler/src/lib.rs
```diff
@@ -921,59 +949,93 @@ enum ServiceTaskError {
...
         {
-            return;
+            if !Self::service_agenda(
+                weight,
+                &mut executed,
+                BlockNumberOrTimestamp::BlockNumber(current_block),
+                BlockNumberOrTimestamp::BlockNumber(when),
+                u32::max_value(),
+            ) {
+                incomplete_since = incomplete_since.min(when);
```
illuzen, line 975: seems like this could just be assignment, since when always increments

illuzen:
LGTM

n13:
Had to fix tests - because we now run everything on block (on initialize) the weights tests were all off, after I removed the timestamp == 0 check. 

Was a good idea to remove that since it did not belong in the code at all, but having it there made the unit tests work.

Goes to show that sometimes Claude has a bad programmer mentality - just make it work, even if it's the wrong thing to do. 

This has been fixed now. The weights tests were all passing if we skip timestamp based weights completely, which proves they were untouched, so I simply updated the expected values.

PR #104, opened 2025-06-29
Some fixes to latest scheduler changes

dastansam:
since we always check if we can consume `MaxScheduledPerBlock` of agendas in `on_initialize` we risk to miss a lot of timestamp based agendas if there are close to `MaxScheduledPerBlock` block based agendas. So, I propose that we take half of the max blocks weight per servicing each type of agendas


illuzen (approved):
Good call

n13:
I want to go over this lets put this in the next release ... hold off a second

n13:
Good fix

PR #111, opened 2025-07-01
feat: Genesis verification

czareko:


grenade (approved):
lgtm. i would like to say thank you for this effort. in my view, taking the time to do this right as you have done here is a big deal. i feel like chains that don't make this effort are a dime a dozen and have a longevity that matches the attention span of the wen moon crowd. i'm proud to be working with a team that takes these things seriously and sees value in this sort of transparency. thank you!

PR #118, opened 2025-07-02
Reversible events

illuzen:
- Add more details to events around reversible transactions
- remove delay policy to align code with high security use case
- remove optionality where we don't want to support it
- rename reversible to high security
- rename explicit_reverser to interceptor
- remove code branches where possible

### DIFF
--- a/taplo.toml
+++ b/taplo.toml
@@ -1,6 +1,10 @@
 # all options https://taplo.tamasfe.dev/configuration/formatter-options.html
 
-# global rules
+# This is a taplo configuration file.
+# For more information, see: https://taplo.tamasfe.dev/configuration/
+include = ["**/*.toml"]
+exclude = ["**/target/**"]
+
 [formatting]
 reorder_arrays = true
 inline_table_expand = false


pallets/reversible-transfers/src/lib.rs
```diff
@@ -507,16 +498,16 @@ pub mod pallet {
             // Release the funds
```
n13, line 498: this is unfreeze below right... release...

pallets/reversible-transfers/src/lib.rs
```diff
@@ -242,16 +226,21 @@ pub mod pallet {
...
             who: T::AccountId,
-            data: ReversibleAccountData<T::AccountId, BlockNumberOrTimestampOf<T>>,
+            interceptor: T::AccountId,
+            recoverer: T::AccountId,
+            delay: BlockNumberOrTimestampOf<T>,
         },
         /// A transaction has been intercepted and scheduled for delayed execution.
-        /// [who, tx_id, execute_at_moment]
+        /// [from, to, interceptor, amount, tx_id, execute_at_moment]
         TransactionScheduled {
```
n13, line 239: nice

n13:
Fixed CI
Also updated the pre checkin script to be correct

PR #125, opened 2025-07-03
cannot start timestamps from 0

n13:
because it will then iterate through all timestamps starting in 1970...

Ok so this was a bug on chain startup - we have to prevent setting "last processed" timestamp to 0 (zero)

Otherwise the scheduler would check all times from 0 (1970) to now, which can take a very, very long time.


pallets/scheduler/src/lib.rs
```diff
@@ -407,11 +409,15 @@ pub mod pallet {
...
             // This ensures no buckets are skipped if block times are longer than bucket intervals
             let current_timestamp = T::TimeProvider::now();
-            Self::service_timestamp_agendas(
-                &mut weight_counter,
-                current_timestamp,
-                u32::max_value(),
-            );
+
+            log::debug!(target: "scheduler", "on_initialize: current_timestamp: {:?}", current_timestamp);
+            if current_timestamp > T::Moment::zero() {
```
illuzen, line 414: why check if it is nonzero?

illuzen (approved):

PR #126, opened 2025-07-03
Improve miner accounting

illuzen:
This changes how miners and the treasury are rewarded.

- all rewards generate TransferProofs
- Treasury reward is fixed, per block, and explicit
- all fees go to miner, this is because we want the incentive to be for miners to include transactions, sorted by fees

It doesn't quite take the last step to forcing miners to use wormholes, but it does enable it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -478,9 +478,9 @@ dependencies = [
 
 [[package]]
 name = "async-channel"
-version = "2.3.1"
+version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


illuzen:
agreed, to be clear, this PR doesn't force them, it only enables it by creating transfer proofs

n13 (commented):
Did this whole treasury reward as a percentage or reward go out the window?

I think that got lost? On purpose?

n13:
Ok we all got convinced that indeed, fixed reward is better. Cool!

n13 (approved):
LGTM

PR #127, opened 2025-07-04
Global nonce for reversibles

dastansam:


illuzen (approved):
good simplification

n13 (approved):
LGTM! Great stuff!

PR #133, opened 2025-07-09
Config Update + SigHandler

illuzen:
- ctrl+c signal handler for miner
- improved logs
- changed various config numbers
- removed unused code

### DIFF
--- a/dilithium-crypto/verify/src/lib.rs
+++ b/dilithium-crypto/verify/src/lib.rs
@@ -6,7 +6,7 @@ pub fn verify(pub_key: &[u8], msg: &[u8], sig: &[u8]) -> bool {
     match PublicKey::from_bytes(pub_key) {
         Ok(pk) => pk.verify(msg, sig, None),
         Err(e) => {
-            log::warn!(target: "dilithium-verify", "public key failed to deserialize {:?}", e);
+            log::warn!("public key failed to deserialize {:?}", e);
             false
         }
     }
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -435,10 +435,10 @@ pub fn new_full<
                                     let current_version = worker_handle.version();
                                     if current_version == version {
                                         if futures::executor::block_on(worker_handle.submit(seal.encode())) {
-                                            log::info!("Successfully mined and submitted a new block via external miner");
+                                            log::debug!(target: "miner", "Successfully mined and submitted a new block via external miner");
                                             nonce = U512::zero();
                                         } else {
-                                            log::warn!("Failed to submit mined block from external miner");
+                                            log::warn!(target: "miner", "Failed to submit mined block from external miner");
                                             nonce += U512::one();
                                         }
                                     } else {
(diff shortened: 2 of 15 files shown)


node/src/service.rs
```diff
@@ -377,12 +392,30 @@ pub fn new_full<
                 let mut current_job_id: Option<String> = None;
 
                 loop {
+                    // Check for cancellation
+                    if mining_cancellation_token.is_cancelled() {
```
n13, line 396: Does this mean normal Ctrl-C won't work anymore? Does this intercept ctrl-c?

node/src/service.rs
```diff
@@ -402,7 +435,10 @@ pub fn new_full<
                             Ok(d) => d,
                             Err(e) => {
                                 log::warn!("Failed to get distance_threshold: {:?}", e);
-                                tokio::time::sleep(Duration::from_millis(250)).await;
+                                tokio::select! {
+                                    _ = tokio::time::sleep(Duration::from_millis(250)) => {},
```
n13, line 439: unrelated but wondering if 250ms is too long a time to do nothing at all in a pow node?

dilithium-crypto/verify/src/lib.rs
```diff
@@ -6,7 +6,7 @@ pub fn verify(pub_key: &[u8], msg: &[u8], sig: &[u8]) -> bool {
     match PublicKey::from_bytes(pub_key) {
         Ok(pk) => pk.verify(msg, sig, None),
         Err(e) => {
-            log::warn!(target: "dilithium-verify", "public key failed to deserialize {:?}", e);
+            log::warn!("public key failed to deserialize {:?}", e);
```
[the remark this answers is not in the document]
illuzen: yes because I think warnings should show up regardless of whether you have that target turned on

node/src/service.rs
illuzen: no it means it stops the whole program immediately instead of waiting for it to find a block before quitting

node/src/service.rs
illuzen: i think this is an edge case because I don't recall ever seeing this log, it just returns an obligatory Result, so we have to handle that case

runtime/src/configs/mod.rs
```diff
@@ -130,29 +129,31 @@ impl pallet_mining_rewards::Config for Runtime {
...
+    type TreasuryBlockReward = ConstU128<UNIT>; // 1 token
     type TreasuryPalletId = TreasuryPalletId;
     type MintingAccount = MintingAccount;
 }
 
 parameter_types! {
-    /// Target block time
-    pub const TargetBlockTime: u64 = 10000;
+    /// Target block time ms
+    pub const TargetBlockTime: u64 = 20000;
```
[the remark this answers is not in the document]
illuzen: yeah and leaves more room for smaller machines to mine. later we can easily tighten it up, increase throughput and make it a marketing event "performance boost"

PR #144, opened 2025-07-19
Shortened difficulty adjustment period to reduce wobble

illuzen:
Screenshot 2025-07-19 at 5 22 44 PM

This wobble in the difficulty is directly downstream of this parameter. It's not a huge problem, but smaller wobbles would make mining more predictable. Right now we sometimes get a 5x change in difficulty in one wobble.


n13:
So this makes it more sensitive in the short term. 

Does it do a exponential moving average (EMA) or is it just a linear adjustment?

illuzen:
> So this makes it more sensitive in the short term.
> 
> Does it do an exponential moving average (EMA) or is it just a linear adjustment?

Just linear adjustment, but we could make ema, its a little more computation every block, but not a ton. Maybe we save ema for when we do kaspa style parallel blocks?

PR #146, opened 2025-07-22
Wormhole pallet storage root check

### LINKED ISSUE #145: Wormhole pallet storage root check
The proofs have a value root_hash which needs to be checked to see if it was a real storage root. To facilitate O(1) storage root lookup, we should add an argument to the verify extrinsic that specifies which block number it should be in. This doesn't need to go in the circuit.

vrom911:
Resolves #145


illuzen (approved):
LGTM

PR #150, opened 2025-07-28
Upgrade sdk dependencies to stable2506 and move forked crates to the workspace

dastansam:
- moved crates: sc-network, sp-state-machine, sp-trie - first commit is the copied version as is
- unified all dependency versions in the workspace toml file
-


n13:
Can you update description with which crates exactly were folded into the repo?

I suppose these are modified by us - what is the benefit of having them inside a monorepo?

PR #151, opened 2025-07-29
Add sync-awareness to pow

illuzen:
This is an attempt to address the issue of miners forking themselves off randomly.

We could add more conditions, like need at least 1 peer to mine?


n13 (approved):
That sounds reasonable

PR #161, opened 2025-08-06
add fmt config 

dastansam:
copied from [here](https://github.com/paritytech/polkadot-sdk/blob/master/.rustfmt.toml)


n13 (approved):
I guess it's needed for the mono repo. 

Not quite sure what the advantage is of having a mono repo but don't mind about formatting styles too much. So GTG

illuzen (approved):
this should make it easier to audit the monorepo

PR #164, opened 2025-08-12
Upgrade sdk dependencies to stable2506 

dastansam:


illuzen (approved):
ok primitives/trie is identical with zk-trie (feat/fmt branch).

PR #162, opened 2025-08-12
Move forked crates to monorepo, upgrade to stable2506

dastansam:
this is the base branch that will be merged to main

PR #168, opened 2025-08-12
Apply sp-trie updates

dastansam:
This is the top of the stack and if this passes the CI, we should merge from the top of the stack and merge #162 to main


illuzen (approved):

illuzen:
don't merge yet, the chain doesn't run bc Litep2p crept back into the defaults

PR #169, opened 2025-08-14
Add difficulty and distance_achieved to chain_getBlock

illuzen:
- merge submit_nonce and verify_for_import into verify_current_block
- start nonce search at 1, not 0
- post_digests in header now contains difficulty, distance_achieved and seal in separate values

it's a little messy to return a 3-tuple in a runtime-api call, but the values we return are all intermediate values on the block verification, so we might as well return them all at once instead of recomputing them.

### DIFF
--- a/client/consensus/qpow/src/lib.rs
+++ b/client/consensus/qpow/src/lib.rs
@@ -14,8 +14,10 @@ use sp_runtime::generic::BlockId;
 use std::{marker::PhantomData, sync::Arc};
 
 #[derive(Clone, Debug, Encode, Decode, PartialEq)]
-pub struct QPoWSeal {
+pub struct QPoWResult {
 	pub nonce: [u8; 64],
+	pub difficulty: [u8; 64],
+	pub distance_achieved: [u8; 64]
 }
 
 pub struct QPowAlgorithm<B, C>
@@ -61,29 +63,8 @@ where
 		_pre_digest: Option<&[u8]>,
 		seal: &RawSeal,
 		_difficulty: Self::Difficulty,
-	) -> Result<bool, Error<B>> {
-		//Executed for mined and imported blocks
-
-		/*
-
-		For now, we will turn this off temporally.
-		In this way, node can mine/import blocks without rewarding anyone.
-
-		// Block miner should exist
-
-		let mut extracted_author: Option<AccountId32> = None;
-		if let Some(pre_digest_bytes) = pre_digest {
-			if let Ok(account) = <AccountId32 as Decode>::decode(&mut &pre_digest_bytes[..]) {
-				extracted_author = Some(account);
-			}
-		}
-
-		let _author = match extracted_author {
-			Some(acc) => acc,
-			None => return Err(Error::Runtime("Failed to extract AccountId32 from pre_digest".into())),
-		};
-
-		*/
+	) -> Result<(bool, U512, U512), Error<B>> {
+		// Executed for mined and imported blocks
 
 		// Convert seal to nonce [u8; 64]
 		let nonce: [u8; 64] = match seal.as_slice().try_into() {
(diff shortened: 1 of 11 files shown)


runtime/src/apis.rs
```diff
@@ -127,18 +127,14 @@ impl_runtime_apis! {
...
 		fn verify_historical_block(header: [u8; 32], nonce: [u8; 64], block_number: u32) -> bool {
 			// Convert u32 to the appropriate BlockNumber type used by your runtime
 			let block_number_param = block_number;
 			pallet_qpow::Pallet::<Self>::verify_historical_block(header, nonce, block_number_param)
 		}
 
-		fn submit_nonce(header: [u8; 32], nonce: [u8; 64]) -> bool {
-			pallet_qpow::Pallet::<Self>::submit_nonce(header, nonce)
+		fn verify_current_block(header: [u8; 32], nonce: [u8; 64], emit_event: bool) -> (bool, U512, U512) {
+			pallet_qpow::Pallet::<Self>::verify_current_block(header, nonce, emit_event)
```
n13, line 137: why is this renamed? Is it not submitting a nonce?

illuzen: it is but there is another api call that is called verify_historic_block that does almost the same thing, and the call does verification and returns data not just accepting submission fwiw

node/src/external_miner_client.rs
```diff
@@ -45,7 +44,7 @@ pub(crate) async fn check_mining_result(
 	client: &Client,
 	miner_url: &str,
 	job_id: &str,
-) -> Result<Option<QPoWSeal>, String> {
+) -> Result<Option<[u8; 64]>, String> {
```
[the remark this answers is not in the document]
illuzen: yeah the struct was not really necessary, we could use NonceType if you insist

client/consensus/qpow/src/lib.rs
```diff
@@ -14,8 +14,10 @@ use sp_runtime::generic::BlockId;
 use std::{marker::PhantomData, sync::Arc};
 
 #[derive(Clone, Debug, Encode, Decode, PartialEq)]
-pub struct QPoWSeal {
+pub struct QPoWResult {
```
n13: result is a much better name

pallets/qpow/src/tests.rs
```diff
@@ -96,7 +96,7 @@ fn test_submit_valid_proof() {
 }
 
 #[test]
-fn test_verify_for_import() {
+fn test_verify_current_block() {
```
n13, line 99: it seems this deletes the existing test? We don't need verify for import test anymore?

illuzen: the import and verify paths have been merged

client/consensus/qpow/src/lib.rs
```diff
@@ -92,22 +65,27 @@ where
...
 			.client
 			.runtime_api()
-			.verify_for_import(parent_hash, pre_hash, nonce)
-			.map_err(|e| Error::Runtime(format!("API error in verify_nonce: {:?}", e)))?
-		{
-			return Ok(false);
+			.verify_current_block(parent_hash, pre_hash, nonce, false)
+			.map_err(|e| Error::Runtime(format!("API error in verify_nonce: {:?}", e)))?;
+
+		// Verify the nonce using QPoW
```
n13, line 78: this comment may be out of place - i am happy to delete most comments..

node/src/service.rs
```diff
@@ -366,7 +366,7 @@ pub fn new_full<
 
 		task_manager.spawn_essential_handle().spawn("qpow-mining", None, async move {
 			log::info!("⛏️ QPoW Mining task spawned");
-			let mut nonce: U512 = U512::zero();
+			let mut nonce: U512 = U512::one();
```
n13, line 369: any particular reason we want to start with nonce 1 instead of 0?

client/consensus/qpow/src/lib.rs
illuzen: yeah it would be more appropriate to put it on top of the verify_current_block call, not below

illuzen: also not sure why _difficulty is even an argument here, the verify call pulls it from the runtime...ugh ok i will clean this up

n13 (approved):
Looks good, some minor comments

PR #181, opened 2025-08-21
Remove duplicate call to algorithm.verify

illuzen:
submit calls algorithm.verify, but later calls import_block which also calls algorithm.verify, so we just remove the first one, especially since it now generates an event

Screenshot 2025-08-21 at 10 35 07 AM


n13:
verify shouldn't create an event - that's a side effect. 

Verify should verify. 

The event should be created on import block, no? 

This means that external miner seal is only checked on import block - I think that's fine. 

I don't like side effects in verify, that is code smell.

PR #200, opened 2025-09-06
Sync fix 2 final

n13:
Note: Compare with branch sync-fix-2 to see the delta in the sync engine code. 

# What issue does this fix

We noticed that nodes where the network is bad fail to sync - the sync stalls and miners have to just reset everything and try again

It was a common occurrence in the chat, and also our own testing

## Causes of the stall
- A good network connection would cover up the issue - no problems
- On a bad connection, even moderately bad, two things would happen:
- 1. The client would make multiple requests on the exact same block range to the remote peer, leading to a ban from the remote peer after 2 attempts. 
- 2. The client itself would ban the remote peer after a timeout during fetch, meaning it would remove that peer from the list and not try it again for a while, successive fails would keep increasing the timespan of the ban. The ban is persisted through restarts which is why users had to wipe the data directory to make the sync start again

### Overview 

1. Don't drop or report peers during a major sync

New CLI parameter: **--sync-max-timeouts-before-drop** defaults to 10. Set to 0 to disable the feature. 

2. Don't make the same request twice so the remote peer does not ban us. 

### Implementation 

- Adaptive smaller requests on failed requests
- No duplicated requests so the remote peer doesn't ban us
- Only enabled in major sync mode so it doesn't interfere with normal operations
- Allowing peers a number of failures before reporting or banning them
- It now blasts through bad connections and gets them at some point

### Safety
This fix is client side only - only a peer that is in major syncing is affected. No effect on network as a whole, during operations. 

## The Sync Engine

The sync engine is complex, it works similar to bit torrent, trying to get all blocks in a range before importing them, repeating requests and sizing and adjusting requests based on the blocks we did get in requests - a request for 64 blocks would for example load less than 64 - the client stores this information, and makes successive request to load the rest; all this is already in the sync engine. Understanding how it works is required for a fix.

On the remote peer side a maximum of 8MB for a single request is enforced (hard coded) , so if blocks add up to more than 8MB, we get fewer. 

But in our test that was not the issue, the issue was that we could not handle larger requests than a few MB on a bad connection without timing out and mutual peer banning.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ target/
 /vendor
 /scripts/apps
 /scripts/js/node_modules
+/test_data
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9820,8 +9820,6 @@ dependencies = [
 [[package]]
 name = "sc-network-sync"
 version = "0.50.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2ae94e1d2003a4d6e069c95765005720aacca8c4e9fc9109cecbca281fd09848"
 dependencies = [
  "array-bytes 6.2.3",
  "async-channel 1.9.0",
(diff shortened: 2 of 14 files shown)


n13:
This is now working as intended, however, I haven't been able to see it truly rescue heisenberg syncs in bad network conditions. 
There might be something missing, need to analyze it more.

n13:
Putting this on ice until we see sync issues with heisenberg.

PR #204, opened 2025-09-10
Monorepo sc-cli

illuzen:
Move sc-cli into client/cli and change ed25519 things over to dilithium.

Not sure we need all these files, just copied the directory in.


n13 (approved):
I checked against the OG branch where the CLI was added - made a branch for this purpose called sc-cli-added-to-monorepo for this purpose. 

There's then some other unrelated changes but the changes to CLI are super clear and easy to understand. 

Looks good! 

The hacky code in node was removed also.

PR #216, opened 2025-09-17
correct ss58 id for known value test

n13:
using 189 - less confusion 

otherwise this is unchanged


czareko (approved):

n13:
> can we extract this to `const SS58_FORMAT: u16 = 189;` and use it in chain spec functions as well?

Will add later

PR #217, opened 2025-09-18
EMA implemented

illuzen:
EMA for difficulty update.

Will clean up, we need to add ema to the prometheus stats and there's also no need for some of the storage data now.


node/src/service.rs
```diff
@@ -518,7 +522,7 @@ pub fn new_full<
 							log::warn!("API error getting threshold: {:?}", e);
 							U512::zero()
 						});
-					let nonces_to_mine = 3000u64;
+					let nonces_to_mine = 3_000_000u64;
```
n13, line 525: why mine 3M nonces? might take too long. Also I changed this to 300 recently, not sure why it is back to 3000.

node/src/service.rs
```diff
@@ -531,10 +535,13 @@ pub fn new_full<
 					};
 
 					let current_version = worker_handle.version();
+					// TODO: what does this check do?
```
n13, line 538: This check checks if the nonce we mined is still relevant - we might have a new block already

node/src/service.rs
n13: Ok lets change it back to 300, although, I have another PR now which also does that.

node/src/service.rs
```diff
@@ -474,11 +477,12 @@ pub fn new_full<
 									if futures::executor::block_on(
 										worker_handle.submit(seal.encode()),
 									) {
-										log::info!("🥇Successfully mined and submitted a new block via external miner");
+										let mining_time = mining_start_time.elapsed().as_secs();
```
n13, line 480: Is mining time purely for printing it out on the command line?

illuzen: yes

node/src/service.rs
n13: Ok nevermind... please merge main into this, set it back to 300 and added the threading fix in main

n13: Actually I merged it, was only 1 line

n13 (approved):
Nice that this is a lot simpler !

primitives/consensus/qpow/src/lib.rs
```diff
@@ -35,17 +28,11 @@ sp_api::decl_runtime_apis! {
...
 		/// Get total work
 		fn get_total_work() -> U512;
 
-		/// Get sum of block times in rolling history
-		fn get_block_time_sum() -> u64;
-
-		/// Get median block time for preconfigured list of elements
-		fn get_median_block_time() -> u64;
+		/// Get block ema
+		fn get_block_time_ema() -> u64;
```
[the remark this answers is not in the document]
n13: We should probably do this even though heisenberg is anyway not backwards compatible.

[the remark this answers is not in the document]
illuzen: Yeah I think let's skip the backwards compatibility this time, since we are resetting testnet

PR #220, opened 2025-09-21
Illuzen/poseidon2

illuzen:
- replaces poseidon with poseidon2 in the state hasher gadgets via new qp-poseidon
- QPoW now uses poseidon2 instead of sha2-256 and sha3-512

Tested that multi-node script works. External miner will have to be updated


n13 (approved):
LGTM

n13 (changes requested):
Fix CI

czareko (approved):

PR #235, opened 2025-09-25
fix: Logs formatting - libs downgrade

czareko:


n13 (approved):
Is there no way to fix these versions in the toml files? 

Or are they transient dependencies?

PR #241, opened 2025-09-25
Fix metadata-hash build and add to CI

dastansam:


n13 (approved):
Makes the release build work - good

czareko (approved):

PR #242, opened 2025-09-25
pre flight check script matches CI with build check

n13:
No changes to code


n13:
looks like it has a bug

PR #269, opened 2025-10-02
mine range needs to ignore zero nonce

n13:
zero nonces are ignored by the normal mine algo also


illuzen (commented):
This should not be what you are looking for, the probability of nonce_element (which is different from nonce) being all zeros is extremely small.

What was the bug you saw?

PR #278, opened 2025-10-10
Volume fee for high security reversal

illuzen:
If a high security account reverses their transaction, the treasury gets 1% of it

### DIFF
--- a/pallets/reversible-transfers/src/lib.rs
+++ b/pallets/reversible-transfers/src/lib.rs
@@ -27,12 +27,7 @@ pub use weights::WeightInfo;
 use alloc::vec::Vec;
 use frame_support::{
 	pallet_prelude::*,
-	traits::{
-		tokens::{
-			fungibles::MutateHold as AssetsHold, ExistenceRequirement, Fortitude, Restriction,
-		},
-		Currency,
-	},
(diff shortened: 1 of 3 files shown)


illuzen:
ok @dastansam i think this is ready for re-review

pallets/reversible-transfers/src/lib.rs
```diff
@@ -820,34 +844,74 @@ pub mod pallet {
 			// Cancel the scheduled task
 			T::Scheduler::cancel_named(schedule_id).map_err(|_| Error::<T>::CancellationFailed)?;
 
-			// For assets, transfer held funds to interceptor via assets-holder; for native
-			// balances, transfer held funds to interceptor
+			// Calculate volume fee only for high-security accounts
+			let (fee_amount, remaining_amount) = if high_security_account_data.is_some() {
+				let volume_fee = T::VolumeFee::get();
+				let fee = volume_fee * pending.amount;
```
dastansam: checked arithmetic should be used here

dastansam:
> ok @dastansam i think this is ready for re-review

don't know what happened, but seems like some changes were reverted back (check for zero is still there, etc.)

PR #279, opened 2025-10-10
fix: Flaky Trie Cache - weak references excluded

czareko:


n13 (approved):
whatever makes this test pass

PR #288, opened 2025-10-20
Illuzen/improve logs again

illuzen:
- reduce "Import block" frequency since it makes the screen become a bottleneck on syncing
- remove unused sha2 and sha3 packages


n13:
Also I like it can still be turned on for each block for debug mode. This was pretty important for analyzing our import stalling back when, so when we have a similar situation we need this to identify the problematic blocks

PR #295, opened 2025-11-03
Verify header in the wormhole proof

dastansam:


pallets/wormhole/src/lib.rs
ethan-crypto: The `block_number` and `header` params are redundant since the public inputs to the proof already contains these values. We can retrieve the block hash from storage using the block number in the public inputs and then constrain that that hash matches the public inputs hash.

pallets/wormhole/src/tests.rs
ethan-crypto: There is a lot of code duplication across the helper functions here and the helpers [here](https://github.com/Quantus-Network/qp-zk-circuits/blob/feat/block-header-proof/wormhole/example/src/utils.rs)

I propose that we consolidate these helpers into the `qp-zk-circuits-common` crate, adding them as utility functions that we import into both places.

illuzen (approved):
Looks good but we reintroduced git deps?

PR #298, opened 2025-11-06
Add arm64 platform to docker image workflows output

dewabisma:
## Problem
I want to run node as a docker container in my mac machine but it said I can't which is kind of sad 😢 . 

## Proposed Solution
Add linux/arm64 to the docker image workflows


n13:
Failing test is just our wonky test that's off by one sometimes, can be ignored.

PR #310, opened 2025-11-09
feat: qp-header

ethan-crypto:
A fork of sp-runtime's Header with a modified hash function to ensure a felt aligned pre-image

### DIFF
--- a/primitives/header/src/lib.rs
+++ b/primitives/header/src/lib.rs
@@ -174,7 +174,7 @@ where
 	}
 }
 
-#[cfg(all(test, feature = "std"))]
+#[cfg(test)]
 mod tests {
 	use super::*;
 	use qp_poseidon::PoseidonHasher;
--- a/primitives/header/Cargo.toml
+++ b/primitives/header/Cargo.toml
@@ -36,5 +36,5 @@ std = [
 	"qp-poseidon/std",
 	"scale-info/std",
 	"sp-core/std",
-	"sp-runtime/std"
+	"sp-runtime/std",
 ]


primitives/header/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+		));
+
+		// digest – injective encoding
+		felts.extend(injective_bytes_to_felts::<Goldilocks>(&self.digest.encode()));
+
+		let poseidon_hash: [u8; 32] = hash_variable_length(felts);
+		poseidon_hash.into()
+	}
+}
+
```
dastansam: I think we should include these [lines](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/runtime/src/generic/header.rs#L135-L153) as well. there might be downstream code that calls this function instead of using `Header` trait one.

dastansam: also, then you could have our impl inside this function and simply call it in the trait impl override

primitives/header/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+	serialization::{injective_bytes_to_felts, unsafe_digest_bytes_to_felts},
+};
+use scale_info::TypeInfo;
+use sp_core::U256;
+use sp_runtime::{
+	generic::Digest,
+	traits::{AtLeast32BitUnsigned, BlockNumber, Hash as HashT, MaybeDisplay, Member},
+	RuntimeDebug,
+};
+use sp_std::vec::Vec;
```
dastansam: also, sp-std is deprecated, you should use `alloc` instead

dastansam (approved):

PR #318, opened 2025-11-14
Rewards address only on validator

n13:
- reverted the entire commit making rewards address mandatory
- added requirement to validator - requires rewards address.

Tested, it works.


dastansam:
Also, don't revert changes to mining doc

PR #322, opened 2025-11-20
Just use alice if --dev and set dev tax to 0

illuzen:
Per @czareko 's complaint

Also setting dev tax to 0 for now as tokenomics have shifted.


n13:
I'd rather delete the "expect" we have on the rewards address

Instead of this, it should default to treasury address when rewards address is not supplied. 

Then we remove the dreaded expect from the code, and we keep CLI the same so people don't forget to set rewards address, but also cover dev mode or other weird things we didn't think about yet.

PR #330, opened 2025-12-03
fix: sc-network v0.51.1 -  peers broadcast fix

czareko:
By default, peers were sending large address lists that included ephemeral ports, causing memory bloat and slow peer-to-peer broadcasting. 
This fix filters incoming peer addresses to only accept blockchain P2P ports (30333-30533) and increases buffer sizes for handling multiple simultaneous peer connections. 
Internal tests showed significant improvements in peer broadcasting speed and reduced node memory usage.

### DIFF
--- a/client/network/src/service.rs
+++ b/client/network/src/service.rs
@@ -1336,17 +1336,8 @@ impl<'a> NotificationSenderReadyT for NotificationSenderReady<'a> {
 /// - TIER 2 (relaxed): If TIER 1 returns 0 addresses, accepts private IPs but still filters ports
 ///
 /// If `disable_filtering` is true, all addresses are returned without filtering.
-fn filter_peer_addresses(
-	addrs: Vec<Multiaddr>,
-	peer_id: &PeerId,
-	disable_filtering: bool,
-) -> Vec<Multiaddr> {
+fn filter_peer_addresses(addrs: Vec<Multiaddr>, peer_id: &PeerId) -> Vec<Multiaddr> {
 	use multiaddr::Protocol;
-
-	// If filtering is disabled, return all addresses
-	if disable_filtering {
-		return addrs;
-	}
 
 	let original_count = addrs.len();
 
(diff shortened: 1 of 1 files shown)


n13 (commented):
I like all the increased param sizes, that's great

I would like for the filtering this to be optional so we can change it without shipping a new binary as a safety switch.

client/network/src/service.rs
```diff
@@ -1299,6 +1328,117 @@ impl<'a> NotificationSenderReadyT for NotificationSenderReady<'a> {
...
+						is_public = false; // Localhost or private IPv4
+					},
+					_ => {},
+				}
+			}
+
+			has_valid_port && is_public
+		})
+		.cloned()
+		.collect();
```
ethan-crypto:
```suggestion
	// TIER 1: Strict filter (preferred ports + public IPs only)
	// TIER 2: Relaxed filter - accept private IPs but still filter ports
	let (strict_filtered, relaxed_filtered): (Vec<_>, Vec<_>) = addrs
		.iter()
		.cloned()
		.fold((Vec::new(), Vec::new()), |(mut strict, mut relaxed), addr| {
			let mut has_valid_port = false;
			let mut is_link_local = false;
			let mut is_public = true;

			for proto in addr.iter() {
				match proto {
					Protocol::Tcp(port) => {
						has_valid_port = port >= MIN_P2P_PORT && port <= MAX_P2P_PORT;
					},
					Protocol::Ip6(ip) if ip.segments()[0] == 0xfe80 => {
						is_link_local = true; // Link-local IPv6
					},
					Protocol::Ip4(ip) if ip.is_loopback() || ip.is_private() => {
						is_public = false; // Localhost or private IPv4
					},
					_ => {}
				}
			}

			let relaxed_ok = has_valid_port && !is_link_local;
			let strict_ok  = relaxed_ok && is_public;

			if relaxed_ok {
				relaxed.push(addr.clone());
			}
			if strict_ok {
				strict.push(addr);
			}

			(strict, relaxed)
		});
```

PR #333, opened 2025-12-08
Use canonical balances pallet and add support for assets in wormhole

dastansam:
This PR does mainly two things:

- Get rid of balances pallet fork and consolidate proofs in the wormhole pallet instead. Proofs, transfer count, and nullifiers all are now stored in the wormhole pallet.
- Add new wormhole proof recorder transaction extension that intercepts successful (post-dispatch) asset and native token transfers and records their transfer proof. Similar to reversible transactions extension except for post dispatch nature. Additionally, there are two new extrinsics in the wormhole pallet for native and asset transfers (they won't be intercepted by the tx extension)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7234,7 +7234,7 @@ dependencies = [
  "pallet-timestamp",
  "parity-scale-codec",
  "primitive-types 0.13.1",
- "qp-poseidon-core 1.0.2",
+ "qp-poseidon-core",
  "qpow-math",
  "scale-info",
  "sp-arithmetic",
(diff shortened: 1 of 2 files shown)


pallets/wormhole/src/tests.rs
```diff
@@ -1,262 +1,142 @@
...
-	fn generate_proof(inputs: CircuitInputs) -> ProofWithPublicInputs<F, C, 2> {
-		let config = CircuitConfig::standard_recursion_config();
-		let prover = WormholeProver::new(config);
-		let prover_next = prover.commit(&inputs).expect("proof failed");
-		let proof = prover_next.prove().expect("valid proof");
-		proof
-	}
 
 	#[test]
-	fn test_wormhole_transfer_proof_generation() {
```
illuzen: did we lose this test entirely?

illuzen:
overall LGTM, can we go back to having descriptions on PRs? Maybe just make them like one paragraph

pallets/wormhole/src/tests.rs
dastansam: no, it's back in later commits

czareko (changes requested):
Please update the wormhole branch first, and then add here only the changes you actually need.

PR #340, opened 2025-12-18
Exponentially decaying token rewards

illuzen:
- 21m max coins
- 50% mining rewards go to treasury
- heuristic: emit tokens = (max_supply - current_supply) / constant 

this heuristic results in a close approximation of an exponential curve. The constant is chosen so that, with no burns, the token supply will be 99% emitted in about 45 years.

### DIFF
--- a/pallets/mining-rewards/src/lib.rs
+++ b/pallets/mining-rewards/src/lib.rs
@@ -113,6 +113,14 @@ pub mod pallet {
 			let emission_divisor = T::EmissionDivisor::get();
 
 			let remaining_supply = max_supply.saturating_sub(current_supply);
+
+			if remaining_supply == BalanceOf::<T>::zero() {
+				log::warn!(
+					"💰 Emission completed: current supply has reached the configured maximum, \
+					 no further block rewards will be minted."
+				);
+			}
+
 			let total_reward = remaining_supply
 				.checked_div(&emission_divisor)
 				.unwrap_or_else(|| BalanceOf::<T>::zero());


pallets/mining-rewards/src/lib.rs
```diff
@@ -52,13 +52,17 @@ pub mod pallet {
...
 
-		/// The base block reward given to treasury
+		/// The divisor used to calculate block rewards from remaining supply
 		#[pallet::constant]
-		type TreasuryBlockReward: Get<BalanceOf<Self>>;
+		type EmissionDivisor: Get<BalanceOf<Self>>;
+
+		/// The portion of rewards that goes to treasury (out of 100)
+		#[pallet::constant]
+		type TreasuryPortion: Get<u8>;
```
dastansam: this would be better represented as a perbill or permill

pallets/mining-rewards/src/mock.rs
```diff
@@ -28,7 +28,8 @@ pub type Block = frame_system::mocking::MockBlock<Test>;
 parameter_types! {
 	pub const BlockHashCount: u64 = 250;
 	pub const SS58Prefix: u8 = 189;
-	pub const BlockReward: u128 = 50;
+	pub const MaxSupply: u128 = 21_000_000;
```
n13: Would make more sense to store this value as unit, e.g. multiply by 10E12?

runtime/src/genesis_config_presets.rs
```diff
@@ -54,12 +54,15 @@ fn dilithium_default_accounts() -> Vec<AccountId> {
...
-	let mut balances =
-		endowed_accounts.iter().cloned().map(|k| (k, 1u128 << 60)).collect::<Vec<_>>();
+	let mut balances = endowed_accounts
+		.iter()
+		.cloned()
+		.map(|k| (k, 100_000 * UNIT))
+		.collect::<Vec<_>>();
 
-	const ONE_BILLION: u128 = 1_000_000_000;
+	const INITIAL_TREASURY: u128 = 21_000_000 * 30 / 100; // 30% tokens go to investors
```
n13: Same here everything should be in UNIT

pallets/mining-rewards/src/lib.rs
```diff
@@ -101,24 +105,54 @@ pub mod pallet {
...
-			log::debug!(target: "mining-rewards", "💰 Original Tx_fees: {:?}", tx_fees);
+			// Log readable amounts (convert to tokens by dividing by 1e12)
+			if let (Ok(total), Ok(treasury), Ok(miner_amt), Ok(current), Ok(fees)) = (
+				TryInto::<u128>::try_into(total_reward),
+				TryInto::<u128>::try_into(treasury_reward),
+				TryInto::<u128>::try_into(miner_reward),
+				TryInto::<u128>::try_into(current_supply),
+				TryInto::<u128>::try_into(tx_fees),
+			) {
+				let unit = 1_000_000_000_000u128;
```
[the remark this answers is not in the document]
illuzen: unfortunately there's no easy way to access it, it's in a different crate, I think it's better to redefine it here instead of creating a dependency, especially since it's only used for logging.

illuzen:
> Maybe 45 years is a bit too long?

Maybe, but I also was trying to keep the initial inflation below 10%, it's currently 7%. I have a spreadsheet I can show you.

n13 (approved):
Looks good now

PR #344, opened 2025-12-22
Enforce miner wormhole address

illuzen:
This does not involve meaningful change to the block header. Instead, we reinterpret the 32 byte accountid as the wormhole address preimage and send rewards there. 

Also deleted the redundant and outdated mining info in the readme and updated mining.md


pallets/mining-rewards/src/lib.rs
```diff
@@ -101,30 +108,67 @@ pub mod pallet {
...
+			}
+
+			let total_reward = remaining_supply
+				.checked_div(&emission_divisor)
+				.unwrap_or_else(|| BalanceOf::<T>::zero());
+
+			// Split the reward between treasury and miner
+			let treasury_portion = T::TreasuryPortion::get();
+			let treasury_reward =
+				total_reward.saturating_mul(treasury_portion.into()) / 100u32.into();
```
dastansam: if we use `Permill` or `Perbill` we wouldn't need to have this division here

PR #364, opened 2026-01-28
fix: Multisig - auto-cleaning expanded

czareko:


n13:
I still think we should clean up at most 5 proposals, and always charge for 5 (which is a small charge) to make this less fragile. 

I am not super comfortable with iterating over 200 items potentially.

n13:
Update:

I checked the Gemini issue, and it is correct, we are refunding fees when we don't find expired proposals yet we are doing the work of checking all active proposals, which we need to charge for.

n13 (approved):
I think an explicit cleanup call would be a better architecture

However, there's nothing wrong with this as is, so GTG

PR #365, opened 2026-01-29
feat: High-Security Integration for Multisig Pallet

czareko:


ethan-crypto (commented):
One issue I noticed is that in the benchmarks we aren't actually computing the worst-case for the cleanup process because we aren't setting the call data to the max size for the expired proposals. This could lead to underweighting.

pallets/multisig/src/benchmarking.rs
ethan-crypto:
```suggestion
		for i in 0..e {
			let proposer_for_this = if i < cleaned_target {
				caller.clone()      // will be removed
			} else {
				signer1.clone()     // will be iterated, not removed
			};
			let system_call = frame_system::Call::<T>::remark { remark: vec![i as u8; 10] };
			let call = <T as Config>::RuntimeCall::from(system_call);
			let encoded_call = call.encode();
			let bounded_call: BoundedCallOf<T> = encoded_call.try_into().unwrap();
			let bounded_approvals: BoundedApprovalsOf<T> = vec![proposer_for_this.clone()].try_into().unwrap();

			let proposal_data = ProposalDataOf::<T> {
				proposer: proposer_for_this.clone(),
				call: bounded_call,
				expiry: expired_block,
				approvals: bounded_approvals,
				deposit: 10u32.into(),
				status: ProposalStatus::Active,
			};
			Proposals::<T>::insert(&multisig_address, i, proposal_data);
		}
```

ethan-crypto (changes requested):
- Right now we aren't making an effective use of the `r` regression coefficient. I suggested some changes that would account for this by defining a `cleaned_target = i.min(r)` which will only clean a subset of the total proposals, resulting in a more accurate weighting and a greater refund for `proposal` calls

- The cleanup path remains expensive per proposal as long as call lives inside the same storage value. We should make the cleanup iteration cost depend only on the small fields we care about. To fix this we could either go with a **split storage** approach: 

```
 ProposalMeta (small: proposer/status/expiry/deposit/…) used by cleanup iteration

 ProposalCall (big) only touched on execute / remove
```
**OR** store call as a **preimage / hash (call_hash, call_len)** and require bytes only when executing.

czareko:
Finally, we decided to simplify the scope again, so benchmarks and weights are less sophisticated now.

ethan-crypto (changes requested):
Overall I really like the new design introducing the execute method and confining all clean up operations to extrinsic calls. However there are still some places where underweighting can occur:

- `cancel`, `remove_expried` and `claim_deposits` are all benchmarked with a small fixed call size. This will inevitably lead to underweighting. We should parameterize cancel weight by stored call length like we do with the other methods that touch proposal storage. For `claimed_deposits` we would need to compute the average `call` data size over the set of proposal in storage.

-  Even though the `approve_dissolve` weight function prices it as constant, it has two different execution paths: **reached** and **not reached**. We should adopt a similar pattern that accounts for this like we did with `propose`; (1) charge worst-case up front, then (2) refund to the cheaper path via `actual_weight`, and (3) benchmark both paths.

- Even though the `approvals` field is much cheaper to decode compared to `call`, the decode cost also scales with approvals length. The benchmarks currently do not parameterize approvals lengths and doesn't cover the worst case (100 approvals), so this could lead to minor underweighting unless we fix the approvals to 100 or parameterize.

PR #391, opened 2026-03-04
fix: Reading work from discarded state

czareko:
The node then stayed idle and stopped producing blocks.

## Root cause

1. **ArchiveCanonical pruning** – The node keeps state only for blocks on the canonical chain. Non-canonical fork blocks have their state pruned.

2. **`get_total_work` needs state** – Best chain selection calls `runtime_api().get_total_work(block_hash)` for each leaf. That requires executing the runtime at that block, so the block’s state must exist.

3. **Many forks** – With many competing forks (reorg depth 168–179 in the logs), some leaves are tips of fork chains whose state has already been pruned.

4. **Failure mode** – When `calculate_chain_work` hit such a block, it returned an error. The whole `best_chain()` call failed, so the mining loop could not get a best block and stopped authoring.

## Solution

Instead of failing the entire best chain selection when `calculate_chain_work` returns a “State already discarded” / `UnknownBlock` error:

- **For leaves:** Skip the problematic leaf, add it to ignored chains, and continue evaluating the rest.
- **For current best:** If its state is pruned, fall back to `find_best_chain(leaves)` and pick the best among evaluable leaves.

The node can now choose the best chain from leaves whose state is available and continue authoring blocks.

### DIFF
--- a/client/consensus/qpow/src/chain_management.rs
+++ b/client/consensus/qpow/src/chain_management.rs
@@ -148,11 +148,14 @@ where
 		let best_number = *best_header.number();
 		log::debug!("Current best block number: {}", best_number);
 
-		let max_reorg_depth = self
-			.client
-			.runtime_api()
-			.get_max_reorg_depth(best_hash)
-			.expect("Failed to get max reorg depth");
+		let max_reorg_depth =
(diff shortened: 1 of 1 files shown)


illuzen (approved):
This seems correct, but we may want to rewrite this whole section to be more obviously correct

n13:
I am checking this now

PR #395, opened 2026-03-06
Achieved difficulty chain selection

illuzen:
## Summary
- Refactored chain selection to use **achieved difficulty** (actual mining work) stored in aux storage instead of target difficulty
- Moved finalization from an **async background task** to **synchronous execution during block import**, eliminating race conditions
- Removed the `SelectChain` abstraction entirely - fork choice is now fully determined during import, so the mining worker queries the client directly
- Removed ~600 lines of now-redundant code from `chain_management.rs`

## Motivation
The previous implementation had two issues:

1. **Race condition in finalization**: Finalization ran on a separate async task listening to import notifications, meaning the next block could be imported before finalization of the previous block completed. This could lead to inconsistent state where competing forks weren't properly pruned.

2. **Unnecessary abstraction**: The `SelectChain` trait (`HeaviestChain`) was vestigial - after moving fork choice into `import_block()`, `best_chain()` was just returning `client.info().best_hash`, adding no value.

By making finalization synchronous (called at the end of `import_block()`) and removing the `SelectChain` layer, we now guarantee that:
1. Finalization completes before the next block import begins
2. Substrate's `displaced_leaves_after_finalizing()` properly prunes forks that diverged before the finalized block
3. The mining worker gets the authoritative best block directly from the client

## Key Changes

**`client/consensus/qpow/src/lib.rs`**
- Added `BE: Backend` type parameter to `PowBlockImport` for `Finalizer` trait bound
- Call `finalize_canonical_at_depth()` synchronously at end of `import_block()`
- Store achieved work in aux storage BEFORE inner import (needed for fork choice callbacks)
- Removed `S: SelectChain` type parameter from `PowBlockImport` and `start_mining_worker`
- Mining worker now calls `client.info().best_hash` and `client.header()` directly

**`client/consensus/qpow/src/chain_management.rs`**
- Added standalone `finalize_canonical_at_depth<B, C, BE>()` function
- Removed `HeaviestChain` struct and `SelectChain` implementation
- Removed all unused helper methods:
  - `evaluate_leaf()`, `try_calculate_chain_work()`, `find_best_chain()`
  - `find_common_ancestor_and_depth()`, `is_chain_ignored()`, `add_ignored_chain()`
- Removed `IGNORED_CHAINS_PREFIX` and ignored chain tracking (Substrate handles this via displaced leaves)
- Cleaned up unused error variants and imports

**`node/src/service.rs`**
- Removed `FullSelectChain` type alias
- Removed `select_chain` creation and passing
- Updated `PowBlockImport` type alias
- Updated `spawn_authority_tasks` signature

## How Fork Pruning Works
When `finalize_block()` is called, Substrate's `displaced_leaves_after_finalizing()` (in `sp-blockchain/src/backend.rs`) identifies leaves whose ancestry doesn't pass through the finalized block. These displaced leaves are removed from the leaves set, and their blocks can be pruned. This eliminates the need for custom "ignored chains" tracking.

### DIFF
--- a/client/consensus/qpow/src/lib.rs
+++ b/client/consensus/qpow/src/lib.rs
@@ -326,8 +326,6 @@ where
 
 		// Store cumulative achieved work BEFORE inner import, because inner import
 		// triggers notifications that call best_chain which needs this data.
-		// This MUST succeed - if it fails, subsequent blocks would read zero for
-		// this block's work, breaking the cumulative chain weight calculation.
 		store_cumulative_achieved_work::<B, C>(&*self.client, block_hash, new_work).map_err(
 			|e| {
 				ConsensusError::ClientImport(format!(
@@ -357,9 +355,7 @@ where
 			},
 		};
 
-		// Finalize blocks synchronously after import to ensure finalization happens
-		// before the next block is imported. This prunes competing forks that are
-		// beyond max_reorg_depth.
+		// Finalization prunes competing forks that are beyond max_reorg_depth.
 		if let Err(e) = finalize_canonical_at_depth::<B, C, BE>(&*self.client) {
 			log::warn!(
 				target: LOG_TARGET,


pallets/qpow/src/lib.rs
```diff
@@ -379,6 +379,28 @@ pub mod pallet {
...
+		pub fn verify_and_get_achieved_difficulty(
+			block_hash: [u8; 32],
+			nonce: NonceType,
+		) -> (bool, U512) {
+			let (valid, _difficulty, hash_achieved) =
+				Self::verify_nonce_internal(block_hash, nonce);
+			let achieved_difficulty = achieved_difficulty_from_hash(hash_achieved);
+
+			if valid {
+				Self::deposit_event(Event::ProofSubmitted {
```
czareko: I did this type of mistake in the past as well. This type of call is not in the state context, so we can't deposit any event, but substrate won't show any error about it of course. :p

Issue #406, opened 2026-03-09
Reversible recover_funds doesn't close out pending reversible transactions


n13:
Add these unit tests at the end of high_security_tests
```
// --- Known issues tracked for future fix (require AccountId → PendingTransfers index) ---

#[test]
#[ignore = "recover_funds does not yet cancel pending transfers or release holds"]
fn recover_funds_includes_held_pending_amounts() {
	new_test_ext().execute_with(|| {
		System::set_block_number(1);
		let hs_user = alice();
		let guardian = bob();
		let dest = charlie();
		let amount = 500;

		let initial_hs_balance = Balances::free_balance(&hs_user);
		let initial_guardian_balance = Balances::free_balance(&guardian);

		assert_ok!(ReversibleTransfers::schedule_transfer(
			RuntimeOrigin::signed(hs_user.clone()),
			dest.clone(),
			amount,
		));

		assert_eq!(
			Balances::balance_on_hold(
				&RuntimeHoldReason::ReversibleTransfers(HoldReason::ScheduledTransfer),
				&hs_user
			),
			amount
		);

		assert_ok!(ReversibleTransfers::recover_funds(
			RuntimeOrigin::signed(guardian.clone()),
			hs_user.clone()
		));

		assert_eq!(
			Balances::free_balance(&guardian),
			initial_guardian_balance + initial_hs_balance,
			"Guardian should receive ALL funds including held amounts"
		);

		assert_eq!(
			Balances::balance_on_hold(
				&RuntimeHoldReason::ReversibleTransfers(HoldReason::ScheduledTransfer),
				&hs_user
			),
			0,
			"No holds should remain after recovery"
		);
	});
}

#[test]
#[ignore = "recover_funds does not yet cancel pending transfers or release holds"]
fn recover_funds_cancels_pending_transfers() {
	new_test_ext().execute_with(|| {
		System::set_block_number(1);
		let hs_user = alice();
		let guardian = bob();
		let dest = charlie();
		let amount = 500;

		let call = transfer_call(dest.clone(), amount);
		let tx_id = calculate_tx_id::<Test>(hs_user.clone(), &call);

		assert_ok!(ReversibleTransfers::schedule_transfer(
			RuntimeOrigin::signed(hs_user.clone()),
			dest.clone(),
			amount,
		));
		assert!(ReversibleTransfers::pending_dispatches(tx_id).is_some());

		assert_ok!(ReversibleTransfers::recover_funds(
[...]

PR #408, opened 2026-03-09
recover_funds should cancel pending first

illuzen:
## Summary

Add `PendingTransfersBySender` index to track pending transfers per account, enabling `recover_funds` to cancel all pending transfers in a single call.

### Core Features
- **`PendingTransfersBySender` storage**: Maps sender accounts to their pending transfer IDs (bounded to `MaxPendingPerAccount = 16`)
- **`recover_funds` enhancement**: Now cancels all pending transfers before transferring remaining balance to guardian
- **`TooManyPendingTransactions` error**: Prevents scheduling more than 16 pending transfers per account

### Refactoring
- Extracted shared `do_cancel_transfer` function used by both `cancel_transfer` and `recover_funds`
- `release_held_funds_with_fee` helper handles volume fee calculation and fund release
- Volume fee (1% burn) applies to each cancelled transfer during recovery

### Test Additions
- `recover_funds_cancels_all_pending_transfers` - verifies all pending transfers are cancelled with fees applied
- `too_many_pending_transactions_error` - verifies the 16-transfer limit is enforced
- `chained_guardian_high_security_account_flow` - runtime integration test for guardian accounts that are also high-security

### Bug Fixes
- Fixed `MockHighSecurity::is_whitelisted` in multisig tests to use exact match instead of contains (was matching "unsafe" as whitelisted because it contains "safe")

## Behavior

- `recover_funds` uses worst-case weight (always charges for 16 cancellations)
- Scheduler errors are ignored during recovery (`strict_scheduler=false`) to ensure recovery completes even if some scheduled tasks are already executed
- `TransactionCancelled` event emitted for each cancelled pending transfer


n13:
Here's my review of **PR #408: recover_funds should cancel pending first**.

---

## Overview

Clean, well-structured PR. The `PendingTransfersBySender` index, the `release_held_funds_with_fee` extraction, and the recovery loop are all solid. Tests are thorough. A few issues worth addressing:

---

## Issues

### 1. Two independent `if let` branches in `release_held_funds_with_fee` - could double-execute

```860:912:pallets/reversible-transfers/src/lib.rs
			if let Ok((call, _)) = T::Preimages::realize::<RuntimeCallOf<T>>(&pending.call) {
				if let Ok(pallet_assets::Call::transfer_keep_alive { id, .. }) =
					call.clone().try_into()
				{
					// ... handles asset path
				}
				if let Ok(pallet_balances::Call::transfer_keep_alive { .. }) =
					call.clone().try_into()
				{
					// ... handles balance path
				}
			}
```

The second `if` should be `else if`. These are mutually exclusive call types, but as written, if a call could somehow satisfy both `try_into()` conversions, both branches would execute - burning and transferring twice. Defensively, the second should be `else if`.

### 2. `recover_funds` charges worst-case weight but never refunds

```494:497:pallets/reversible-transfers/src/lib.rs
		#[pallet::weight(
			<T as Config>::WeightInfo::recover_funds()
				.saturating_add(<T as Config>::WeightInfo::cancel().saturating_mul(T::MaxPendingPerAccount::get().into()))
		)]
```

An account with 0 pending transfers pays the same as one with 16. Since the function already returns `DispatchResultWithPostInfo`, you could track the actual number of cancellations and return consumed weight for a refund:

```rust
let num_cancelled: u64 = /* count from loop */;
let actual_weight = <T as Config>::WeightInfo::recover_funds()
    .saturating_add(
        <T as Config>::WeightInfo::cancel()
            .saturating_mul(num_cancelled)
    );
Ok(Some(actual_weight).into())
```

Not critical for an emergency function, but it's free weight back for the guardian.

### 3. `recover_funds` benchmark doesn't include pending transfers

```220:235:pallets/reversible-transfers/src/benchmarking.rs
	fn recover_funds() -> Result<(), BenchmarkError> {
		let account: T::AccountId = whitelisted_caller();
		let guardian: T::AccountId = benchmark_account("guardian", 0, SEED);
		fund_account::<T>(&account, BalanceOf::<T>::from(10000u128));
		fund_account::<T>(&guardian, BalanceOf::<T>::from(10000u128));
		let delay = T::DefaultDelay::get();
		setup_high_security_account::<T>(account.clone(), delay, guardian.clone());
		#[extrinsic_call]
		_(RawOrigin::Signed(guardian.clone()), account.clone());
		Ok(())
	}
```

The benchmark sets up a HS account with **zero** pending transfers, so it only benchmarks the `transfer_all` path. The composite weight annotation adds `cancel * 16` to compensate, but that's an approximation - the inline recovery loop (uses `take`, swallows scheduler errors, calls `release_held_funds_with_fee` directly) has different overhead than the full `cancel` extrinsic. Ideally the benchmark would schedule `MaxPendingPerAccount` transfers first to measure the actual worst case.

### 4. Stale data if scheduling fails mid-way (pre-existing, minor)

```746:766:pallets/reversible-transfers/src/lib.rs
			PendingTransfers::<T>::insert(tx_id, new_pending);

			// Add to sender's pending list
			PendingTransfersBySender::<T>::try_mutate(&from, |list| {
				list.try_push(tx_id).map_err(|_| Error::<T>::TooManyPendingTransactions)
			})?;

			let bounded_call = T::Preimages::bound(Call::<T>::execute_transfer { tx_id }.into())?;

			// Schedule the `do_execute` call
			T::Scheduler::schedule_named(
				// ...
			)
			.map_err(|e| {
				// ...
				Error::<T>::SchedulingFailed
			})?;

			// hold funds...
```

If `Preimages::bound`, `schedule_named`, or `hold` fails, `PendingTransfers` and `PendingTransfersBySender` already have entries. This is safe because FRAME extrinsics are transactional by default, so all storage mutations roll back on error. Just noting for awareness - if this inner function is ever called from a non-transactional context, it would need guarding.

---

## Positives

- **DRY refactoring**: `release_held_funds_with_fee` nicely deduplicates the fee calculation + fund release logic between `cancel_transfer` and `recover_funds`.
- **`take()` in the recovery loop** is clean - atomically reads and clears the sender index in one operation.
- **Error handling improvement**: `burn_held` and `transfer_on_hold` results are now properly propagated with `?` instead of silently discarded with `let _ =`.
- **Test coverage** is solid - covers the happy path (multiple cancellations + fee math + events), the 16-transfer limit, and balance/issuance accounting.
- **Bug fix** in the multisig mock (`is_whitelisted` exact match vs contains) is a nice catch.

---

## Verdict

Issues 1-3 are worth addressing before merge. #1 is a correctness/safety concern (use `else if`), #2 is a nice-to-have UX improvement, and #3 would make the weight accounting more accurate. Otherwise this is a solid PR.

n13:
1 - 3 addressed

n13:
Fixes #406

n13 (approved):
GTG

PR #413, opened 2026-03-11
chore: sc-network full update to 0.55.1

czareko:


n13:
I have everything I need. Let me compile the review.

## PR #413 Review — Impact on the Peer‑Ban Tolerance Fix

### Short answer
Your "don't immediately ban peers on timeout" fix is **not affected** by this PR. The tolerance logic lives in `client/network-sync/`, which PR 413 does not touch.

### Where the fix lives
```1039:1106:client/network-sync/src/engine.rs
let threshold = self.strategy.peer_drop_threshold();

let should_drop_peer = !should_gate || *entry >= threshold;
...
RequestFailure::Network(OutboundFailure::Timeout) =>
    if should_drop_peer {
        debug!(target: LOG_TARGET, "dropping peer after timeout! {:?}", peer_id);
        ...
        self.network_service.disconnect_peer(
```

Plus the block‑request back‑off in `client/network-sync/src/strategy/chain_sync.rs` and the `--max-timeouts` / block‑request‑timeout CLI flags in `node/src/cli.rs`. PR 413's file list:

- `client/network/Cargo.toml`, `build.rs`, `benches/*`
- `client/network/src/{bitswap,config,discovery,event,lib,service,service/metrics,service/signature}.rs`
- `client/network/src/protocol/{message,notifications}.rs`, `notifications/{behaviour,tests}.rs`

Nothing under `client/network-sync/`, and none of the `sc-network` public symbols the fix depends on (`OutboundFailure`, `RequestFailure`, `NetworkService::disconnect_peer`, `PeerStoreProvider`) change.

### BUT — two sibling hardenings ARE being silently reverted

These live in the very file set the PR rewrites and are in the same spirit as the ban‑tolerance fix (helping slow / overloaded bootnodes). They matter because tighter buffers produce more timeouts, which — even with your threshold — eventually trip drops.

**1. Inbound stream budget cut 8x** in `client/network/src/service.rs`:

```117:127:client/network/src/service.rs
const NOTIFY_HANDLER_BUFFER_SIZE: usize = 32;
...
const PER_CONNECTION_EVENT_BUFFER_SIZE: usize = 24;
...
/// Maximum number of negotiating inbound streams.
/// Increased from 2048 to handle many peers simultaneously opening substreams
/// for DHT queries, sync requests, etc. on bootnodes.
const MAX_NEGOTIATING_INBOUND_STREAMS: usize = 16384;
```

PR 413 inlines literals and reverts `16384 → 2048` (the `32` and `24` values are preserved, just un‑named). That 8× reduction is exactly what caused pile‑ups on bootnodes under load.

**2. Kademlia packet size limit removed** in `client/network/src/discovery.rs`:

```114:114:client/network/src/discovery.rs
const KAD_MAX_PACKET_SIZE: usize = 2 * 1024 * 1024;
```

```253:253:client/network/src/discovery.rs
			config.set_max_packet_size(KAD_MAX_PACKET_SIZE);
```

PR 413 deletes both the const and the `set_max_packet_size` call, falling back to libp2p's default. Large FIND_NODE responses from bootnodes with long peer lists will be truncated/dropped.

### Other things worth flagging (not related to the peer‑ban fix)

- **Default backend flip**: `NetworkBackendType::default()` goes `Libp2p → Litep2p`, while the code still errors `"This build only supports the Libp2p network backend"` if Litep2p is selected. Confirm the CLI path in `node/src/command.rs` explicitly passes `Libp2p` on every spawn, otherwise new binaries will fail at startup.
- **Notifications tests deleted**: `client/network/src/protocol/notifications/tests.rs` (≈220 lines) is removed. Not a runtime concern, but weakens coverage for future notification‑protocol changes.
- **Benches stubbed out** (acknowledged in the file comments as libp2p‑only in the Quantus fork).
- **`Keypair`/`PublicKey` wrapped** in a new `service::signature::{Keypair, PublicKey}` enum with a single `Libp2p` variant. Internal refactor; no behavioural impact on ban logic.

### Recommendation
Approve the ban‑tolerance part — it's untouched and the APIs it calls are stable. Before merging, ask the author to either:

1. Restore `MAX_NEGOTIATING_INBOUND_STREAMS = 16384` and the `set_max_packet_size(2 MiB)` call (preserving the original rationale comments), or
2. Explicitly justify dropping them in the PR description so the decision is tracked.

Also confirm `NetworkBackendType` is pinned to `Libp2p` at every instantiation site now that the default has flipped.

n13 (approved):

n13:
"Restore MAX_NEGOTIATING_INBOUND_STREAMS = 16384 and the set_max_packet_size(2 MiB) call (preserving the original rationale comments), or
"

I do not recall changing these values

PR #428, opened 2026-03-14
fix: Dilithium API - from_raw public constuctor

czareko:
Add from_raw constructor to DilithiumPair
After the previous change that hid the secret and public fields, there was no way to reconstruct a pair from raw bytes. This adds from_raw(public, secret) so callers can restore a pair from stored/serialized key material (e.g. wallet restore), with validation via create_keypair.


czareko:
@codex review

illuzen (approved):

Issue #431, opened 2026-03-14
Pallet Treasury Review


n13:
Here's my review of the treasury pallet:

---

**Overall**: The pallet is well-structured and simple. Root-only access control is correct, portion validation is sound, and the `TreasuryProvider` trait integration with mining-rewards works properly. That said, there are a few items worth flagging.

---

### Security Concerns

**1. Zero-account fallback silently swallows funds**

```101:112:pallets/treasury/src/lib.rs
	impl<T: Config> Pallet<T> {
		/// Get the treasury account. Returns zero account if not configured.
		pub fn account_id() -> T::AccountId {
			TreasuryAccount::<T>::get().unwrap_or_else(|| {
				T::AccountId::decode(&mut sp_runtime::traits::TrailingZeroInput::zeroes())
					.unwrap_or_else(|_| {
						// Fallback: zero account
						T::AccountId::decode(&mut &[0u8; 32][..])
							.unwrap_or_else(|_| panic!("Cannot create fallback AccountId"))
					})
			})
		}
```

If `TreasuryAccount` is `None` (storage cleared, migration error, etc.), this returns a zero-address that nobody controls. Mining-rewards will happily `mint_into` that address, permanently locking funds. Consider returning `Option<T::AccountId>` and letting the caller decide, or panicking early in genesis/integrity checks to guarantee it's always set.

**2. No validation on `set_treasury_account`**

```77:82:pallets/treasury/src/lib.rs
		pub fn set_treasury_account(origin: OriginFor<T>, account: T::AccountId) -> DispatchResult {
			ensure_root(origin)?;
			TreasuryAccount::<T>::put(&account);
			Self::deposit_event(Event::TreasuryAccountUpdated { new_account: account });
			Ok(())
		}
```

Root can set the treasury to the zero address (or any address). A minimal guard against the zero-account would prevent accidental misconfiguration.

**3. Redundant decode fallback is dead code**

The inner `T::AccountId::decode(&mut &[0u8; 32][..])` branch is unreachable -- `TrailingZeroInput::zeroes()` produces an infinite stream of zero bytes, so decoding `AccountId32` from it will always succeed. Dead fallback branches create a false sense of safety; simplify to a single decode.

---

### Correctness

**4. `TreasuryPortion` defaults to 0 via `ValueQuery`**

```39:39:pallets/treasury/src/lib.rs
	pub type TreasuryPortion<T: Config> = StorageValue<_, u8, ValueQuery>;
```

[...]

PR #434, opened 2026-03-15
fix: Treasury Pallet - Review #431

czareko:
#431

### DIFF
--- a/pallets/treasury/src/lib.rs
+++ b/pallets/treasury/src/lib.rs
@@ -44,18 +44,21 @@ pub mod pallet {
 
 	#[pallet::genesis_config]
 	pub struct GenesisConfig<T: Config> {
-		pub treasury_account: T::AccountId,
-		pub treasury_portion: Permill,
+		pub treasury_account: Option<T::AccountId>,
+		pub treasury_portion: Option<Permill>,
 	}
 
 	impl<T: Config> Default for GenesisConfig<T>
 	where
 		T::AccountId: From<[u8; 32]>,
 	{
-		/// Placeholder for tests/chain-spec defaults. Production must set treasury_account
-		/// explicitly.
+		/// Default for test runtimes and chain specs that omit treasury.
+		/// Production uses genesis_config_presets which set treasury explicitly.
 		fn default() -> Self {
-			Self { treasury_account: [1u8; 32].into(), treasury_portion: Permill::from_percent(50) }
+			Self {
+				treasury_account: Some([1u8; 32].into()),
+				treasury_portion: Some(Permill::from_percent(50)),
+			}
 		}
 	}
 
@@ -65,11 +68,19 @@ pub mod pallet {
 		T::AccountId: From<[u8; 32]> + PartialEq,
 	{
 		fn build(&self) {
-			assert!(self.treasury_portion <= Permill::one(), "Treasury portion must be <= 100%");
+			let account = self
+				.treasury_account
+				.as_ref()
+				.expect("Treasury account must be set in genesis; chain is misconfigured");
+			let portion = self
+				.treasury_portion
+				.as_ref()
+				.expect("Treasury portion must be set in genesis; chain is misconfigured");
+			assert!(*portion <= Permill::one(), "Treasury portion must be <= 100%");
 			let zero: T::AccountId = [0u8; 32].into();
-			assert!(self.treasury_account != zero, "Treasury account must not be zero address");
-			TreasuryAccount::<T>::put(self.treasury_account.clone());
-			TreasuryPortion::<T>::put(self.treasury_portion);
+			assert!(account != &zero, "Treasury account must not be zero address");
+			TreasuryAccount::<T>::put(account.clone());
+			TreasuryPortion::<T>::put(*portion);
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


pallets/treasury/src/lib.rs
```diff
@@ -33,30 +36,38 @@ pub mod pallet {
...
 
 	impl<T: Config> Default for GenesisConfig<T>
 	where
 		T::AccountId: From<[u8; 32]>,
 	{
+		/// Placeholder for tests/chain-spec defaults. Production must set treasury_account
+		/// explicitly.
 		fn default() -> Self {
-			Self { treasury_account: [0u8; 32].into(), treasury_portion: 50 }
+			Self { treasury_account: [1u8; 32].into(), treasury_portion: Permill::from_percent(50) }
```
ethan-crypto: I think the safest approach here would be to make missing treasury config fail fast during genesis instead of silently misconfiguring the chain with a magic nonzero treasury address. So that would mean either reverting back to the zero address or updating the type to `Option<AccountId>`,

n13:
Here's my review of **PR #434: Treasury Pallet - Review #431**.

---

## Overview

This PR changes `TreasuryPortion` storage from `u8` (0-100) to `Permill`, switches it to `OptionQuery` to enforce genesis configuration, adds zero-address rejection for the treasury account, and cleans up all downstream consumers. Good improvements overall -- a few things worth discussing.

---

## Issues

### ~~1. Missing storage migration (blocking for live chains)~~

[deleted, non issue]

### 2. Default GenesisConfig uses `[1u8; 32]` -- Ethan's concern is valid

```rust
Self { treasury_account: [1u8; 32].into(), treasury_portion: Permill::from_percent(50) }
```

This is a magic address nobody controls. If someone stands up a chain using the default config without overriding `treasury_account`, funds silently go to an unspendable address. The old zero-address default was bad too, but at least it would now be caught by the zero-address assertion.

Consider making `GenesisConfig::treasury_account` an `Option<T::AccountId>` so genesis explicitly fails if unconfigured, rather than silently using a dummy address.

### 3. Empty `integrity_test` is dead code

```rust
#[pallet::hooks]
impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
    #[cfg(test)]
    fn integrity_test() {
        // ...comment explaining why it can't validate...
    }
}
```

An empty test-only hook that does nothing should just be removed. If it can't validate anything, it doesn't need to exist.

### 4. `From<[u8; 32]> + PartialEq` bounds spread everywhere

The zero-address check requires `T::AccountId: From<[u8; 32]> + PartialEq`, which now appears on:
- `BuildGenesisConfig`
- `Pallet<T>` call impl
- `Pallet<T>` inherent impl
- `TreasuryAccountGetter`
- `TreasuryPortionGetter` (this one doesn't even touch AccountId!)
- `TreasuryProvider` impl

This leaks implementation details across every consumer. A cleaner approach might be adding a `ZeroAccount` associated type or constant to `Config`, or doing the zero-check only in `set_treasury_account` and `build()` rather than in every `account_id()` call. At minimum, `TreasuryPortionGetter` should not need `AccountId` bounds just to return a `Permill`.

### 5. Double-panic in `account_id()` is aggressive

```rust
pub fn account_id() -> T::AccountId {
    let account = TreasuryAccount::<T>::get()
        .expect("Treasury account must be set in genesis; chain is misconfigured");
    let zero: T::AccountId = [0u8; 32].into();
    assert!(account != zero, "...");
    account
}
```

Genesis `build()` and `set_treasury_account` already prevent zero addresses, so the only way this assert fires in production is if someone directly writes zero to storage (e.g., a bad migration). A panic in a getter that's called every block is risky -- if it ever triggers, it halts the chain irrecoverably. The genesis/extrinsic guards are sufficient; this runtime assertion could be a `debug_assert!` instead.

---

## Minor Nits

- `ensure!(portion <= Permill::one(), Error::<T>::InvalidPortion)` -- this is technically reachable (someone could submit a hand-crafted extrinsic with `Permill::from_parts(2_000_000)`), so keeping it is fine, but worth a note that `from_percent()` and `from_rational()` already clamp.

- The genesis `assert!(self.treasury_portion <= Permill::one(), ...)` is also always true for any `Permill` constructed through normal APIs -- the only way to violate it is manual `from_parts()` in test code.

---

## What's Good

- **`u8` to `Permill`** is the right call -- enables fractional percentages (e.g., 33.3%) and eliminates the `Permill::from_percent(u32::from(...))` boilerplate scattered through mining-rewards.
- **`OptionQuery` for both storage items** enforces that genesis must configure these values. No more silent zero-account fallback.
- **Zero-address rejection** on `set_treasury_account` prevents locking funds.
- **Test coverage** is substantially improved -- boundary tests (0%, 100%), panic-when-unconfigured tests, and TreasuryProvider trait conformance.
- **Benchmarks re-run** with correct proof sizes for the new `Permill` (4 bytes vs 1 byte).
- **Mining-rewards cleanup** is much cleaner: `Treasury::portion().mul_floor(total_reward)` reads nicely.

---

## Summary

The core type change (`u8` -> `Permill`) and the `OptionQuery` switch are solid improvements. The main concerns are: (1) confirm no migration is needed or add one, (2) reconsider the default genesis address, (3) tighten up the trait bounds so they don't bleed into unrelated getters, and (4) consider softening the runtime panics in `account_id()` to avoid irrecoverable chain halts.

ethan-crypto (approved):
LGTM

PR #510, opened 2026-04-10
fix: correct planck genesis endowment and preset logging

n13:
## Summary

- **Fix planck endowment bug**: `planck_config_genesis()` was calling `endowed_accounts.extend(planck_treasury_account())` which extended with the single multisig AccountId instead of the individual treasury signers. Changed to `endowed_accounts.extend(planck_treasury_signers())` so the three actual signers are endowed.
- **Add shared `log_genesis_accounts` helper**: All three presets (dev, heisenberg, planck) now use a single DRY function that logs endowed accounts, treasury address, treasury signers, and tech collective members on startup — regardless of how the chain spec was loaded.

## Context

Running `--chain planck_live_spec` produced wrong endowed accounts (3 bogus addresses from byte iteration of the multisig AccountId) and the tech collective was correctly seeded but endowed accounts did not match.

## Note

This PR does **not** include a regenerated `planck.json` chain spec. That will be done separately after a new runtime release is cut, using `scripts/genesis_generate_spec.sh` with the srtool-built WASM.

## Test plan

- [x] `cargo check -p quantus-runtime` passes
- [x] `build-spec --chain planck_live_spec --raw` produces correct accounts
- [x] CI passes
- [x] `cargo test -p quantus-runtime` passes

### DIFF
--- a/runtime/src/genesis_config_presets.rs
+++ b/runtime/src/genesis_config_presets.rs
@@ -196,23 +196,52 @@ fn genesis_template(
 	v
 }
 
-/// Return the development genesis config.
-pub fn development_config_genesis() -> Value {
-	let mut endowed_accounts = vec![];
-	endowed_accounts.extend(dilithium_default_accounts());
-
-	// Add the test address derived from TEST_WORMHOLE_SECRET.
-	// This is useful for testing ZK spending with a known secret.
-	let test_account = test_wormhole_account();
-	endowed_accounts.push(test_account.clone());
-
-	let ss58_version = sp_core::crypto::Ss58AddressFormat::custom(189);
-	for account in endowed_accounts.iter() {
-		log::info!("🍆 Endowed account: {:?}", account.to_ss58check_with_version(ss58_version));
+fn log_genesis_accounts(
+	preset: &str,
+	endowed: &[AccountId],
+	treasury_account: &AccountId,
+	treasury_signers: &[AccountId],
+	tech_collective: &[AccountId],
+) {
+	let ss58 = ss58_version();
+	for account in endowed {
+		log::info!("[{preset}] 💰 Endowed: {:?}", account.to_ss58check_with_version(ss58));
 	}
 	log::info!(
-		"🕳️  Test ZK address (use TEST_WORMHOLE_SECRET to spend): {:?}",
-		test_account.to_ss58check_with_version(ss58_version)
+		"[{preset}] 🏦 Treasury: {:?}",
+		treasury_account.to_ss58check_with_version(ss58)
+	);
+	for signer in treasury_signers {
+		log::info!(
+			"[{preset}] 🔑 Treasury signer: {:?}",
+			signer.to_ss58check_with_version(ss58)
+		);
+	}
+	for member in tech_collective {
+		log::info!(
+			"[{preset}] 🏛️  Tech collective: {:?}",
+			member.to_ss58check_with_version(ss58)
+		);
+	}
+}
+
+/// Return the development genesis config.
+pub fn development_config_genesis() -> Value {
+	let mut endowed_accounts = dilithium_default_accounts();
+	let test_account = test_wormhole_account();
+	endowed_accounts.push(test_account.clone());
+	let treasury_account = development_treasury_account();
+	let tech_collective = development_tech_collective_seed();
+	log_genesis_accounts(
+		"dev",
+		&endowed_accounts,
+		&treasury_account,
+		&dilithium_default_accounts(),
+		&tech_collective,
+	);
+	log::info!(
+		"[dev] 🕳️  Test ZK: {:?}",
+		test_account.to_ss58check_with_version(ss58_version())
 	);
 
 	#[cfg(feature = "runtime-benchmarks")]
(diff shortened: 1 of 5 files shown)


n13:
Found two review items in [PR #510](https://github.com/Quantus-Network/chain/pull/510), ordered by severity.

- **Medium — “consistent logging regardless of chain spec load path” is not actually achieved**
  
  `log_genesis_accounts` is only invoked inside runtime preset builders. When operators run `--chain planck` / `--chain heisenberg`, node loads embedded JSON specs and bypasses those preset functions, so the new logs do not appear.
  
  ```215:228:node/src/command.rs
  	fn load_spec(&self, id: &str) -> Result<Box<dyn sc_service::ChainSpec>, String> {
  		Ok(match id {
  			"dev" =>
  				Box::new(chain_spec::development_chain_spec()?) as Box<dyn sc_service::ChainSpec>,
  			"heisenberg_live_spec" =>
  				Box::new(chain_spec::heisenberg_chain_spec()?) as Box<dyn sc_service::ChainSpec>,
  			"" | "heisenberg" => Box::new(chain_spec::ChainSpec::from_json_bytes(include_bytes!(
  				"chain-specs/heisenberg.json"
  			))?) as Box<dyn sc_service::ChainSpec>,
  			"planck_live_spec" =>
  				Box::new(chain_spec::planck_chain_spec()?) as Box<dyn sc_service::ChainSpec>,
  			"planck" => Box::new(chain_spec::ChainSpec::from_json_bytes(include_bytes!(
  				"chain-specs/planck.json"
  			))?) as Box<dyn sc_service::ChainSpec>,
  ```
  
  ```199:217:runtime/src/genesis_config_presets.rs
  fn log_genesis_accounts(
  	preset: &str,
  	endowed: &[AccountId],
  	treasury_account: &AccountId,
  	treasury_signers: &[AccountId],
  	tech_collective: &[AccountId],
  ) {
  	let ss58 = ss58_version();
  	for account in endowed {
  		log::info!("[{preset}] 💰 Endowed: {:?}", account.to_ss58check_with_version(ss58));
  	}
  	log::info!("[{preset}] 🏦 Treasury: {:?}", treasury_account.to_ss58check_with_version(ss58));
  	for signer in treasury_signers {
  		log::info!("[{preset}] 🔑 Treasury signer: {:?}", signer.to_ss58check_with_version(ss58));
  	}
  	for member in tech_collective {
  		log::info!("[{preset}] 🏛️  Tech collective: {:?}", member.to_ss58check_with_version(ss58));
  	}
  }
  ```

- **Low — Added tests don’t cover the actual Planck endowment regression path**
  
  The bug fix is in `planck_config_genesis` endowment composition, but new tests only exercise `seed_tech_collective` voting behavior. Good tests, but they don’t guard against the specific endowment issue regressing.
  
  ```1286:1298:runtime/tests/governance/tech_collective.rs
  	#[test]
  	fn test_genesis_seeded_members_can_vote() {
  		TestCommons::new_fast_governance_test_ext().execute_with(|| {
  			let proposer = TestCommons::account_id(1);
  			let voter = TestCommons::account_id(2);
  
  			Balances::make_free_balance_be(&proposer, 3000 * UNIT);
  			Balances::make_free_balance_be(&voter, 2000 * UNIT);
  
  			quantus_runtime::genesis_config_presets::seed_tech_collective(&[
  				proposer.clone(),
  				voter.clone(),
  			]);
  ```

No other functional regressions stood out in the changed code.

Validation I ran locally:
- `cargo check -p quantus-runtime` ✅
- `cargo test -p quantus-runtime seeded_member -- --nocapture` ✅
- `cargo test -p quantus-runtime governance::tech_collective:: -- --nocapture` ✅
- `build-spec --chain planck_live_spec --raw` shows correct Planck signers in endowed logs ✅
- `build-spec --chain planck --raw` does **not** show those preset logs (supports finding #1) ✅

runtime/src/genesis_config_presets.rs
```diff
@@ -347,14 +362,21 @@ pub fn seed_tech_collective(members: &[AccountId]) {
...
 pub fn planck_config_genesis() -> Value {
+	let treasury_signers = planck_treasury_signers();
+	let tech_collective = planck_tech_collective_seed();
+	let treasury_account = planck_treasury_account();
 	let mut endowed_accounts = vec![planck_faucet_account()];
-	endowed_accounts.extend(dilithium_default_accounts());
-	for account in endowed_accounts.iter() {
-		log::info!("🍆 Endowed account: {:?}", account.to_ss58check_with_version(ss58_version()));
-	}
+	endowed_accounts.extend(treasury_signers.clone());
```
czareko: We don't have to endow multisig members

n13: They need initial funds in order to be able to set up the multisig for the treasury

They'd only need a tiny amount but it seems endowed is hard-coded, didn't want to change this. 

What would be a better option? Rely on faucet?

n13: Also on mainnet we don't have faucet so I think that is not a solution. It needs to work on mainnet too. 

So maybe we should give them 1 coin each?

n13:
Note
test tests::force_hashed_values_preserve_distinct_zero_prefixed_values_poseidon ... FAILED

the test fail is intentional. the trie db remove PR fixes this #514 

This test was added to exercise the problem we found when trying to use tech collective voting and is_member

The test basically is the core problem of having non-injective hashing on user provided values

We fix this with a new architecture in #514 and  https://github.com/Quantus-Network/chain/pull/515

runtime/src/genesis_config_presets.rs
n13: Ok fixed - adding 1 coin each

czareko (approved):

PR #512, opened 2026-04-10
fix: injective hash for trie value nodes (prevent Poseidon hash collisions)

n13:
## Summary

- **Fix non-injective value hashing in the trie**: `PoseidonHasher::hash()` uses `bytes_to_felts_compact`, which is non-injective for variable-length data — distinct byte sequences can produce identical hashes. With `MAX_INLINE_VALUE = 0` (all values hashed externally), this can silently corrupt storage by making different values look identical in the trie.
- **Introduce `TrieLayout::hash_value()`** — a new trait method that separates *value* hashing from *node* hashing. The default falls back to `H::hash()`, but `LayoutV0`/`LayoutV1` override it to use `qp_poseidon_core::hash_bytes`, which is injective (length-prefixed encoding). Trie node hashing remains unchanged (compact, ZK-friendly).
- **Harden `MemoryDB` against null-node collisions**: the wrapper now includes a `forced` fallback map so values matching `memory-db`'s internal `null_node_data` sentinel are never silently dropped on `emplace`.

### 1. `TrieLayout::hash_value()` — new trait method (`trie-db/src/lib.rs`)

Added to the `TrieLayout` trait with a default implementation (`H::hash(value)`). This gives each layout the ability to override how *values* are hashed without affecting how *trie nodes* are hashed. The default preserves backward compatibility for any layout that doesn't override it.

### 2. All value-hashing call sites updated (`trie-db`)

| File | Change |
|------|--------|
| `triedbmut.rs` (2 sites) | `NodeToEncode::Node` handling now calls `L::hash_value(value)` + `db.emplace()` instead of `db.insert()` |
| `iter_build.rs` (4 sites) | All `process_inner_hashed_value` implementations (`TrieBuilder`, `TrieRoot`, `TrieRootPrint`, `TrieRootUnhashed`) call `T::hash_value()` |
| `node.rs` | `Value::to_owned_value` uses `L::hash_value()` for cache consistency |

### 3. `LayoutV0` / `LayoutV1` override (`sp-trie/src/lib.rs`)

Both layouts override `hash_value()` to call `injective_value_hash::<H>()`, which uses `qp_poseidon_core::hash_bytes` — an injective hash function that length-prefixes the input before hashing into field elements.

The `trie_root` / `trie_root_unhashed` implementations were also updated to use `trie_db::trie_visit` (which respects `L::hash_value()`) instead of the external `trie-root` crate functions (which hardcode `H::hash()`).

### 4. `MemoryDB` wrapper hardened (`sp-trie/src/lib.rs`)

`memory_db::MemoryDB::emplace()` silently drops values that exactly match its `null_node_data` sentinel (default: `[0u8; 8]`). When `L::hash_value()` produces a key different from `H::hash()`, the inner DB's `contains` check can fail even after a successful `emplace`. The wrapper now stores such values in a `forced: HashMap<H::Out, DBValue>` fallback that `get`/`contains` check as a last resort.

### 5. `StorageProof` dual-insertion (`sp-trie/src/storage_proof.rs`)

When deserializing a `StorageProof` into a `MemoryDB`, each node is now inserted under *both* `H::hash(node)` (for standard node lookups) and `injective_value_hash(node)` (for value lookups). This ensures reconstructed DBs can satisfy both access patterns.

### Modified test assertions

| File | Test | Change |
|------|------|--------|
| `recorder.rs` | `recorder_transaction_accessed_keys_works` | `Blake2Hasher::hash(value)` assertions changed to `Layout::hash_value(value)` — the trie now stores the injective hash, so `get_hash()` returns the new hash. Added `use trie_db::TrieLayout` import. |
| `lib.rs` | `test_realistic_chain_storage` | Switched from raw `MemoryDBMeta` to the `MemoryDB` wrapper — this test inserts `vec![0x00; 8]` which collides with `null_node_data`, so it needs the `forced` fallback. |

### Test-only layout (`ForceHashedValuesLayoutV1`)

Also overrides `hash_value()` with `injective_value_hash` and removes its custom `trie_root`/`trie_root_unhashed` overrides (now inherits defaults that correctly use `trie_visit`).

### Test results

- **sp-trie**: 61 passed, 0 failed, 1 ignored
- **trie-db**: 14 passed, 0 failed

## Known remaining work

- **sp-state-machine**: 6 tests fail due to hardcoded expected root hashes and proof sizes that changed with the new value hash. These need their expected values updated (the logic is correct, only the golden values are stale).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11901,6 +11901,7 @@ dependencies = [
  "parity-scale-codec",
  "parking_lot 0.12.4",
  "qp-poseidon",
+ "qp-poseidon-core",
  "rand 0.8.5",
  "scale-info",
  "schnellru",
--- a/primitives/trie-db/src/iter_build.rs
+++ b/primitives/trie-db/src/iter_build.rs
@@ -388,7 +388,9 @@ where
 	}
 
 	fn process_inner_hashed_value(&mut self, prefix: Prefix, value: &[u8]) -> TrieHash<T> {
-		self.db.insert(prefix, value)
+		let hash = T::hash_value(value);
+		self.db.emplace(hash, prefix, value.to_vec().into());
+		hash
 	}
 }
 
@@ -420,7 +422,7 @@ impl<T: TrieLayout> ProcessEncodedNode<TrieHash<T>> for TrieRoot<T> {
 	}
 
 	fn process_inner_hashed_value(&mut self, _prefix: Prefix, value: &[u8]) -> TrieHash<T> {
-		<T::Hash as Hasher>::hash(value)
+		T::hash_value(value)
 	}
 }
 
@@ -474,7 +476,7 @@ impl<T: TrieLayout> ProcessEncodedNode<TrieHash<T>> for TrieRootPrint<T> {
 
 	fn process_inner_hashed_value(&mut self, _prefix: Prefix, value: &[u8]) -> TrieHash<T> {
 		println!("Hashed node: {:x?}", &value);
-		<T::Hash as Hasher>::hash(value)
+		T::hash_value(value)
 	}
 }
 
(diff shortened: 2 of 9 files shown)


n13:
No longer doing this - replaced by new architecture in #515

PR #515, opened 2026-04-11
Zk-trie implemented

illuzen:
Summary

Introduces pallet-zk-trie, a 4-ary Poseidon Merkle tree for storing ZK transfer proofs, replacing the redundant TransferProof storage in pallet-wormhole.

Key changes:

New pallets/zk-trie/ with cumulative Merkle tree using injective Poseidon (4 bytes/felt) for leaves and non-injective (8 bytes/felt) for internal nodes
ZkTrieRecorder trait for external pallets to insert leaves
Runtime API (ZkTrieApi) and RPC methods (zkTrie_getState, zkTrie_getMerkleProof) returning raw bytes for direct ZK circuit consumption
Tree root stored in block digest via DigestItem::Other
Removed TransferProof, TransferProofKey, TransferProofData, and LeafInputsHash from wormhole pallet


n13:
Summary:

1. We should propagate a ZK trie insert error to the caller of this method, and it should probably fail the entire transaction otherwise users could end with unspendable wormholes?

It should never error of course- but if it does it should be fatal.

2. Why is "from" dropped from leaf data?


The rest is cleanup stuff... 

---

## PR #515 Review: Zk-trie implemented

**Summary:** Introduces `pallet-zk-trie`, a 4-ary Poseidon Merkle tree for ZK transfer proofs, replacing the flat `TransferProof` storage map in wormhole. Adds Runtime API + RPC endpoints. +1392 / -137 across 18 files.

The overall design is solid -- a separate pallet with a clean `ZkTrieRecorder` trait for cross-pallet communication follows existing project patterns (`TransferProofRecorder`, `TreasuryProvider`, etc.). The tree implementation, proof generation, and verification logic are well-structured. That said, there are several issues worth addressing before merge.

---

### Critical / High

**1. `from` address dropped from leaf data**

The old `TransferProofData` was `(asset_id, count, from, to, amount)` -- five fields. The new `ZkLeaf` is `(to, transfer_count, asset_id, amount)` -- four fields, with `from` removed. This means the ZK circuit can no longer verify the sender. If that's intentional (e.g., the circuit doesn't need to prove sender identity), it should be documented. If not, this is a data loss regression.

**2. Silent failure on trie insertion**

In wormhole's new `record_transfer`, the return value of `T::ZkTrie::record_transfer()` is discarded:

```596:620:pallets/wormhole/src/lib.rs (PR diff)
// Insert into ZK trie for Merkle proof generation
T::ZkTrie::record_transfer(
    to.clone().into(),
    transfer_count_u64,
    asset_id.clone(),
    amount,
);
```

If `insert_leaf` fails (e.g., `MaxDepthReached`), the transfer still completes and events still fire, but no proof is recorded. The user has tokens they can never prove ownership of via ZK. This should at minimum log a warning, or ideally propagate the error.

**3. `DigestItem::Other` is ambiguous**

```815:819:pallets/zk-trie/src/lib.rs (PR diff)
let root = Root::<T>::get();
let item = sp_runtime::generic::DigestItem::Other(root.to_vec());
<frame_system::Pallet<T>>::deposit_log(item);
```

`DigestItem::Other` with a raw 32-byte payload is indistinguishable from any other pallet that does the same. The existing codebase only uses `Other` in tests. Consider using `DigestItem::Consensus` or `DigestItem::PreRuntime` with a 4-byte engine ID (e.g., `*b"zktr"`), or at least prefix the payload so consumers can identify it.

---

### Medium

**4. Duplicate tests (DRY violation)**

`test_capacity_at_depth` and `test_hash_node` appear identically in both `pallets/zk-trie/src/tests.rs` and the `#[cfg(test)] mod tests` block at the bottom of `pallets/zk-trie/src/tree.rs`. One copy should be removed.

**5. Unused computed variables**

In `tree.rs` `update_path`:

```rust
let _position_in_parent = (current_index % (ARITY as u64)) as usize;
```

And in `verify_proof`:

```rust
let mut _current_index = proof.leaf_index;
// ...
_current_index /= ARITY as u64;
```

Both are computed but never used. The `_` prefix suppresses the warning but the computation is wasted. Remove them.

**6. TransferCount-to-u64 conversion is fragile**

```596:608:pallets/wormhole/src/lib.rs (PR diff)
let transfer_count_u64: u64 = {
    let encoded = current_count.encode();
    let mut bytes = [0u8; 8];
    let len = encoded.len().min(8);
    bytes[..len].copy_from_slice(&encoded[..len]);
    u64::from_le_bytes(bytes)
};
```

This manually SCALE-encodes then reinterprets bytes. It works because `TransferCount` is `u64`, but if the type ever changes (e.g., `u128`), it silently truncates. A `TryInto<u64>` or `Into<u64>` bound on `Config::TransferCount` would be safer and cleaner.

**7. No weight info**

`pallet-zk-trie` has no `WeightInfo` trait. `insert_leaf` does O(depth) storage reads and writes, and `get_merkle_proof` does O(depth * ARITY) reads. Without benchmarked weights, there's no fee accounting for the computational cost. The `Cargo.toml` has `runtime-benchmarks` feature but no actual benchmarks.

**8. Leaf hash recomputation on sibling reads**

`get_leaf_hash` in `tree.rs` reads the leaf from storage and re-hashes it every time:

```1558:1566:pallets/zk-trie/src/tree.rs (PR diff)
fn get_leaf_hash<T: Config>(index: u64) -> Hash256
where
    AccountIdOf<T>: AsRef<[u8]>,
{
    match crate::Leaves::<T>::get(index) {
        Some(leaf) => hash_leaf::<T>(&leaf),
        None => empty_hash(),
    }
}
```

During `update_path` and `generate_proof`, this is called for each sibling at the leaf level (up to 3 per level). Consider storing leaf hashes alongside leaf data, or caching them at level 0 in `Nodes`, to avoid redundant Poseidon computations.

---

### Low / Nits

**9. No storage migration for removed `TransferProof`**

The `TransferProof` storage map is removed from wormhole. If the chain has been running, this leaves orphaned storage. A migration to clean it up (or at least a note about it) would be good hygiene.

**10. Magic error codes in RPC**

Error codes 9000-9003 in `zktrie_rpc.rs` are magic numbers. Consider defining them as constants.

**11. `ZkMerkleProofRpc.leaf_data` is opaque**

The RPC returns `leaf_data: Vec<u8>` which is SCALE-encoded `ZkLeaf`. RPC consumers (likely JS/TS) need to know the encoding schema. Consider adding a comment noting the encoding, or returning structured fields instead.

---

### CI Status

The "Fast Checks (Format)" job is currently failing, with "Build & Test Matrix" and "Analysis (Clippy & Doc)" skipped as a result. The format check needs to pass before the other checks can run.

---

### Summary Table

| Severity | Issue | Actionable? |
|----------|-------|-------------|
| High | `from` dropped from leaf -- intentional? | Clarify / restore |
| High | Silent ZK trie insertion failure | Log or propagate error |
| High | `DigestItem::Other` ambiguity | Use engine ID |
| Medium | Duplicate tests in tree.rs & tests.rs | Remove one copy |
| Medium | Unused `_position_in_parent`, `_current_index` | Remove |
| Medium | Fragile `TransferCount` -> `u64` conversion | Add trait bound |
| Medium | No weights / benchmarks | Add benchmarks |
| Medium | Leaf hash recomputed on every sibling read | Cache leaf hashes |
| Low | No migration for removed storage | Add migration |
| Low | Magic RPC error codes | Use constants |
| Low | Opaque `leaf_data` in RPC response | Document encoding |

Overall a well-structured PR with clean separation of concerns. Addressing the high-severity items (especially the silent failure and missing `from` field) before merge would make this much more robust.

n13:
Findings for [PR #515](https://github.com/Quantus-Network/chain/pull/515), ordered by severity:

- **High — `quantus-runtime` test target no longer compiles**
  - `runtime/src/transaction_extensions.rs` still calls `Wormhole::transfer_proof(...)` in multiple tests, but `TransferProof` was removed from `pallet-wormhole` in this PR.
  - Repro: `cargo test -p quantus-runtime transaction_extensions::tests::event_based_proof_recording_native_transfer -- --nocapture`
  - Result: `E0599` (`no function or associated item named transfer_proof`) at 7 call sites in `runtime/src/transaction_extensions.rs`.

- **High — ZK trie insertion failures are silently swallowed**
  - In `pallets/wormhole/src/lib.rs`, `record_transfer` increments `TransferCount` and then calls `T::ZkTrie::record_transfer(...)` without checking the return value.
  - In `pallets/zk-trie/src/lib.rs`, `ZkTrieRecorder for Pallet<T>` returns `Self::insert_leaf(...).ok()`, converting real errors (e.g. `MaxDepthReached`) into `None`.
  - Impact: transfer/event can succeed while no leaf is inserted, leaving users without a usable Merkle proof.

- **Medium — Weight accounting is stale for proof recording extension**
  - `runtime/src/transaction_extensions.rs` still prices proof recording as `reads_writes(n, 2*n)` with comment `TransferProof + TransferCount`.
  - After this PR, recording uses `zk-trie` path updates (`Leaves`, `LeafCount`, `Depth`, `Nodes`, `Root`, hashing), so actual cost is significantly higher and depth-dependent.
  - This can undercharge and distort block weight/fee economics.

- **Medium — Sender field dropped from merkle leaf data**
  - `pallets/zk-trie/src/lib.rs` defines `ZkLeaf` as `(to, transfer_count, asset_id, amount)`.
  - `pallets/wormhole/src/lib.rs` `record_transfer` still receives `from`, but does not include it in trie insertion.
  - If sender binding is required by downstream ZK verification assumptions, this is a semantic regression (at minimum needs explicit documentation/tests asserting sender is intentionally excluded).

- **Medium — `TransferCount` → `u64` conversion is fragile**
  - In `pallets/wormhole/src/lib.rs`, conversion uses SCALE-encode + first 8 bytes copy.
  - If `Config::TransferCount` changes from `u64` later, truncation/collision risk is silent.
  - Safer to enforce a direct `TryInto<u64>`/`Into<u64>` bound and fail loudly on incompatible types.

### Validation I ran
- `cargo test -p pallet-zk-trie` ✅
- `cargo test -p pallet-wormhole` ✅
- `cargo test -p quantus-runtime transaction_extensions::tests::event_based_proof_recording_native_transfer -- --nocapture` ❌ (compile errors above)

n13 (approved):
LGTM

Issue #3, opened 2025-05-08
Runtime Upgrades - Tech community with levels based on evidence


czareko:
It will be similar to Polkadot Tech Fellowship, but simplified to 5 levels.

PR #517, opened 2026-04-13
Scheduler + Multisig Audit 2

illuzen:
Addressed the following audit items.

**Scheduler: Reschedule drops retry state**

Both `do_reschedule` and `do_reschedule_named` move a task to a new address but never transfer its `Retries` entry:

```rust
/* pallets/scheduler/src/lib.rs:724-743 */
fn do_reschedule(
    (when, index): TaskAddressOf<T>,
    new_time: DispatchTime<BlockNumberFor<T>, T::Moment>,
) -> Result<TaskAddressOf<T>, DispatchError> {
    let new_time = Self::resolve_time(new_time)?;
    if new_time == when {
        return Err(Error::<T>::RescheduleNoChange.into());
    }
    let task = Agenda::<T>::try_mutate(when, |agenda| {
        let task = agenda.get_mut(index as usize).ok_or(Error::<T>::NotFound)?;
        ensure!(!matches!(task, Some(Scheduled { maybe_id: Some(_), .. })), Error::<T>::Named);
        task.take().ok_or(Error::<T>::NotFound)
    })?;
    Self::cleanup_agenda(when);
    Self::deposit_event(Event::Canceled { when, index });
    Self::place_task(new_time, task).map_err(|x| x.0)
}
```

```rust
/* pallets/scheduler/src/lib.rs:792-812 */
fn do_reschedule_named(
    id: TaskName,
    new_time: DispatchTime<BlockNumberFor<T>, T::Moment>,
) -> Result<TaskAddressOf<T>, DispatchError> {
    let new_time = Self::resolve_time(new_time)?;
    let lookup = Lookup::<T>::get(id);
    let (when, index) = lookup.ok_or(Error::<T>::NotFound)?;
    if new_time == when {
        return Err(Error::<T>::RescheduleNoChange.into());
    }
    let task = Agenda::<T>::try_mutate(when, |agenda| {
        let task = agenda.get_mut(index as usize).ok_or(Error::<T>::NotFound)?;
        task.take().ok_or(Error::<T>::NotFound)
    })?;
    Self::cleanup_agenda(when);
    Self::deposit_event(Event::Canceled { when, index });
    Self::place_task(new_time, task).map_err(|x| x.0)
}
```

The config stays keyed at the old address, so `service_task` finds nothing when it looks up retries at the new address.
If the task fails when it is executed, no retries are ever attempted.

Suggested fix: In both functions, after placing the task at the new address, take the old retry config and re-insert
it at the new address.

```rust
let new_address = Self::place_task(new_time, task).map_err(|x| x.0)?;
if let Some(retry_config) = Retries::<T>::take((when, index)) {
    Retries::<T>::insert(new_address, retry_config);
}
Ok(new_address)
```

**Scheduler: False permanently overweight tasks**

`service_block_agendas` and `service_timestamp_agendas` each have their own local `executed` counter:

```rust
/* pallets/scheduler/src/lib.rs:844,898 */
let mut executed = 0;
```

This counter is used to determine `is_first`, which controls whether an overweight task is considered permanently
overweight (and deleted) or just temporarily overweight (and kept for retry):

```rust
/* pallets/scheduler/src/lib.rs:1077-1086 */
match Self::execute_dispatch(weight, task.origin.clone(), call) {
    Err(()) if is_first => {
        T::Preimages::drop(&task.call);
        Self::deposit_event(Event::PermanentlyOverweight {
            task: (when, agenda_index),
            id: task.maybe_id,
        });
        Err((Unavailable, None))
    },
    Err(()) => Err((Overweight, Some(task))),
```

This logic is present and correct in the upstream scheduler, but due to the split between block and timestamp agendas,
*without* sharing the `executed` counter across the two domains, a task can be misclassified as permanently overweight.
If block tasks take up 80% of the block time, and a large timestamp task is first in the queue that would take up 30%
of the block time, it *would* fit in the next block fine, but is instead removed as permanently overweight.

Suggested fix: Share the `executed` counter across the two domains, replace the local variables with mutable reference
parameters in both functions, matching `service_agenda`.

**Scheduler: Timestamp tasks can fire early**

`DispatchTime::After` with `BlockNumberOrTimestamp::Timestamp` is not actually an *after* time:

```rust
/* pallets/scheduler/src/lib.rs:591-612 */
fn resolve_time(
    when: DispatchTime<BlockNumberFor<T>, T::Moment>,
) -> Result<BlockNumberOrTimestampOf<T>, DispatchError> {
    // [...]
    let when = match when {
        DispatchTime::After(x) => {
            match x {
                BlockNumberOrTimestamp::Timestamp(_) => {
                    x.normalize(T::TimestampBucketSize::get())
                }
            };
        }
    };
```

Both the target timestamp and the current time are normalized to bucket values before comparison. This means the
execution condition is `floor(now / bucket_size) >= floor(target / bucket_size)`. With the current settings:

```rust
/* runtime/src/lib.rs:89-90 */
pub const TARGET_BLOCK_TIME_MS: u64 = 12_000;
```

```rust
/* runtime/src/configs/mod.rs:144 */
pub const TimestampBucketSize: u64 = 2 * TARGET_BLOCK_TIME_MS; // Nyquist frequency
```

This means a task can be executed as much as 24 seconds earlier than expected, *before* accounting for any block time
variance. To restore precise not-before-time semantics, the exact timestamp must be kept and checked instead of the
normalized bucket value.

Also, this is an incorrect use of the term Nyquist frequency as this is a duration, not a frequency.

**Multisig: claim_deposits worst-case weight undercharges**

```rust
/* pallets/multisig/src/lib.rs:918-922 */
#[pallet::weight(<T as Config>::WeightInfo::claim_deposits(
    T::MaxTotalProposalsInStorage::get(),  // Worst-case iterated
    T::MaxTotalProposalsInStorage::get().saturating_div(2),  // Worst-case cleaned
    T::MaxCallSize::get()  // Worst-case avg call size
))]
```

There is no justification for the division by 2 for the number of cleaned proposals. The worst-case for cleaned
proposals should be equal to the worst-case for iterated proposals. The benchmark also shows that `iterated == cleaned`
is a valid scenario. On top of that, the benchmark's `min(r, i)` clamp flattens out the `r` coefficient across the
entire range, undercharging when `r` is large.

Suggested fix: First, fix the weight annotation to use the actual worst-case:

```rust
#[pallet::weight(<T as Config>::WeightInfo::claim_deposits(
    T::MaxTotalProposalsInStorage::get(),  // Worst-case iterated
    T::MaxTotalProposalsInStorage::get(),  // Worst-case cleaned
    T::MaxCallSize::get()  // Worst-case avg call size
))]
```

One possible improvement for the benchmark would be to restrict the sampled benchmark region so that `r <= i` is always
true, which avoids the `min(r, i)` clamp in the sampled points. For example:

```rust
/* pallets/multisig/src/benchmarking.rs:385-392 */
#[benchmark]
fn claim_deposits(
    i: Linear<
        { T::MaxTotalProposalsInStorage::get() / 2 },
        { T::MaxTotalProposalsInStorage::get() },
    >,
    r: Linear<0, { T::MaxTotalProposalsInStorage::get() / 2 }>,
    c: Linear<0, { T::MaxCallSize::get().saturating_sub(100) }>,
```

This is not a complete fix. It avoids the clamp distortion in the sampled region, but it does not cover the full valid
domain, especially small `i` values and cases where both `i` and `r` are larger than `MaxTotalProposalsInStorage / 2`.
Without Substrate support for dependent variables, there is no way to express the full domain in one benchmark.


n13:
Good, I have a thorough picture now. Here's my review of PR #517.

---

## PR #517 Review: Scheduler + Multisig Audit 2

Overall this is a solid set of audit fixes. The four issues are real, the fixes are correctly targeted, and the test coverage is thorough. A few items worth flagging:

---

### 1. Reschedule preserves retry config -- Looks good

The fix in `do_reschedule` and `do_reschedule_named` is clean and correct:

```742:749:pallets/scheduler/src/lib.rs
		let new_address = Self::place_task(new_time, task).map_err(|x| x.0)?;
		// Transfer retry configuration to the new address
		if let Some(retry_config) = Retries::<T>::take((when, index)) {
			Retries::<T>::insert(new_address, retry_config);
		}
		Ok(new_address)
```

`Retries::take` atomically reads-and-removes from the old key, and the insert places it at the new address. The two tests (`reschedule_preserves_retry_config` and `reschedule_named_preserves_retry_config`) both verify the full lifecycle: config transfers, the failed task retries, and eventually succeeds. No issues here.

---

### 2. Shared `executed` counter -- Looks good

Hoisting `executed` into `on_initialize` and threading `&mut u32` through both `service_block_agendas` and `service_timestamp_agendas` is the right fix. The `is_first` check at line 994 (`*executed == 0`) now correctly reflects whether *any* task has run in this block, regardless of domain.

```994:994:pallets/scheduler/src/lib.rs
			let result = Self::service_task(weight, now, when, agenda_index, *executed == 0, task);
```

The test `timestamp_task_not_permanently_overweight_after_block_tasks_execute` exercises the exact scenario from the audit: a block task consuming ~67% of weight followed by a timestamp task needing ~50%. Good.

---

### 3. Timestamp early-fire fix -- Has a concern

The fix adds an extra bucket bump after normalization:

```611:633:pallets/scheduler/src/lib.rs
					BlockNumberOrTimestamp::Timestamp(target) => {
						let bucket = x.normalize(T::TimestampBucketSize::get());
						let bucket_time = bucket.as_timestamp().unwrap_or(target);
						BlockNumberOrTimestamp::Timestamp(
							bucket_time.saturating_add(T::TimestampBucketSize::get()),
						)
					},
```

The `normalize` function already rounds *up* to the next bucket boundary (e.g., `normalize(15000, 10000) = 20000`). Adding another full `bucket_size` on top means the task is now placed **two** buckets ahead:

- Target = 15000, bucket_size = 10000
- `normalize(15000)` = 20000 (already the *end* of the bucket containing 15000)
- +10000 = 30000

So a task requested "after 15000ms" won't fire until time >= 20001 (when bucket 30000 becomes serviceable). That's potentially **5 seconds** later than needed with the current 10s bucket. The original bug was that it could fire too early, but this overcompensates. It would be more precise to do:

```rust
BlockNumberOrTimestamp::Timestamp(target) => {
    let bucket = BlockNumberOrTimestamp::Timestamp(target)
        .normalize(T::TimestampBucketSize::get());
    let bucket_time = bucket.as_timestamp().unwrap_or(target);
    if bucket_time <= target {
        // Only bump if the bucket boundary is at or before the target
        BlockNumberOrTimestamp::Timestamp(
            bucket_time.saturating_add(T::TimestampBucketSize::get()),
        )
    } else {
        bucket
    }
}
```

That said, the current approach is *safe* (tasks never fire early, which is the critical invariant). Whether the extra latency is acceptable is a product decision. The tests pass because they're written against the new behavior. Just worth acknowledging this is a conservative fix that adds up to one extra bucket of delay.

Also -- `DispatchTime::At` with a timestamp still goes through the raw `BlockNumberOrTimestamp::BlockNumber(x)` path (line 598), so `schedule_at` with a timestamp isn't affected. Only `schedule_after` with a timestamp gets this treatment. That asymmetry is fine since `At` is an absolute time and `After` is relative, but it's worth documenting somewhere that `At(Timestamp(x))` is actually treated as `At(BlockNumber(x))` -- that looks like a pre-existing issue unrelated to this PR.

---

### 4. Multisig `claim_deposits` weight fix -- Looks good, minor note

Changing the weight annotation from `saturating_div(2)` to the full `MaxTotalProposalsInStorage` is correct for worst-case charging:

```918:922:pallets/multisig/src/lib.rs
		#[pallet::weight(<T as Config>::WeightInfo::claim_deposits(
		T::MaxTotalProposalsInStorage::get(),  // Worst-case iterated
		T::MaxTotalProposalsInStorage::get(),  // Worst-case cleaned
		T::MaxCallSize::get()  // Worst-case avg call size
	))]
```

The benchmark range restriction (`i` in upper half, `r` in lower half) avoids the `min(r,i)` clamp distortion during regression fitting. The PR description is upfront that this doesn't cover the full domain -- that's acceptable given Substrate's limitations.

One minor thing: the `r` range now starts at `0` instead of `1`. Since the benchmark body does `(r as u32).min(i)`, `r = 0` means `cleaned_target = 0` and no proposals are cleaned. That's a valid degenerate case to sample, so this is fine -- just confirming it was intentional.

---

### Summary

| Fix | Verdict | Notes |
|-----|---------|-------|
| Reschedule retry transfer | Approve | Clean, correct, well-tested |
| Shared `executed` counter | Approve | Exact fix for the audit finding |
| Timestamp early-fire | Approve with note | Correct but conservative -- adds up to one extra bucket of latency beyond what's strictly needed |
| `claim_deposits` weight | Approve | Proper worst-case charging |

The PR is merge-ready. The only thing I'd flag for discussion is whether the extra bucket delay in fix #3 is acceptable, or if a tighter conditional bump would be preferred.

n13 (approved):
LGTM

czareko (approved):

Issue #483, opened 2026-04-08
Scheduler fixes


n13:
Need to find elegant fix for this - the scheduler is complex and fixing a tiny minor kinda non-bug might introduce real bugs.

To evaluate

Scheduler: fix Lookup/Agenda consistency for deferred named tasks

When service_task cannot execute a named task (preimage unavailable or temporarily overweight), the task returns to the Agenda for future resolution. Previously the Lookup entry was removed eagerly before the outcome was known, leaving the task unreachable by name -- cancel_named and reschedule_named would both fail with NotFound.

This is a concrete problem for reversible-transfers: if a transfer finalization task is deferred due to block weight pressure, cancel_transfer calls cancel_named which fails, and the transfer executes unhindered despite the user's cancellation attempt.

Changes:

service_task: Lookup is now only removed on terminal outcomes (successful dispatch, permanently overweight). Tasks that return to the Agenda keep their Lookup and preimage reference intact.
do_cancel_named: Returns NotFound when the Lookup points to an empty or out-of-range Agenda slot, instead of silently succeeding and emitting a false Canceled event.
Docs: Added known-limitation section for stranded tasks when preimage is unavailable.

PR #526, opened 2026-04-15
Handle Deferred Tasks Correctly

illuzen:
In the scheduler we were dropping references to tasks when they got rescheduled for being overweight. Also tasks that had no preimage would also become unreachable. 

This fixes both issues. https://github.com/Quantus-Network/chain/issues/483


n13:
## 1. Does this fix a real problem? Is the description accurate?

**Yes, it fixes a real bug. But the description undersells and slightly mislabels the problem.**

### The overweight-deferral bug (Commit 1) - REAL

On `main`, `service_task` eagerly removes the `Lookup` (name-to-address mapping) at line 1085-1087 *before knowing the outcome*:

```1082:1087:pallets/scheduler/src/lib.rs
		// Eagerly remove the name->address lookup. If the task succeeds or gets rescheduled
		// via a retry, the new address will be re-inserted by place_task. If it fails,
		// the name is freed.
		if let Some(ref id) = task.maybe_id {
			Lookup::<T>::remove(id);
		}
```

The comment claims "the new address will be re-inserted by `place_task`" - but that only happens for the *retry-on-failure* path (via `schedule_retry` -> `place_task`). For the **overweight-but-not-first** path (line 1127: `Err(()) => Err((Overweight, Some(task)))`), the task goes straight back into the same Agenda slot with **no `place_task` call**. The Lookup is gone, the task is unreachable by name.

Issue #483 confirms the real-world consequence: a reversible-transfer finalization task deferred by weight pressure becomes uncancellable.

### The preimage-zombie bug (Commit 2) - REAL, but the description is imprecise

The description says tasks "would also become unreachable." That's true but understates the problem. The existing test `postponed_named_task_cannot_be_rescheduled` on `main` documents the full horror:

```1489:1501:pallets/scheduler/src/tests.rs
		// The agenda still contains the call.
		let agenda = Agenda::<Test>::iter().collect::<Vec<_>>();
		assert_eq!(agenda.len(), 1);
		assert_eq!(
			agenda[0].1,
			vec![Some(Scheduled {
				maybe_id: Some(name),
				priority: 127,
				call: hashed,
				origin: root().into(),
				_phantom: Default::default(),
			})]
		);
```

The task is a **permanent zombie**: Lookup removed, preimage reference dropped (line 1102: `T::Preimages::drop`), yet the task stays alive (`Some(task)`) in the Agenda. Every block, `IncompleteBlockSince` causes the scheduler to re-process it, `peek` fails again, `drop` is called again on an already-dropped reference, and the task goes back. It wastes weight every block forever.

The existing test even proves that submitting the preimage later doesn't help (lines 1503-1508) - because the reference was already dropped.

### Name collision vulnerability on `main` (not mentioned in PR)

There's an additional bug the PR implicitly fixes that isn't mentioned in the description. On `main`:

1. Named task is deferred (overweight) -> Lookup removed
2. Someone calls `schedule_named` with the same name -> succeeds (line 779 check passes since Lookup is gone)
3. Now two tasks share the same `maybe_id`, but only the new one has a Lookup
4. Next block: old task is retried, `service_task` eagerly removes the *new* task's Lookup
5. The new task becomes orphaned

With the PR, step 2 correctly fails with `FailedToSchedule` because the Lookup is preserved.

---

## 2. Could this introduce problems?

### Retry path - SAFE

The `Ok(result)` path removes Lookup, then calls `schedule_retry` -> `place_task` which re-inserts Lookup. The `as_retry()` method sets `maybe_id: None`, so retried tasks intentionally have no name, and `place_task` correctly doesn't insert a Lookup for them. This worked the same on main.

### `do_cancel_named` change - SAFE, strictly better

The old code silently succeeded when the Lookup pointed to an empty Agenda slot (the `if let Some(s)` would just skip cleanup and still emit a false `Canceled` event). The new code returns `NotFound`. If the inner `Agenda::try_mutate` returns Err, the outer `Lookup::try_mutate_exists` rolls back the `lookup.take()`, so the Lookup is preserved on error. Correct transactional behavior.

### Preimage-unavailable now terminal - CORRECT for this codebase, but a behavioral divergence

This is the biggest change. On `main`, unavailable-preimage tasks zombie forever. With the PR, they're immediately killed. This diverges from upstream Polkadot SDK, where @bkchr explicitly chose to keep them alive because "unavailable pre-image could be uploaded later" ([polkadot-sdk#7790](https://github.com/paritytech/polkadot-sdk/pull/7790)).

**However**, that upstream reasoning doesn't apply here because Quantus's code calls `T::Preimages::drop` before returning the task (line 1102), which releases the preimage request. Even if someone later submits the preimage, the scheduler's reference is gone. The existing test on main proves this - the task never executes even after the preimage is submitted. So making it terminal is the correct fix *given the existing `drop` call*.

### No double-remove risk

Each terminal path removes Lookup exactly once, and non-terminal paths (overweight deferral) don't remove it at all.

---

## 3. DoS / Security concerns

### IMPROVED: Zombie weight drain is eliminated

On `main`, an attacker can schedule many tasks with never-submitted preimages. Each becomes a permanent zombie that consumes weight every block via the `IncompleteBlockSince` retry loop. The PR kills them on first encounter - a clear security improvement.

### IMPROVED: Name-squatting / Lookup poisoning is fixed

As described above, the name collision on main allows an attacker to orphan new named tasks by exploiting the deferred-task Lookup gap. The PR closes this.

### NO NEW ATTACK SURFACE: Lookup preservation doesn't enable anything new

Preserving the Lookup for overweight-deferred tasks means the name remains "occupied." An attacker could deliberately congest blocks to keep deferring a victim's named task, but the victim retains the ability to cancel or reschedule - which is strictly better than main where the victim loses all control.

### One area to watch (minor)

If a task is perpetually deferred (never enough weight), it now keeps its Lookup forever rather than losing it. This is correct behavior (the task is alive, the Lookup should exist), but it means the Lookup storage slot is occupied longer. This isn't a new cost - the Agenda slot was already occupied - it just maintains consistency.

---

## Summary

| Aspect | Verdict |
|--------|---------|
| Fixes real overweight-deferral bug? | **Yes** - named tasks losing their Lookup is a real issue with real consequences |
| Fixes real preimage-zombie bug? | **Yes** - tasks were permanent zombies consuming weight every block |
| PR description accurate? | **Partially** - understates the zombie problem, doesn't mention the name collision fix |
| Could introduce regressions? | **Low risk** - retry path works identically, `do_cancel_named` is strictly better |
| Diverges from upstream? | **Yes** on preimage-unavailable being terminal, but correctly so given the existing `drop` call |
| Security impact? | **Net positive** - eliminates zombie weight drain and Lookup poisoning |

The PR is sound. The one thing I'd flag is that the root cause of the preimage issue is actually the `T::Preimages::drop` call happening before returning the task as `Some` - that's what makes recovery impossible. The PR correctly treats the symptom (kill the zombie) but if Quantus ever wants to align with upstream's "preimage can be uploaded later" behavior, the `drop` call would need to be deferred instead.

n13 (approved):
Seems legit - see review above. 

This is an improvement!

Issue #4, opened 2025-05-08
Substrate template update to stable2503


czareko:
https://github.com/paritytech/polkadot-sdk-solochain-template/commit/7582c5ea8448638e4f5605baea2887203f30ce4a

PR #537, opened 2026-04-20
Mining Bottlenecks

illuzen:
This PR

- decreases the max rebuild frequency to every 2 seconds
- decouples block building from mining so large blocks don't stop mining via rebuild triggers

The new mining loop should be *much* easier to read and reason about now.

### DIFF
--- a/client/consensus/qpow/src/lib.rs
+++ b/client/consensus/qpow/src/lib.rs
@@ -486,137 +486,163 @@ where
 	let worker = MiningHandle::new(client.clone(), block_import, justification_sync_link);
 	let worker_ret = worker.clone();
 
-	let task = async move {
-		// Main block building loop - runs until trigger stream closes
-		// Wait for a trigger (Initial, BlockImported, or NewTransactions)
-		// continue skips to the next iteration to wait for another trigger
-		while let Some(trigger) = trigger_stream.next().await {
-			if sync_oracle.is_major_syncing() {
-				debug!(target: LOG_TARGET, "Skipping proposal due to sync.");
-				worker.on_major_syncing();
-				continue;
+	// Latest build request - overwrites previous if builder is slow.
+	// Uses a Mutex<Option> for the value + a channel for wake notification.
+	let pending_build: Arc<parking_lot::Mutex<Option<Block::Hash>>> =
+		Arc::new(parking_lot::Mutex::new(None));
+	let (notify_tx, mut notify_rx) = futures::channel::mpsc::channel::<()>(1);
+
+	// Task 1: Convert triggers into build requests
+	let trigger_task = {
+		let client = client.clone();
+		let worker = worker.clone();
+		let pending_build = pending_build.clone();
+		let mut notify_tx = notify_tx;
+		async move {
+			while let Some(trigger) = trigger_stream.next().await {
+				if sync_oracle.is_major_syncing() {
+					debug!(target: LOG_TARGET, "Skipping proposal due to sync.");
+					worker.on_major_syncing();
+					continue;
+				}
+
+				let best_hash = client.info().best_hash;
+
+				// Optimization, skip if we already imported this block
+				if trigger == RebuildTrigger::BlockImported && worker.best_hash() == Some(best_hash)
+				{
+					continue;
+				}
+
+				// Set the latest build request (overwrites any previous)
+				*pending_build.lock() = Some(best_hash);
+				let _ = notify_tx.try_send(());
 			}
-
-			let best_hash = client.info().best_hash;
-			let best_header = match client.header(best_hash) {
-				Ok(Some(header)) => header,
-				Ok(None) => {
-					warn!(
-						target: LOG_TARGET,
-						"Unable to pull new block for authoring. \
-						 Best header not found for hash: {:?}",
-						best_hash
-					);
-					continue;
-				},
-				Err(err) => {
-					warn!(
-						target: LOG_TARGET,
-						"Unable to pull new block for authoring. \
-						 Header lookup error: {}",
-						err
-					);
-					continue;
-				},
-			};
-
-			// Skip redundant block import triggers if we're already building on this hash.
-			// Initial and NewTransactions triggers should proceed to rebuild.
-			if trigger == RebuildTrigger::BlockImported && worker.best_hash() == Some(best_hash) {
-				continue;
-			}
-
-			// The worker is locked for the duration of the whole proposing period. Within this
-			// period, the mining target is outdated and useless anyway.
-
-			let difficulty = match qpow_get_difficulty::<Block, C>(&*client, best_hash) {
-				Ok(x) => x,
-				Err(err) => {
-					warn!(
-						target: LOG_TARGET,
-						"Unable to propose new block for authoring. \
-						 Fetch difficulty failed: {}",
-						err,
-					);
-					continue;
-				},
-			};
-
-			let inherent_data_providers = match create_inherent_data_providers
-				.create_inherent_data_providers(best_hash, ())
-				.await
-			{
-				Ok(x) => x,
-				Err(err) => {
-					warn!(
-						target: LOG_TARGET,
-						"Unable to propose new block for authoring. \
-						 Creating inherent data providers failed: {}",
-						err,
-					);
-					continue;
-				},
-			};
-
-			let inherent_data = match inherent_data_providers.create_inherent_data().await {
-				Ok(r) => r,
-				Err(e) => {
-					warn!(
-						target: LOG_TARGET,
-						"Unable to propose new block for authoring. \
-						 Creating inherent data failed: {}",
-						e,
-					);
-					continue;
-				},
-			};
-
-			let mut inherent_digest = Digest::default();
-			let rewards_preimage_bytes = rewards_preimage.to_vec();
-			inherent_digest.push(DigestItem::PreRuntime(POW_ENGINE_ID, rewards_preimage_bytes));
-
-			let proposer = match env.init(&best_header).await {
-				Ok(x) => x,
-				Err(err) => {
-					warn!(
-						target: LOG_TARGET,
-						"Unable to propose new block for authoring. \
-						 Creating proposer failed: {:?}",
-						err,
-					);
-					continue;
-				},
-			};
-
-			let proposal =
-				match proposer.propose(inherent_data, inherent_digest, build_time, None).await {
-					Ok(x) => x,
-					Err(err) => {
-						warn!(
-							target: LOG_TARGET,
-							"Unable to propose new block for authoring. \
-							 Creating proposal failed: {}",
-							err,
-						);
-						continue;
-					},
-				};
-
-			let build = MiningBuild::<Block, _> {
-				metadata: MiningMetadata {
-					best_hash,
-					pre_hash: proposal.block.header().hash(),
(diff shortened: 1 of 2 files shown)


n13:
Comments:

I think 3 needs to be addressed, it could screw up importing ... 

4 too...



# Review of PR #537 — Mining Bottlenecks

Happy to dig in! Overall this is a solid, well-motivated refactor. The new mining loop is noticeably easier to reason about, the `create_proposal` extraction is clean, and the 2s rate-limit choice addresses a real throughput concern. Here are my findings.

## What I like

1. **Decoupling makes rate limiting actually effective.** In the old loop, `trigger_stream` was only polled once per build cycle. For fast builds (~50ms) with `MAX_REBUILDS_PER_SEC=2`, the 500 ms rate limit kicked in. But in the rebuild-loop scenario (moderate tx load, sub-second builds) mining time between rebuilds was starved. Continuously polling the stream in `trigger_task` means the delay window is now correctly honored — miners actually get their 2 s.
2. **The `best_hash != target_hash` discard check** in `create_proposal` prevents the worker from being poisoned with a stale build after a network import lands mid-propose. Nice.
3. **`create_proposal` extraction** — much more readable, and independently testable should tests be added later. The `match … return None` style matches the "log and skip" semantics the old code had.
4. **Types/bounds/builds cleanly** — `cargo check` and `cargo clippy -p sc-consensus-qpow` are both clean, tests still pass (there are none in this crate, but also no regression).

## Issues worth addressing before merge

### 1. Stale doc comments referencing the old constant/arg name

Two doc comments still reference the pre-rename names:

```442:443:client/consensus/qpow/src/lib.rs
/// - A new block is imported from the network
/// - New transactions arrive (rate limited to MAX_REBUILDS_PER_SEC)
```

```272:272:client/consensus/qpow/src/worker.rs
/// Rate limiting prevents excessive rebuilds - we limit to `max_rebuilds_per_sec`.
```

Both should refer to `MIN_SECS_BETWEEN_TX_REBUILDS` / `min_rebuild_interval`.

### 2. Silent failure on `try_send` (conflicts with "fail early" convention)

```519:519:client/consensus/qpow/src/lib.rs
let _ = notify_tx.try_send(());
```

`TrySendError::Full` is the expected, intentionally-ignored case (means "a wake is already queued, no-op") — that's fine. But `TrySendError::Disconnected` means `build_task` died, and in this code path that's unreachable-unless-panic but worth logging at warn rather than being truly silent. Something like:

```rust
match notify_tx.try_send(()) {
    Ok(()) | Err(e) if e.is_full() => {}
    Err(e) => warn!(target: LOG_TARGET, "notify channel disconnected: {e}"),
}
```

### 3. Regression window: miners can have `metadata() == None` during fast network imports

In the old loop, `on_build(build)` was *always* called at the end of a build, so `handle_local_mining`/`handle_external_mining` had work to mine on even if the build was already stale vs. the new tip.

In the new loop, when `client.info().best_hash` changes during the propose call, the build is discarded:

```631:635:client/consensus/qpow/src/lib.rs
// Check if best_hash changed during building
if client.info().best_hash != best_hash {
    debug!(target: LOG_TARGET, "Best hash changed during block building, discarding");
    return None;
}
```

If the network is importing blocks faster than local `build_time` (or right at startup when `worker.best_hash()` is `None`), we can enter a loop of "take → build → discard → take → build → discard…" with `on_build` never firing. Miners would see `metadata().is_none()` indefinitely.

Mitigations to consider:
- Also check `sync_oracle.is_major_syncing()` inside `build_task` before calling `create_proposal` (requires `SO: Clone` — already bounded — and a clone into `build_task`). If `is_major_syncing` is true, the old code would simply not start the build; the new code starts it and then (correctly) discards it after wasting the build time.
- Bumping the log level from `debug!` to `info!` (or at least `debug!` with build-time included) on discard so this is observable in production.

### 4. `on_major_syncing` / `on_build` race during sync transition

`on_major_syncing` clears the build, but if `build_task` is already midway through a `propose` when sync starts, its `on_build` at the end will overwrite the cleared state (the best-hash guard doesn't help here — a build started before sync may still have a matching `best_hash`).

Small window, probably benign, but again cloning `sync_oracle` into `build_task` and re-checking right before `worker.on_build(build)` would close it:

```rust
if sync_oracle.is_major_syncing() { continue; }
worker.on_build(build);
```

### 5. Minor: make the interval a `Duration` directly

`Duration::from_secs` has been `const` since Rust 1.32, so the u64 + conversion at the call site can collapse to a single constant:

```rust
const MIN_INTERVAL_BETWEEN_TX_REBUILDS: Duration = Duration::from_secs(2);
```

Then `UntilImportedOrTransaction::new(..., MIN_INTERVAL_BETWEEN_TX_REBUILDS)`. Minor aesthetic.

### 6. No tests added

`sc-consensus-qpow` has zero tests. Given that this PR intentionally changes timing/coordination behavior, even a single unit test for `UntilImportedOrTransaction` (FakeTime-driven firing cadence) would make future changes here much safer. Not a blocker, but highly valuable.

## Tiny nits

- `let mut notify_tx = notify_tx;` rebinding inside the block is unnecessary — the outer binding is not captured elsewhere; you can just declare it `mut` at the destructure: `let (mut notify_tx, mut notify_rx) = futures::channel::mpsc::channel::<()>(1);`.
- `// Optimization, skip if we already imported this block` — the comma reads oddly; "Skip if we're already building on this hash" matches the old wording and is clearer.

## Sanity verified locally

- `cargo check -p sc-consensus-qpow` clean
- `cargo clippy -p sc-consensus-qpow` clean (no new warnings)
- `cargo test -p sc-consensus-qpow` passes (0 tests exist in the crate)
- Trigger/build ordering analysis: notifications after `pending_build` mutation, receiver takes after receiving notification, `parking_lot::Mutex` operations are synchronous between the two async tasks which are multiplexed on the same task via `join!` — no observable reordering bugs

## TL;DR

Ship it once the two stale doc comments and the silent `try_send` are addressed. Items 3–6 would be nice follow-ups (especially #3 if you've observed it under real mining traffic). Great cleanup — the new loop really is much easier to follow.

illuzen:
Addressed 1,3,5 I don't think 2, 4 and 6 are worth addressing. Added a comment for 2.

n13:
Pulled `ef85ca7` and re-verified everything. Quick re-review below.

## What was addressed cleanly

**#1 — Stale doc comments** ✅
Both spots now reference the current names:

```429:443:client/consensus/qpow/src/lib.rs
/// Minimum interval between transaction-triggered rebuilds.
/// Set high enough to prevent the "rebuild loop" under high tx load where block construction
/// time dominates and effective mining time approaches zero, causing block times to spike.
const MIN_INTERVAL_BETWEEN_TX_REBUILDS: Duration = Duration::from_secs(2);
...
/// - New transactions arrive (rate limited to MIN_INTERVAL_BETWEEN_TX_REBUILDS)
```

```272:272:client/consensus/qpow/src/worker.rs
/// Rate limiting prevents excessive rebuilds via `min_rebuild_interval`.
```

**#5 — `Duration` constant** ✅
Collapsed to a single `const Duration` (`lib.rs:432`) and used directly at the call site (`lib.rs:484`). Reads great.

**#3 — Discard log visibility** ✅ (partially)
Bumped to `info!` and clarified the message:

```631:635:client/consensus/qpow/src/lib.rs
// Check if best_hash changed during building
if client.info().best_hash != best_hash {
    info!(target: LOG_TARGET, "Best hash changed during block building, discarding proposal");
    return None;
}
```

This makes the behavior observable in production logs, which was half my ask. The behavioral concern itself is below ⤵.

## Dismissed items — my read

**#2 — Silent `try_send`**
Author kept the `let _ = ...` and added an explanatory inline comment:

```519:519:client/consensus/qpow/src/lib.rs
let _ = notify_tx.try_send(()); // Err is ok: Full (wake queued) or Disconnected (will exit)
```

That's defensible. Minor accuracy nit: `Disconnected` here would only occur if `build_task` already ended, which in this code path means it panicked — and `futures::join!` will propagate that panic to the outer task, killing the whole worker. So in practice this branch is unreachable; the comment's "will exit" is right in spirit. I'd accept this as-is.

**#4 — `on_major_syncing`/`on_build` race**
Still present. n13's "4 too" comment lines up with my concern — there's a window where:

1. Build is in flight on best_hash = `H`
2. Major sync starts, `trigger_task` calls `worker.on_major_syncing()` (clears build to `None`)
3. Build finishes, `client.info().best_hash` is still coincidentally `H` (sync just started, hasn't moved tip yet), so the discard guard does *not* fire
4. `worker.on_build(build)` re-populates the build the syncing-clear just emptied

Then a miner can read `metadata()`, find a nonce, and `submit()`. `submit()` will call `block_import.import_block()` while the import queue is also processing synced blocks. `PowBlockImport` is internally safe (it has the rollback path for `delete_cumulative_achieved_work` and proper fork-choice via `is_heavier`), so this won't corrupt anything — but it can:

- Add an extra fork to aux storage that loses on heaviest-chain anyway
- Compete for the import-queue lock with sync imports (modest CPU/IO impact)

Closing this is one extra `sync_oracle.is_major_syncing()` check immediately before `worker.on_build(build)`. Cheap, and would address the only credible angle on n13's "could screw up importing" concern. I think this one is genuinely worth doing — but it's a 3-line change that can land as a follow-up if the goal is to ship.

**#6 — Tests**
Still none. Author's call.

## Residual concern on #3 even after the log bump

The `info!` makes the discard visible, but the underlying scenario remains: during minor catch-up (a few blocks behind, not enough to trigger `is_major_syncing`), the loop can churn `take → propose for ~3s → discard → repeat` indefinitely while imports stream in. During that window:

- `worker.metadata()` stays `None`, so miners are idle (probably the right thing)
- BUT the `propose` call itself is not free — it executes inherents, opens a `BlockBuilder`, runs the runtime, etc., all of which contend with the import queue's WASM executor and DB

This is, I suspect, what n13 meant by "could screw up importing": not corruption, but CPU/IO contention slowing down the very sync we need to finish before we can mine usefully. The simplest mitigation is the same as #4 — re-check `sync_oracle.is_major_syncing()` in `build_task` before calling `create_proposal`, OR add a heuristic like "skip if `client.info().best_hash != target_hash` *before* spending 3s on the build" so we cut the loop early on rapidly-changing tips. The early-skip is `O(1)` and would have prevented a lot of the wasted-CPU symptom you described in the stress test.

A minimal version, inside `build_task` right after taking `target_hash`:

```rust
let Some(target_hash) = pending_build.lock().take() else { continue };
if client.info().best_hash != target_hash {
    info!(target: LOG_TARGET, "Skipping stale build target {target_hash:?}");
    continue;
}
```

Cheap, symmetric with the existing post-build check, no new state.

## Sanity verified locally

- `cargo check -p sc-consensus-qpow` clean
- `cargo clippy -p sc-consensus-qpow` clean (no new warnings)
- `cargo test -p sc-consensus-qpow` still passes (still 0 tests; #6 unchanged)

## TL;DR

The addressed items look good — clean diff, no regressions. I'd be comfortable approving as-is. If you want one more nudge before merging, I'd ask for the 3-line `sync_oracle` re-check inside `build_task` (closes #4 and the residual half of #3 that the log bump didn't actually fix). Everything else is fair game for a follow-up PR. Nice iteration!

n13 (approved):
Approved with above comments - up to author to decide if they need addressing

czareko (approved):

PR #541, opened 2026-04-21
Innerhash param fix

n13:
when a node runs as listener - not authoring blocks - no inner hash parameter is required


n13:
Testing Composer 2 PR review

Here is a concise review of [PR #541 — Innerhash param fix](https://github.com/Quantus-Network/chain/pull/541).

## Summary

The change matches how the node actually uses `rewards_address`: it is only passed into `spawn_authority_tasks` when the node is an authority, so requiring `--rewards-inner-hash` for every non-dev full node was stricter than the runtime wiring.

```800:828:node/src/service.rs
	if role.is_authority() {
		#[cfg(feature = "tx-logging")]
		spawn_authority_tasks(
			&mut task_manager,
			client,
			transaction_pool,
			pow_block_import,
			sync_service,
			prometheus_registry,
			rewards_address,
			miner_listen_port,
			tx_stream_for_worker,
			tx_stream_for_logger,
			allow_mining_without_peers,
		);
		#[cfg(not(feature = "tx-logging"))]
		spawn_authority_tasks(
			&mut task_manager,
			client,
			transaction_pool,
			pow_block_import,
			sync_service,
			prometheus_registry,
			rewards_address,
			miner_listen_port,
			tx_stream_for_worker,
			allow_mining_without_peers,
		);
	}
```

For a non-authority (listener) node, `rewards_address` is never read, so a dummy `AccountId32::new([0u8; 32])` is safe and avoids inventing a fake “real” rewards account.

## What works well

- **Correct gate:** `config.role.is_authority()` lines up with the only consumer (`spawn_authority_tasks` / mining).
- **Dev unchanged:** `shared_params.is_dev()` still uses the treasury account before the authority check.
- **Clearer error:** Messaging that the flag is required when running as a validator matches behavior better than a blanket “required.”
- **Optional on full nodes:** Operators can still pass `--rewards-inner-hash` on listeners if they want the log line with the wormhole address; nothing breaks.

## Minor notes (non-blocking)

1. **Comment wording:** Saying it is “never consumed by `service::new_full`” is a bit loose — the argument is still passed in, it is just unused when `!role.is_authority()`. Something like “unused for non-authority roles” would be more precise.
2. **Docs:** `MINING.md` still describes `--rewards-inner-hash` as generally required; worth a small follow-up so it states “validators / block producers” (or equivalent) so it matches the CLI.
3. **Style:** If the repo prefers almost no comments, the new block comment could be dropped without losing clarity.

## Verdict

**Approve** from a correctness and safety perspective: the placeholder is only relevant in branches where `rewards_address` is never used, and validators still must supply `--rewards-inner-hash` outside `--dev`. Nice, focused fix for listener-only deployments.

illuzen (approved):

PR #546, opened 2026-04-22
Address Eiger Multisig Audit - 2

illuzen:
There are three sections of the audit (included in the PR), but I am splitting it roughly per pallet to make the PRs smaller.


n13:
# PR #546 — Address Eiger Multisig Audit (Part 2)

**24 commits across 14 files (+1,095 / -825)**
Scope: multisig, treasury, scheduler, reversible-transfers, runtime config

---

## Findings

### CRITICAL: `execute()` propagates inner call error → transactional rollback

**File:** `pallets/multisig/src/lib.rs`

The refactored `execute()` now returns the inner call's `DispatchError` as its own return value:

```rust
result
    .map(|_| PostDispatchInfo { actual_weight: Some(total_weight), pays_fee: Pays::Yes })
    .map_err(|e| DispatchErrorWithPostInfo {
        post_info: PostDispatchInfo { actual_weight: Some(total_weight), pays_fee: Pays::Yes },
        error: e.error,
    })
```

In FRAME, returning `Err` from a dispatchable triggers **transactional rollback** — so if the inner call fails (e.g. insufficient balance for a transfer), the proposal removal and deposit return are **also reverted**. The proposal gets stuck in storage with no way to re-execute it.

The old `do_execute()` correctly returned `Ok(())` regardless of inner call outcome, and reported success/failure only through the `ProposalExecuted` event's `result` field.

**Recommendation:** Always return `Ok(PostDispatchInfo)` from `execute`. The inner call's success/failure should only be reflected in the `ProposalExecuted` event's `result` field — which matches the README's documented behavior: *"The execute extrinsic itself succeeds even if the inner call fails."*

---

### HIGH: `execute()` weight does not include inner call weight

**File:** `pallets/multisig/src/lib.rs`

The weight annotation charges `WeightInfo::execute(MaxCallSize)` which only covers bookkeeping overhead (benchmarked with a `remark` call). The `actual_weight` then adds `bookkeeping + inner_call_weight`, which can **exceed the pre-charge** for heavy inner calls.

Substrate's `pallet-proxy` solves this by adding `call.get_dispatch_info().call_weight` to the annotation. Since `execute` reads the call from storage (not available at annotation time), the weight can't be computed from params alone — but the proposal is read in the body, so the actual call weight could be included.

**Recommendation:** Consider reading the proposal early and adding the decoded call's declared weight to the benchmark weight. Alternatively, document that `execute` is weight-safe only for calls within benchmark bounds.

---

### HIGH: No migration for removed `MultisigData` fields and `DissolveApprovals`

**File:** `pallets/multisig/src/lib.rs`

`MultisigData` lost `deposit` and `active_proposals` fields, `DissolveApprovals` storage was removed entirely, and `ProposalStatus` lost `Executed`/`Cancelled` variants. Existing on-chain `MultisigData` will fail to decode. `StorageVersion` is set to 0 (baseline), which is correct only if the chain has no existing multisigs.

**Recommendation:** If the chain has launched and has existing state, add a storage migration. If this is pre-launch, document that `StorageVersion` 0 assumes a clean slate.

---

### MEDIUM: Silent signer deduplication may surprise users

**File:** `pallets/multisig/src/lib.rs`

`create_multisig` now silently deduplicates signers instead of returning `DuplicateSigner` error. `[alice, alice, bob]` with `threshold=2` silently becomes a 2-of-2 multisig. The post-dedup validation catches `threshold > unique_signers`, but users may not realize their intended 3-signer setup became 2-signer.

**Recommendation:** Consider emitting a warning event when duplicates are removed, or returning an error like the old behavior. Silent dedup is a valid design choice but worth documenting prominently.

---

### MEDIUM: `call_index(6)` gap from removed `approve_dissolve`

**File:** `pallets/multisig/src/lib.rs`

`approve_dissolve` used `call_index(6)`. After removal, `execute` stays at `call_index(7)`. This leaves a gap in the pallet's ABI. Old encoded calls targeting index 6 will fail to dispatch (safe), but the gap is permanent.

**Recommendation:** This is fine for correctness. Just noting for awareness — the gap should not be reused by future extrinsics to avoid ambiguity with historical transactions.

---

## What Looks Good

### Multisig Design

- Dissolve mechanism removed — eliminates griefing vectors (dust attacks blocking dissolution) and simplifies the pallet significantly
- `active_proposals` derived from `proposals_per_signer.values().sum()` — eliminates redundant state that could desync
- Call validation at propose time for ALL multisigs (not just HS) — catches malformed calls early
- `BoundedVec<u8>` for call data — size enforced at type level, removes `CallTooLarge` runtime check
- `ProposalReadyToExecute` only emitted when threshold first crossed — no duplicate events on extra approvals
- `proposal_nonce` uses `checked_add` instead of `saturating_add` — explicit `ProposalNonceExhausted` error
- `ProposalApproved` → `SignerApproved` rename clarifies semantics

### Defensive Programming

- Storage version baselines (v0) added to multisig, treasury, and reversible-transfers pallets
- `defensive!()` macros added for invariant violations in scheduler (Lookup/Agenda sync), reversible-transfers (preimage realize, PendingTransfers consistency)
- `defensive_map_err` / `defensive_ok_or` used instead of plain `ok_or` for internal-only error paths

### Treasury & Runtime

- `TreasuryAccountUpdated` event now includes `old_account` for audit trail
- Documentation added explaining that account change doesn't migrate existing funds
- `MultisigFee` consolidated: 100 + 500 deposit → 600 burned fee (same total economic cost)
- `recover_funds` added to high-security whitelist
- Extensive `proof_size` design rationale documented in runtime config

### Testing

- ~400 lines of new test coverage for audit findings
- Reentrancy test: multisig executing a call back into the multisig pallet
- Nonce overflow test with `u32::MAX`
- Fee atomicity: verifies fee is NOT burned when deposit reservation fails
- Edge cases: `claim_deposits` with zero expired, cancel on approved proposals

---

## Files Changed

| File | + | - | Scope |
|------|--:|--:|-------|
| `pallets/multisig/src/lib.rs` | 284 | 342 | Core logic refactor |
| `pallets/multisig/src/tests.rs` | 487 | 166 | Audit test coverage |
| `pallets/reversible-transfers/src/lib.rs` | 100 | 69 | Defensive assertions |
| `pallets/scheduler/src/lib.rs` | 48 | 27 | Defensive checks, docs |
| `pallets/treasury/src/lib.rs` | 29 | 2 | Event enrichment, storage version |
| `runtime/src/configs/mod.rs` | 64 | 8 | Fee consolidation, proof_size docs |
| `pallets/multisig/src/benchmarking.rs` | 13 | 58 | Remove dissolve benchmarks |
| `pallets/multisig/src/weights.rs` | 0 | 70 | Remove dissolve weights |
| `pallets/multisig/README.md` | 37 | 67 | Updated documentation |

---

ethan-crypto (dismissed):
## Blocking findings

### 1. `execute()` still propagates inner-call failure as an outer extrinsic failure

The README now states that `execute` should **always remove** the proposal, return the proposal deposit, emit `ProposalExecuted`, and have the event carry the inner call’s `Ok/Err` result. ([[GitHub](https://github.com/Quantus-Network/chain/pull/546/files)][1])

The implementation removes the proposal and returns the deposit before dispatch, but then returns `result.map(...).map_err(...)`, turning an inner dispatch error into an outer `execute` error. ([[GitHub](https://github.com/Quantus-Network/chain/raw/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][2]) On modern Substrate/FRAME this is especially dangerous because dispatch errors are normally executed within a transactional rollback model, so the storage cleanup and deposit return can be reverted even though the code intended to do them before dispatch. ([[Substrate and Polkadot Stack Exchange](https://substrate.stackexchange.com/questions/1519/transaction-rollback-in-substrate?utm_source=chatgpt.com)][3])

Suggested fix: after wrapper-level validation passes, treat the inner dispatch result as event data, not as the outer extrinsic result. Return `Ok(PostDispatchInfo { actual_weight: Some(total_weight), pays_fee: Pays::Yes })` regardless of whether the inner call returned `Ok` or `Err`. Wrapper errors like “proposal not found”, “not approved”, “expired”, or “invalid call decode” should still return outer `Err`.

Add a regression test where the inner call fails and assert that:

```text
execute(...) returns Ok
ProposalExecuted.result is Err(...)
proposal is removed
proposal deposit is unreserved/returned
the proposal cannot be executed again
```

### 2. `execute()` undercharges block weight for arbitrary inner calls

The call annotation charges only `WeightInfo::execute(T::MaxCallSize::get())`, even though the doc comment says the weight includes both multisig bookkeeping and the inner call’s declared weight. ([[GitHub](https://github.com/Quantus-Network/chain/blob/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][4]) The generated `execute(c)` weight is just a linear function of encoded call size plus multisig storage reads/writes; it has no parameter for the actual dispatch weight of the stored runtime call. ([[GitHub](https://raw.githubusercontent.com/Quantus-Network/chain/illuzen/eiger-runtime-audit/pallets/multisig/src/weights.rs)][5]) The benchmark also uses a `System::remark` payload by size, so it does not benchmark a worst-case arbitrary inner runtime call. ([[GitHub](https://raw.githubusercontent.com/Quantus-Network/chain/illuzen/eiger-runtime-audit/pallets/multisig/src/benchmarking.rs)][6])

The implementation computes `call.get_dispatch_info().call_weight` only after it has already passed pre-dispatch weight checks, and then returns `bookkeeping + actual_call_weight` as post-dispatch info. ([[GitHub](https://github.com/Quantus-Network/chain/raw/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][2]) That can allow a cheap-looking `execute` extrinsic to include a heavy inner call that was not reserved against block weight up front.

Suggested fix options:

```text
Option A:
- Add a Config::MaxInnerCallWeight bound.
- At propose time, decode the call and reject calls whose declared call weight exceeds MaxInnerCallWeight.
- Store the call weight in ProposalData if useful.
- Annotate execute as bookkeeping(MaxCallSize) + MaxInnerCallWeight.
- Return the actual lower weight post-dispatch for fee refunds.

Option B:
- Change execute so the call itself is supplied as an argument and can be included in the static weight expression, while still checking it matches the stored proposal hash/bytes.
```

The important invariant is: **the annotated/predispatch weight must be a safe upper bound for any inner call that `execute` can dispatch.**

### 3. High-security whitelist is checked at `propose`, but not at `execute`

The runtime’s high-security transaction extension validates the signed caller’s high-security state against `HighSecurityConfig::is_whitelisted(call)`. ([[GitHub](https://raw.githubusercontent.com/Quantus-Network/chain/illuzen/eiger-runtime-audit/runtime/src/transaction_extensions.rs)][7]) The reversible-transfers pallet also documents that once an account is set as high-security, it can only make reversible transfers and is not allowed other calls. ([[GitHub](https://raw.githubusercontent.com/Quantus-Network/chain/illuzen/eiger-runtime-audit/pallets/reversible-transfers/src/lib.rs)][8])

In multisig, the whitelist check happens only when the proposal is created: `propose` decodes the call and checks `T::HighSecurity::is_high_security(&multisig_address)` and `is_whitelisted(&decoded_call)`. ([[GitHub](https://github.com/Quantus-Network/chain/blob/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][4]) `execute` later decodes and dispatches the stored call as the multisig account, but does not re-check the current high-security status or current whitelist before dispatch. ([[GitHub](https://github.com/Quantus-Network/chain/raw/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][2])

That leaves a bypass window:

```text
1. Multisig is not high-security.
2. Signer proposes a non-whitelisted call.
3. Multisig later becomes high-security, for example by executing set_high_security.
4. The old non-whitelisted proposal can still be executed as the now-high-security multisig.
```

It also creates a runtime-upgrade edge case: a call that was whitelisted or valid at proposal time may no longer be allowed when executed.

Suggested fix: after decoding the call in `execute`, but before removing the proposal and dispatching, re-run:

```rust
if T::HighSecurity::is_high_security(&multisig_address)
    && !T::HighSecurity::is_whitelisted(&call)
{
    return Self::err_with_weight(Error::<T>::CallNotAllowedForHighSecurityMultisig, ...);
}
```

Add a regression test for “propose while non-HS, enable HS, then attempt to execute non-whitelisted proposal”.

### 4. Storage layout changed, but `StorageVersion` remains `0`

The new `MultisigData` layout now contains `creator`, `signers`, `threshold`, `proposal_nonce`, and `proposals_per_signer`; proposal status is now only `Active` or `Approved`. ([[GitHub](https://github.com/Quantus-Network/chain/blob/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][4]) The pallet storage version is still initialized as `StorageVersion::new(0)`. ([[GitHub](https://github.com/Quantus-Network/chain/blob/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][4]) The README diff also reflects removal of stored executed/cancelled history and dissolution-related storage. ([[GitHub](https://github.com/Quantus-Network/chain/pull/546/files)][1])

If this pallet has ever been deployed with the old layout, this needs an explicit migration. Otherwise, old `MultisigData`, proposal statuses, and `DissolveApprovals` entries can fail decoding, remain orphaned, or corrupt accounting assumptions.

Suggested fix:

```text
- If this is a clean-slate/unlaunched pallet: document that clearly in the runtime release notes.
- If not: bump StorageVersion to 1 and add an on_runtime_upgrade migration.
- Include try-runtime pre/post checks for:
  - MultisigData translation
  - ProposalStatus translation/removal policy
  - removal of DissolveApprovals
  - proposal deposit/accounting invariants
```

## Non-blocking but worth fixing

The code silently sorts and deduplicates signers during `create_multisig`, then validates the threshold against the deduped signer count. ([[GitHub](https://github.com/Quantus-Network/chain/blob/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][4]) The README’s validation section still says “No duplicate signers,” while also saying signers are automatically sorted. ([[GitHub](https://github.com/Quantus-Network/chain/pull/546/files)][1]) Either reject duplicates explicitly or document that duplicates are accepted but ignored. Silent normalization is fine if intentional, but it should be impossible for callers to misunderstand the resulting threshold semantics.

There are also stale comments/docs around storage status. The config comment still describes `MaxTotalProposalsInStorage` as including “Active + Executed + Cancelled,” but the new model removes executed/cancelled proposals immediately and stores only active/approved proposals. ([[GitHub](https://github.com/Quantus-Network/chain/blob/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][4]) Clean this up so future reviewers do not reason about nonexistent stored states.

## What looks good

The direction of the refactor is solid: using `BoundedCallOf<T>` for proposal calls, decoding at proposal time, and validating high-security proposals before storage are all improvements. ([[GitHub](https://github.com/Quantus-Network/chain/blob/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][4]) Removing the proposal before the inner dispatch is also the right reentrancy shape; it just needs the outer `execute` return value fixed so the cleanup actually persists on inner failure. ([[GitHub](https://github.com/Quantus-Network/chain/raw/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs)][2])

[1]: https://github.com/Quantus-Network/chain/pull/546/files "Address Eiger Multisig Audit - 2 by illuzen · Pull Request #546 · Quantus-Network/chain · GitHub"
[2]: https://github.com/Quantus-Network/chain/raw/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs "raw.githubusercontent.com"
[3]: https://substrate.stackexchange.com/questions/1519/transaction-rollback-in-substrate?utm_source=chatgpt.com "Transaction rollback in Substrate"
[4]: https://github.com/Quantus-Network/chain/blob/illuzen/eiger-runtime-audit/pallets/multisig/src/lib.rs "chain/pallets/multisig/src/lib.rs at illuzen/eiger-runtime-audit · Quantus-Network/chain · GitHub"
[5]: https://raw.githubusercontent.com/Quantus-Network/chain/illuzen/eiger-runtime-audit/pallets/multisig/src/weights.rs "raw.githubusercontent.com"
[6]: https://raw.githubusercontent.com/Quantus-Network/chain/illuzen/eiger-runtime-audit/pallets/multisig/src/benchmarking.rs "raw.githubusercontent.com"
[7]: https://raw.githubusercontent.com/Quantus-Network/chain/illuzen/eiger-runtime-audit/runtime/src/transaction_extensions.rs "raw.githubusercontent.com"
[8]: https://raw.githubusercontent.com/Quantus-Network/chain/illuzen/eiger-runtime-audit/pallets/reversible-transfers/src/lib.rs "raw.githubusercontent.com"

n13:
This is good to go, approved.

AI review is happy, we don't need migration

## Summary

All four blocking issues from the prior review have been addressed in code and backed by regression tests. The 52 multisig tests pass locally and CI is green on all the latest commits.

## Blocking findings — re-review

### 1. `execute()` no longer propagates inner-call failure — FIXED

```1093:1118:pallets/multisig/src/lib.rs
		let result =
			call.dispatch(frame_system::RawOrigin::Signed(multisig_address.clone()).into());

		Self::deposit_event(Event::ProposalExecuted {
			multisig_address,
			proposal_id,
			proposer: proposal.proposer,
			call: proposal.call.to_vec(),
			approvers: proposal.approvals.to_vec(),
			result: result.as_ref().map(|_| ()).map_err(|e| e.error),
		});

		let actual_call_weight = match &result {
			Ok(info) | Err(DispatchErrorWithPostInfo { post_info: info, .. }) =>
				info.actual_weight.unwrap_or(proposal.call_weight),
		};
		let total_weight = bookkeeping_weight.saturating_add(actual_call_weight);

		Ok(PostDispatchInfo { actual_weight: Some(total_weight), pays_fee: Pays::Yes })
```

The dispatch result is now used purely as event payload; `execute` always returns `Ok`, so storage cleanup and deposit return persist even when the inner call fails. Regression test `execute_succeeds_even_when_inner_call_fails` confirms proposal removal + deposit refund + `Err` in event when an underfunded transfer is dispatched.

### 2. `execute()` weight pre-charges the worst-case inner call — FIXED

```1014:1021:pallets/multisig/src/lib.rs
		#[pallet::call_index(6)]
		#[pallet::weight({
			<T as Config>::WeightInfo::execute(T::MaxCallSize::get())
				.saturating_add(T::MaxInnerCallWeight::get())
		})]
```

A new `Config::MaxInnerCallWeight` was added. `propose()` validates the decoded call weight against it and stores `call_weight` in `ProposalData` (used for accurate post-dispatch refunds):

```584:590:pallets/multisig/src/lib.rs
		let call_weight = decoded_call.get_dispatch_info().call_weight;
		let max_inner_weight = T::MaxInnerCallWeight::get();
		if call_weight.any_gt(max_inner_weight) {
			return Self::err_with_weight(Error::<T>::CallWeightExceedsLimit, 1);
		}
```

Test `propose_rejects_call_exceeding_max_inner_weight` builds a real overweight call (`mock_heavy_call::too_heavy` at 2B ref_time / 2.1 MB proof size vs. limit 1B / 1 MB) and asserts: `CallWeightExceedsLimit`, no proposal stored, nonce unchanged, no balance reserved/burned. Solid atomicity check.

Runtime config sets `MaxInnerCallWeight = 1s ref_time, 2.5 MB proof_size` (~50% of max block) which is a sensible cap.

### 3. HS whitelist re-checked at execute time — FIXED

```1071:1079:pallets/multisig/src/lib.rs
		if T::HighSecurity::is_high_security(&multisig_address) &&
			!T::HighSecurity::is_whitelisted(&call)
		{
			return Self::err_with_weight(Error::<T>::CallNotAllowedForHighSecurityMultisig, 2);
		}
```

This closes the "propose-while-non-HS, enable-HS, then execute" bypass. Two regression tests cover both directions: `execute_rejects_non_whitelisted_call_after_hs_enabled` (rejects with proposal kept in storage) and `execute_allows_whitelisted_call_after_hs_enabled`.

### 4. `StorageVersion` still `0` — needs explicit confirmation

```149:157:pallets/multisig/src/lib.rs
	const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);

	#[pallet::pallet]
	#[pallet::storage_version(STORAGE_VERSION)]
	pub struct Pallet<T>(_);
```

The pallet has shipped on `testnet/planck`, and this PR removes fields (`deposit`, `active_proposals`), removes the `DissolveApprovals` storage, and removes `ProposalStatus::{Executed, Cancelled}` variants. Old encoded `MultisigData` will fail to decode against the new layout. The version stays at `0` with no `on_runtime_upgrade` migration.

If the team is treating this as a clean-slate / testnet-reset launch, this is fine — please make that explicit in the PR description and runtime release notes. Otherwise it needs StorageVersion bumped to 1 plus a migration that:
- Translates `MultisigData` (drop `deposit`/`active_proposals`)
- Drops any pending `Executed`/`Cancelled` proposals (or maps to `Active`)
- Removes `DissolveApprovals` keys
- Unreserves any previously reserved per-multisig deposits

This is the only item that doesn't have a code-level fix in this round.

## Other items addressed

- **Call-index gap**: `execute` was moved from `call_index(7)` back to `(6)`, closing the hole left by removing `approve_dissolve`. Acceptable as part of pre-launch ABI changes.
- **Silent dedup documentation**: README now explicitly documents auto-dedup behavior in the validation list and a dedicated "Signer Order Doesn't Matter" section, with a `[alice, bob, bob, charlie]` example.
- **Stale `MaxTotalProposalsInStorage` comment**: now reads "Only Active and Approved proposals are stored" in both `lib.rs` and `runtime/configs/mod.rs`.
- **`Config::HighSecurity` re-check** + `MaxInnerCallWeight` are reflected in the README configuration and "How It Works" sections.

## Minor housekeeping

- `spec_version` on the PR is `128` while `main` is now `129`. Will need a rebase and a bump (e.g. to `130`) before merge — and definitely a bump if a storage migration is added.
- The benchmark for `execute` still uses a `System::remark` inner call, but since `execute`'s annotated weight is now `bookkeeping(MaxCallSize) + MaxInnerCallWeight`, this is fine — benchmarks only need to cover the bookkeeping component.

## Verification

- `cargo test --package pallet-multisig --lib`: 52 passed, 0 failed
- `cargo check --package pallet-multisig` (default + `runtime-benchmarks`): clean
- `gh pr checks 546`: all 4 CI jobs pass on the latest commit (`9351bf1`)

## Recommendation

The three blocking dispatch / weight / whitelist findings are conclusively resolved with strong regression coverage. The only remaining item is an explicit answer on Critical 4 (clean-slate launch vs. migration). Once that's confirmed (or a migration added), this is good to merge.

Related: [Initial review of PR 546](19874c2c-99e0-4fa4-bf38-e08acebcae09)

This is good

n13 (approved):
GTG! 🚀

PR #552, opened 2026-04-25
Eiger scheduler audit 2

illuzen:
See issues defined in ./audits in https://github.com/Quantus-Network/chain/pull/546/changes


n13:
# PR #552 Review — Eiger Substrate & Node Audit Remediation

PR: https://github.com/Quantus-Network/chain/pull/552
Audit: `audits/quantus-substrate-audit-20260421.pdf`

## Coverage Summary

The Eiger audit identified **44 findings**: 1 high, 6 medium, 14 low, and 23 informational. PR #552 contains **30 commits** and addresses approximately **27 of 44** findings. The reversible-transfers chapter and several scheduler/multisig findings remain unaddressed.

### Coverage Map

| Finding | Severity | PR Status | Commit | Assessment |
|---|---|---|---|---|
| `RUNTIME-O-01` | Info | Addressed | `753bd78` | Good — `defensive!` macros distinguish invariant violations |
| `RUNTIME-O-02` | Info | Documented | `fa3383e` | Doc-only; explicit decision rationale |
| `RUNTIME-O-03` | Info | Addressed | `7d53760` | Storage versions declared (multisig=0, scheduler=4) |
| `TREASURY-F-01` | Med | Pre-addressed | (PR #434) | OK |
| `TREASURY-F-02` | Med | Pre-addressed | (PR #434) | OK |
| `TREASURY-F-03` | Low | Addressed | `cce06fc` | Test genesis now initialises treasury |
| `TREASURY-O-01` | Info | Partial | `6061d1a` | Only docs+event update; no balance migration |
| `TREASURY-O-02` | Info | Pre-addressed | (PR #434) | OK |
| `TREASURY-O-03` | Info | **Not addressed** | — | Standalone pallet remains |
| `TREASURY-O-04` | Info | **Not addressed** | — | Mint failures still silent |
| `MULTISIG-F-01` | **High** | **Regression** | `0dba30e` | See below |
| `MULTISIG-F-02` | Low | **Not addressed** | — | Dynamic-fee floor still early |
| `MULTISIG-F-03` | Low | Addressed | `3a6db34` | `DispatchResultWithPostInfo` returns added |
| `MULTISIG-F-04` | Low | **Not addressed** | — | `claim_deposits` weight still off |
| `MULTISIG-F-05` | Low | Addressed | `8f0b7a3` | `total_returned` uses real unreserved sum |
| `MULTISIG-F-06` | Low | Addressed | `5580b5f`, `348f851`, `dd13871` | **Re-architected — see risks** |
| `MULTISIG-F-07` | Low | Mooted | `dd13871` | Dissolution removed altogether |
| `MULTISIG-O-01` | Info | Addressed | `5049a83` | Decode pulled earlier; phase-numbered |
| `MULTISIG-O-02` | Info | Addressed | `ed9d50f` | `normalize_signers` consolidates sort+dedup |
| `MULTISIG-O-03` | Info | Addressed | `95c1343` | `Vec<u8>` → `BoundedCallOf<T>` |
| `MULTISIG-O-04` | Info | Addressed | `8ad628b` | `ProposalApproved`→`SignerApproved`, status simplified |
| `MULTISIG-O-05` | Info | Addressed | `b7e2153` | `checked_add` + `ProposalNonceExhausted` |
| `MULTISIG-O-06` | Info | Addressed | `2d9206f` | Removes redundant `active_proposals` |
| `MULTISIG-O-07` | Info | Addressed | `9062d5b` | README updated |
| `MULTISIG-O-08` | Info | Addressed | `325a0c9` | Documents fee-attribution decision |
| `MULTISIG-O-09` | Info | Addressed | `9eefa1c` | +394 lines of edge-case tests |
| `MULTISIG-O-10` | Info | **Not addressed** | — | Legacy `Currency` still used |
| `MULTISIG-O-11` | Info | Addressed | `eb91c0f` | Min 2 unique signers |
| `MULTISIG-O-12` | Info | Addressed | `30963d9` | `checked_div` + defensive! |
| `SCHEDULER-F-01` | Med | Addressed | `d0b94cd` | Bucket+1 ensures After>target; docs greatly improved |
| `SCHEDULER-F-02` | Med | **Not addressed** | — | Split-counter "permanently overweight" issue remains |
| `SCHEDULER-F-03` | Low | Addressed | `10091fb` | New `RetryPeriodMismatch`, upfront validation |
| `SCHEDULER-F-04` | Med | **Not addressed** | — | Retry config still lost on reschedule |
| `SCHEDULER-F-05` | Med | **Not addressed** | — | Inherited upstream behaviours |
| `SCHEDULER-F-06` | Med | Addressed | `2801399` | Preimage ownership preserved on retry; +tests |
| `SCHEDULER-F-07` | Low | Addressed | `6693864` | Timestamp base weight in `on_initialize` |
| `SCHEDULER-F-08` | Low | **Not addressed** | — | v3 shim panics still present |
| `SCHEDULER-O-01` | Info | Addressed | `caa25ad` | Cleanup, removed `traits` module |
| `SCHEDULER-O-02` | Info | Addressed | `163bb31` | +286 lines of timestamp tests |
| `REVERSIBLE-F-01` | Med | **Not addressed** | — | HS effectively permanent |
| `REVERSIBLE-F-02` | Low | **Not addressed** | — | Asset path reuses native weight |
| `REVERSIBLE-F-03` | Low | **Not addressed** | — | Orphaned scheduled holds |
| `REVERSIBLE-O-01` | Info | **Not addressed** | — | Naming/doc drift |
| `REVERSIBLE-O-02` | Info | **Not addressed** | — | README terminology drift |

## Critical Regressions & Risks

### 1. MULTISIG-F-01 (the only HIGH) — fix is incomplete and creates the same vulnerability the audit flagged

The pre-dispatch weight declaration is essentially unchanged across the PR:

```rust
// pallets/multisig/src/lib.rs:989-993
#[pallet::weight({
    // Worst case: max bookkeeping + max possible call weight (from benchmarks)
    // The actual weight will be refunded based on the real call's weight
    <T as Config>::WeightInfo::execute(T::MaxCallSize::get())
})]
```

Inspecting the benchmark, `WeightInfo::execute(c)` measures only the multisig pallet's own bookkeeping plus a per-byte decode coefficient — there is **no inner-call weight component**:

```rust
// pallets/multisig/src/weights.rs:139-149
fn execute(c: u32, ) -> Weight {
    Weight::from_parts(35_758_975, 17022)
        .saturating_add(Weight::from_parts(755, 0).saturating_mul(c.into()))
        .saturating_add(T::DbWeight::get().reads(2_u64))
        .saturating_add(T::DbWeight::get().writes(2_u64))
}
```

Post-dispatch, the code does compute `bookkeeping + actual_call_weight` and returns it as `actual_weight`. But Substrate's `PostDispatchInfo.actual_weight` can only **refund** weight; it cannot exceed the pre-dispatch declaration. So:

- A `total_weight > WeightInfo::execute(MaxCallSize)` is silently clamped down at the dispatcher layer — fee charged is based on pre-dispatch weight, which omits the inner call's cost.
- The block-weight reservation is still based on the (small) pre-dispatch weight, so spammy `execute()` calls can still over-pack a block with under-priced compute.

This is the *exact* economic vulnerability EQ-QNT-MULTISIG-F-01 identified, now with code that **looks** fixed but isn't. The correct fix needs the inner call available at the weight-macro site — e.g. require it as a parameter to `execute`, or charge `MaximumBlockWeight` and refund precisely at post-dispatch.

### 2. Storage layout broken without migration

`MULTISIG-O-03` declared `StorageVersion::new(0)` as the explicit baseline. Subsequent commits then **deleted two fields from `MultisigData`** without bumping the version or providing a migration:

- `2d9206f4` (`MULTISIG-O-06`) — removed `active_proposals: u32` (replaced by a derived method).
- `dd138716` (`Fully remove deposit`) — removed `deposit: Balance` and the `Balance` generic parameter.

```rust
// pallets/multisig/src/lib.rs:151-155
const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);

#[pallet::pallet]
#[pallet::storage_version(STORAGE_VERSION)]
```

SCALE codec is positional — removing fields from a stored struct invalidates existing encoded values. The multisig pallet has been on the chain since the Planck Testnet (`49365469`), so any state on testnet/staging will fail to decode after upgrade. Either:

- this PR is meant to land **before** any production multisig data exists (treat as genesis baseline — copy the scheduler's "no migration path" rationale comment), or
- a `OnRuntimeUpgrade` migration is required (decode-old → drop fields → re-store) plus a `STORAGE_VERSION` bump.

The scheduler did this correctly:

```rust
// pallets/scheduler/src/lib.rs:198-206
/// Note: There is no migration path FROM upstream v4 TO this fork's v4, as they have
/// incompatible storage layouts. This fork's v4 is the genesis baseline for this chain.
const STORAGE_VERSION: StorageVersion = StorageVersion::new(4);
```

Multisig should make the same explicit decision and document it.

### 3. Multisig dissolution silently removed — major design change buried in PR

Commits `348f851` (*"just remove the dissolve thing"*) and `dd13871` (*"Fully remove deposit"*) eliminated the entire dissolution mechanism, the `MultisigDeposit`, `DissolveApprovals` storage, and `MultisigDissolved`/`DissolveApproved` events. `MultisigFee` was raised from `100 * MILLI_UNIT` to `600 * MILLI_UNIT` to compensate.

This goes well beyond the audit's recommendation (which only asked F-06/F-07 to be made robust). The new policy — *"multisigs are permanent, deposits are non-refundable storage rent"* — is a fundamental product change that:

- Breaks the contract for any creator who already paid a refundable deposit on testnet.
- Eliminates a recovery path for funds inadvertently sent to a multisig.
- Should be called out explicitly in PR description / release notes (currently only in README).

### 4. Behaviour changes that may surprise integrators

- `MULTISIG-O-02`: duplicate signers are now silently deduped instead of returning `DuplicateSigner`. Any client relying on that error path now sees `Ok(...)`.
- `MULTISIG-O-04`: `ProposalApproved` event renamed to `SignerApproved`. Indexers (SubSquid etc.) must be updated.
- `MULTISIG-O-04`: `ProposalReadyToExecute` is now emitted *only on first* threshold crossing (previously every approval after threshold). Listeners that relied on later emissions need updating.
- `MULTISIG-O-11`: minimum 2 unique signers — clients constructing 1-of-1 multisigs (previously legal) now get `NotEnoughSigners`.
- `MULTISIG-O-03`: `propose` parameter type went from `Vec<u8>` to `BoundedCallOf<T>` — extrinsic SCALE shape unchanged, but JS/TS API generated types will differ.
- `SCHEDULER-F-01`: timestamp `After` now lands one extra bucket past the target. New tasks fire later than they used to (existing in-flight tasks already have a stored bucket, so they're unaffected).

### 5. Smaller polish issues spotted in the new code

- `pallets/multisig/src/lib.rs` `propose` retains stale phase numbering: `PHASE 1` jumps to `PHASE 3` to `PHASE 4` (no `PHASE 2`).
- `pallets/multisig/README.md:7` still describes a "fee+deposit system" although the deposit was removed.
- `pallets/multisig/README.md:168` still references "enable multisig dissolution" although dissolution no longer exists.
- `pallets/multisig/src/lib.rs` the docstring for `cleanup_expired_proposals_for_signer` is duplicated (added in the rename commit, original kept).

## Strengths

- The defensive-error remediation (`RUNTIME-O-01`) is consistent across `multisig`, `scheduler`, and `reversible-transfers`.
- `SCHEDULER-F-01`/`F-06`/`O-02` add genuinely valuable documentation and tests; the bucket-after-bucket fix is correct and the preimage-ownership transfer logic is now race-free.
- `SCHEDULER-F-03` upfront-rejects mismatched retry period types, replacing a silent runtime error.
- `MULTISIG-O-01` properly hoists call decoding so non-HS multisigs also reject malformed payloads at proposal time.
- `MULTISIG-O-09`/`SCHEDULER-O-02` add ~680 new test lines covering edge cases.

## Recommended Action Items Before Merge

1. **Block on MULTISIG-F-01.** Either (a) require the inner `RuntimeCall` as a parameter so its dispatch info can feed `#[pallet::weight]`, or (b) declare worst-case total block weight upfront. Without this, the only HIGH finding is functionally unfixed despite the commit message.
2. **Add a migration story for multisig storage**, or replicate the scheduler's "fork genesis baseline" comment + bump `STORAGE_VERSION` past 0 to invalidate any leftover testnet state explicitly. Note: this is the next migration target, not 0.
3. **Pick up the unaddressed mediums**: `SCHEDULER-F-02`, `F-04`, `F-05`, `REVERSIBLE-F-01` — or document why they're deferred.
4. **Reversible-transfers entirely untouched** (5 findings). Either include fixes here or split them into a follow-up PR with explicit tracking.
5. **Highlight the dissolution-removal design change** in the PR description and runtime release notes, not just the README.
6. **README cleanup pass**: dissolve mentions on line 168, "fee+deposit" on line 7, stale phase numbering, duplicated docstring.
7. **Spec version bump.** A change of this magnitude clearly needs `spec_version: 128` → ≥ 129 with the appropriate runtime upgrade scaffolding.

## Overall Assessment

This PR is a **substantial, mostly-positive cleanup pass** that lands almost all of the multisig informational findings, the scheduler housekeeping issues, and a couple of mediums. The new tests and documentation are real wins.

However, it carries **two serious problems**:

1. The lone HIGH-severity finding (`MULTISIG-F-01`) appears fixed but is still vulnerable due to Substrate's pre-dispatch weight clamping rules. **This is a regression risk** — the issue is now harder to spot in review.
2. **Multisig storage layout was broken silently.** Without a migration or an explicit "genesis baseline" justification, deploying this to any chain with existing multisig data is unsafe.

In addition, removing dissolution is a significant policy change that deserves its own dedicated discussion rather than being tucked into an audit-remediation PR.

Recommendation: **request changes** — reopen MULTISIG-F-01, make storage versioning intent explicit (with migration if needed), surface the dissolution-removal decision, and either tackle or explicitly defer the unaddressed mediums and the entire reversible-transfers chapter.

n13:
Generally would recommend more minimal fixes because any sweeping change has the chance to introduce new issues that are not in the review... 

We need to fix msig weights

The other msig fixes are a bit of a wash - is it really better than before in any substantial way? I will need to re-review this myself again to figure out if this is now all good. We had a pretty solid system before, it may have been over the top, but we considered all the cases, and it was orthogonal in functionality. 

Preventing the dissolution of a msig by sending small assets there seems to be a very small edge case? I don't think it's worth worrying about. Assets getting stuck in non-existient msig likewise seems like a trivial concern.

Agree that we don't need to fix scheduler issues that are already present in the upstream. Before we do that we could instead design a minimal scheduler pallet that's more simple - the OG substrate scheduler was already too complicated, and our changes with time based scheduling made it even more complicated. For now, doing nothing on this is the best option, as fixing anything in this has a very high chance of introducing new errors.

ethan-crypto (changes requested):
## Blockers

### 1. `multisig::execute` still appears to under-declare pre-dispatch weight

The `execute` extrinsic’s weight declaration still only accounts for multisig bookkeeping / encoded call size, but not the stored inner call’s `call.get_dispatch_info().call_weight`.

The code computes the bookkeeping weight plus the actual inner call weight after dispatch and returns that as `actual_weight`, but that does not fully address the original issue. FRAME uses the static `#[pallet::weight]` declaration for pre-dispatch block weight accounting; post-dispatch weight is primarily a refund/correction mechanism, not a way to reserve additional block capacity after the inner call has already executed.

Suggested fix: make the inner call or encoded call hash available at the weight declaration site, verify it matches the stored proposal, and declare:

```rust
bookkeeping_weight + call.get_dispatch_info().call_weight
```

up front.

Alternatively, charge a conservative maximum upfront and refund precisely, though that is less ergonomic.

### 2. Multisig storage layout changes need a migration or an explicit genesis-baseline decision

This PR removes fields from `MultisigData` and removes dissolution-related storage/events, but I do not see a corresponding storage migration. `STORAGE_VERSION` also still appears to be `StorageVersion::new(0)`.

If this runtime can be upgraded over existing state, this is a storage-compatibility issue. Please either:

* add an `OnRuntimeUpgrade` migration from the previous encoded layout to the new one, or
* explicitly document that this pallet state is only safe as a genesis/testnet reset baseline and not an in-place runtime upgrade.

The scheduler pallet documents its forked storage-version baseline clearly; multisig should do the same or provide a migration.

### 3. `execute` docs and implementation disagree on inner-call failure behavior

The README says the `execute` extrinsic itself succeeds even if the inner call fails, and users should inspect `ProposalExecuted.result`.

The implementation appears to map the inner dispatch error back into the outer extrinsic result. That means callers/UIs will see the extrinsic as failed even though the proposal has already been removed and deposit handling has already occurred.

Please choose one contract and make code/docs/tests align:

* If the intended contract is “proposal is consumed and the event carries the inner result,” then `execute` should return `Ok(PostDispatchInfo { ... })` regardless of the inner call result.
* If the intended contract is “outer extrinsic fails when the inner call fails,” then the README/tests should be updated, and the state-persistence behavior should be made explicit.

### 4. Runtime version bump appears missing

This PR changes runtime behavior and runtime files, but I do not see a corresponding `spec_version` bump. Please include the runtime version bump and any associated upgrade notes/scaffolding.

## High-priority follow-ups

### 5. `propose` charges fee/reserves deposit before the final fallible mutation

`propose` withdraws the non-refundable fee and reserves the proposal deposit before `Multisigs::try_mutate`, whose closure can still return errors such as nonce exhaustion or insertion failure.

This is probably low-likelihood, but it is still a correctness footgun unless the function is transactional. Please either:

* move fee/deposit changes after all remaining fallible checks, or
* wrap the extrinsic/function with transactional storage rollback semantics.

### 6. Dissolution removal is a product/API change, not just audit remediation

This PR removes dissolution functionality, including `approve_dissolve`, `MultisigDeposit`, dissolution approvals, and dissolved-related events. The README now states multisigs are permanent and the creation fee is never returned.

That may be a valid design choice, but it should be called out clearly in the PR description and release notes. Ideally, it should be split from the audit-remediation PR unless maintainers explicitly want this policy change bundled here.

## Smaller issues

* `approve` is documented as requiring the proposal to be `Active`, but the implementation does not appear to enforce `proposal.status == Active`. It seems to allow additional approvals on an already `Approved` proposal until the approvals vector fills. That may be intentional, but the docs and tests should match the implementation.
* `claim_deposits` calculates and passes `avg_call_size`, but the generated weight function appears to ignore `_c`. Either remove the unused parameter from the model or regenerate/fix the benchmark if call size is supposed to affect cost.
* GitHub flags the multisig README as containing hidden or bidirectional Unicode text. Please clean that up before merge, especially in an audit remediation PR.

## Things that look good

* The scheduler timestamp `After` behavior now normalizes to a bucket and advances one more bucket, which matches the “strictly after target” remediation.
* The scheduler retry reschedule path looks improved: both anonymous and named reschedule paths appear to transfer retry configuration from the old task address to the new one.
* The scheduler preimage ownership handling on retry is a solid improvement: it only drops the original preimage reference when no retry was scheduled, preserving ownership when a retry clone is created.

## Suggested verdict

**Request changes.** I would not merge until at least the `execute` pre-dispatch weight issue, multisig storage migration/baseline decision, `execute` success/error contract, and runtime version bump are resolved.

illuzen:
looks like the multisig review items got moved onto this scheduler audit

PR #555, opened 2026-04-27
feat: Miner-Stack - update

czareko:
Brings `miner-stack/` up to date with the current node + miner contracts.
### Node ↔ miner protocol
- Switch from the deprecated HTTP `--external-miner-url` to the new QUIC-based
  protocol: node now runs as the QUIC server (`--miner-listen-port 9833`) and
  miners connect to it as clients via `serve --node-addr`.
- Pin `quantus-node` to a static IPv4 (`172.28.0.10`) on the bridge network —
  `quantus-miner` requires a `SocketAddr` for `--node-addr` and cannot resolve
  the Docker DNS name on its own.
- Expose `9833/udp` on the host so miners on other machines can attach.
### Rewards
- Replace SS58 `REWARDS_ADDRESS` with `REWARDS_INNER_HASH` (32-byte wormhole
  preimage), matching the current node CLI. The wormhole address is derived
  by the runtime.
### Configuration
- New env knobs: `CPU_WORKERS`, `GPU_DEVICES`, `MINER_LOG`,
  `MINER_LISTEN_PORT`, `MINER_METRICS_PORT`.
- `.env.example` cleaned up; image versions default to `latest`.
### Docs & dashboards
- README rewritten: setup, prerequisites (wormhole keypair generation),
  monitoring, multi-miner topology, troubleshooting.
- Refreshed `quantus-miner` Grafana dashboard.

### DIFF
--- a/.github/workflows/miner-stack-compose.yml
+++ b/.github/workflows/miner-stack-compose.yml
@@ -1,0 +1,63 @@
+name: miner-stack compose config
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  config:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Render default Compose config
+        working-directory: miner-stack
+        env:
+          REWARDS_INNER_HASH: "0x0000000000000000000000000000000000000000000000000000000000000000"
+        run: docker compose config >/tmp/quantus-miner-stack.default.yml
+
+      - name: Check default Compose config
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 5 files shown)


ethan-crypto (changes requested):
Everything looks good other than some small concerns I have regarding env-var semantics:

- The docs describe MINER_LISTEN_PORT / MINER_METRICS_PORT as service ports but they appear to be host-port mappings. 
- CPU_WORKERS documents auto-detect while Compose forces 4 workers

I think we should either wire these through to the actual commands or rename/document them as host-only variables.

illuzen (approved):

ethan-crypto (approved):

PR #557, opened 2026-04-29
Eiger scheduler audit 3

illuzen:
Mostly cosmetic, lots of comments to explain things for future auditors, but two bugs fixed

- retry failure was not handled correctly
- retry config did not carry period type forward

and made weights more accurate


n13:
AI is raving about this PR! 

Approved.


## Summary

This PR addresses 6 Eiger audit findings (4 findings + 2 observations) on `pallet-scheduler`. The two real bug fixes are excellent and well-tested:

1. **F-07: Preimage refcount leak/drop on retry** — solid fix
2. **F-01: Silent failure when retry period type doesn't match task type** — defensively rejected at set-time

The rest is documentation, weight accuracy improvements, code organization (moving `ScheduleNamed` trait to primitives), and substantial new test coverage. ([illuzen/eiger-scheduler-audit-3](https://github.com/Quantus-Network/chain/pull/557))

## What's solid

### F-07 — preimage ownership tracking is the headline fix

This is genuinely subtle. `Bounded<Call>` is a hash+length pair; `as_retry()` clones the *struct* but the preimage refcount is shared, not bumped. The OLD `service_task` flow was:

```57:79:pallets/scheduler/src/lib.rs
// OLD
if let Some(retry_config) = maybe_retry_config {
    if failed { Self::schedule_retry(...); }
}
T::Preimages::drop(&task.call);  // ALWAYS drops — buggy
```

That always dropped one refcount, even when a retry clone now needed that same reference. And `schedule_retry`'s failure branch did its *own* `Preimages::drop(&task.call)`, leading to a double-drop in the place_task-fails case.

New flow returns `bool` from `schedule_retry` indicating whether ownership transferred:

```1192:1207:pallets/scheduler/src/lib.rs
				let retry_scheduled = if let Some(retry_config) = maybe_retry_config {
					if failed {
						Self::schedule_retry(weight, now, when, agenda_index, &task, retry_config)
					} else {
						// Task succeeded, no retry needed
						false
					}
				} else {
					// No retry config
					false
				};

				// Only drop preimage if retry wasn't scheduled (ownership not transferred)
				if !retry_scheduled {
					T::Preimages::drop(&task.call);
				}
```

Three branches in `schedule_retry`, all correct:
- **place_task succeeds** → `true`, caller skips drop, retry clone keeps the ref.
- **place_task fails** (`Err((_, retry_task))`) → `false`, no internal drop (the retry clone never acquired its own refcount), caller drops the original. The comment makes this reasoning explicit.
- **`saturating_add` returns Err** (period type mismatch — guarded against by F-01 but defense-in-depth) → `false`, caller drops.

The two test cases `retry_preserves_preimage_ownership` and `retry_unrequests_preimage_on_success` directly exercise the refcount via `Preimage::is_requested(&hash)` across the retry chain. Good coverage.

### F-01 — retry period type mismatch now rejected upfront

Before, you could `set_retry` with a `Timestamp` period on a `BlockNumber`-scheduled task; the mismatch only surfaced inside `schedule_retry` via `now.saturating_add(&period) → Err(IncompatibleTypesError)`, which silently bailed out without firing a `RetryFailed` event. Now both `set_retry` and `set_retry_named` call:

```1340:1346:pallets/scheduler/src/lib.rs
		let types_match = matches!(
			(task_when, period),
			(BlockNumberOrTimestamp::BlockNumber(_), BlockNumberOrTimestamp::BlockNumber(_)) |
				(BlockNumberOrTimestamp::Timestamp(_), BlockNumberOrTimestamp::Timestamp(_))
		);
		ensure!(types_match, Error::<T>::RetryPeriodMismatch);
```

The new `RetryPeriodMismatch` error is the right name. Two new tests (`mismatched_retry_period_rejected_at_set_retry`, `mismatched_retry_period_rejected_for_timestamp_task`) verify both directions. This aligns nicely with the user's "fail early" principle.

### Weight accuracy: `service_timestamp_agendas_base`

`on_initialize` was previously doing the timestamp agenda housekeeping (read/take/put on `IncompleteTimestampSince` and `LastProcessedTimestamp`) without charging for it. Now there's a proper benchmark and the weight is consumed before the call:

```379:390:pallets/scheduler/src/lib.rs
			if current_timestamp > T::Moment::zero() {
				// Consume base weight for timestamp-based agenda housekeeping
				// (IncompleteTimestampSince and LastProcessedTimestamp reads/writes)
				if weight_counter
					.try_consume(T::WeightInfo::service_timestamp_agendas_base())
					.is_err()
				{
					return weight_counter.consumed();
				}
```

Storage path comments in `weights.rs` were also corrected: `IncompleteSince` → `IncompleteBlockSince` (the actual storage name).

### Trait relocation

`ScheduleNamed` moved from `pallets/scheduler/src/traits.rs` (deleted) into `primitives/scheduler/src/lib.rs`, alongside `BlockNumberOrTimestamp` and `DispatchTime`. Cleaner — the trait is conceptually a primitive shared across consumers. No external import sites needed updating because downstream code already imports via `qp_scheduler::ScheduleNamed`.

### Documentation additions

The `DispatchTime`/`After(Timestamp)` semantics doc on the enum is genuinely helpful — flags that `Timestamp(x)` is *absolute*, not a relative delay, and that bucket normalization adds bucket-sized granularity. This is exactly the kind of footgun documentation that prevents downstream pallet authors from making mistakes.

### Test coverage

Significant additions (~470 lines), covering:
- `retry_preserves_preimage_ownership` — refcount through 3-retry exhaustion
- `retry_unrequests_preimage_on_success` — refcount cleanup on success
- `mismatched_retry_period_rejected_at_set_retry`, `mismatched_retry_period_rejected_for_timestamp_task` — F-01
- `schedule_after_timestamp_zero_rejected`, `schedule_after_timestamp_zero_at_genesis` — edge cases
- `timestamp_large_time_gap_processes_multiple_buckets` — multiple-bucket catch-up
- `cancel_task_after_retry_rescheduling` — confirms named lookup is gone post-retry (and documents how to cancel by `(when, index)` instead)
- `timestamp_retry_scheduling_and_cancellation` — timestamp-domain retry path

## Issues / nits

These are all minor — none are blockers.

### 1. Weight overcount in `service_timestamp_agendas_base` (minor, conservative)

The weight claims `r:1 w:1` for `LastProcessedTimestamp`, but in the `IncompleteTimestampSince = Some` path (which the benchmark exercises), `LastProcessedTimestamp::get()` is *not* called — only `put()`. So actual access in that path is r:0 w:1 for that storage. In the `None` path it would be r:1 w:1.

```128:144:pallets/scheduler/src/weights.rs
	/// Storage: `Scheduler::IncompleteTimestampSince` (r:1 w:1)
	/// Proof: `Scheduler::IncompleteTimestampSince` (`max_values`: Some(1), `max_size`: Some(9), added: 504, mode: `MaxEncodedLen`)
	/// Storage: `Scheduler::LastProcessedTimestamp` (r:1 w:1)
	/// Proof: `Scheduler::LastProcessedTimestamp` (`max_values`: Some(1), `max_size`: Some(9), added: 504, mode: `MaxEncodedLen`)
	fn service_timestamp_agendas_base() -> Weight {
```

The weight chooses the worst-case path, which is the right defensive choice. Just flagging that the benchmark setup steers into the cheaper path. If you want strict accuracy you could parameterize the benchmark over both branches; otherwise this is fine.

### 2. `cancel_task_after_retry_rescheduling` documents a UX gap

Once a named task fails and a retry is scheduled, the named `Lookup` entry is removed (because `as_retry()` produces an unnamed clone). The test even asserts:

```pallets/scheduler/src/tests.rs
		// Try to cancel by name - should fail since lookup is gone
		assert_noop!(Scheduler::cancel_named(root().into(), task_name), Error::<Test>::NotFound);
```

This is *correct behavior* given the design, but it means a caller who set up retries and wants to cancel the whole task graph has to know the unnamed retry's `(when, index)` — they can't just `cancel_named`. Worth a doc comment on `set_retry_named` calling out this implication, since the test is implicitly asserting the foot-gun. Not a blocker, just a thought.

### 3. `service_task` retry/drop block could be a tiny bit tighter

```1192:1207:pallets/scheduler/src/lib.rs
				let retry_scheduled = if let Some(retry_config) = maybe_retry_config {
					if failed {
						Self::schedule_retry(weight, now, when, agenda_index, &task, retry_config)
					} else {
						// Task succeeded, no retry needed
						false
					}
				} else {
					// No retry config
					false
				};
```

Could collapse to:

```rust
let retry_scheduled = maybe_retry_config
    .filter(|_| failed)
    .map(|cfg| Self::schedule_retry(weight, now, when, agenda_index, &task, cfg))
    .unwrap_or(false);
```

…but the current form is more readable for the audit context (clear comments on each branch). I'd leave it.

### 4. `IncompleteSince` legacy name in storage proof comments (already fixed — verifying)

The diff updates the comments to `IncompleteBlockSince`. Confirmed correct against the actual storage definitions at lines 268/272/277 of `lib.rs`. ✅

### 5. `service_timestamp_agendas_base` benchmark exits with `max = 0`

```pallets/scheduler/src/benchmarking.rs
	}: {
		Scheduler::<T>::service_timestamp_agendas(&mut WeightMeter::new(), &mut 0, current_time, 0);
	}
```

Calling with `max = 0` makes the `while count_down > 0 && ...` loop body never execute, so the benchmark really does isolate the housekeeping. Good. (The `verify` block correctly checks `LastProcessedTimestamp` got updated.)

### 6. Saturating_add error path in `schedule_retry` is now unreachable

After F-01, `set_retry`/`set_retry_named` reject mismatches at config time, so the `Err(_) =>` branch in `schedule_retry` for `now.saturating_add(&period)` should never fire in practice. The branch is kept for defense-in-depth, which is fine — but consider a `log::error!("invariant violation: retry period type mismatch reached schedule_retry")` to surface an unreachable case if it ever does. Per the user-rule "fail early — never have a silent failure", this branch currently emits `RetryFailed` and returns `false`, which is the right behavior; an extra warn-level log line about the invariant violation would help future debugging.

## Recommendation

**Approve.** The two bug fixes (F-07, F-01) are correct, well-reasoned, well-commented, and well-tested. Documentation improvements are genuinely useful, weight accounting is more honest, and the trait reorganization is clean. The 78-test scheduler suite passes, and downstream `pallet-reversible-transfers` builds cleanly against the new primitives location.

Items 1, 2, 6 above are post-merge polish — none should block.

Nice work landing this — the F-07 preimage ownership reasoning was genuinely tricky, and the comments make it easy for future auditors to verify the invariant.

n13 (approved):
Good changes, approved

PR #562, opened 2026-05-13
replace static variable with feature for fast governance during testing

n13:
Every audit flagged this, so removing the static governance override

### DIFF
--- a/runtime/src/governance/definitions.rs
+++ b/runtime/src/governance/definitions.rs
@@ -80,57 +80,26 @@ impl Consideration<AccountId, Footprint> for PreimageDeposit {
 	}
 }
 
-/// Global dynamic configuration for ALL governance tracks
-/// This allows tests to override the production timing values at runtime
-pub struct GlobalTrackConfig;
+/// Collapses every referenda timing window to 2 blocks when the
+/// `fast-governance` feature is enabled; otherwise leaves the track untouched.
+/// Selected at compile time so production builds carry no override state.
+#[cfg(feature = "fast-governance")]
+fn apply_test_timing(
+	mut info: pallet_referenda::TrackInfo<Balance, BlockNumber>,
+) -> pallet_referenda::TrackInfo<Balance, BlockNumber> {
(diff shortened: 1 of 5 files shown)


runtime/src/governance/definitions.rs
```diff
@@ -80,57 +80,26 @@ impl Consideration<AccountId, Footprint> for PreimageDeposit {
...
-			track.prepare_period = prepare;
-			track.decision_period = decision;
-			track.confirm_period = confirm;
-			track.min_enactment_period = enactment;
-		}
-		track
-	}
+#[cfg(not(feature = "fast-governance"))]
+#[inline]
+fn apply_test_timing(
```
illuzen: either this function is named poorly (test code running in prod?)  or maybe just feature gate a {} scope?

n13: Fair - should be possible to just not calls this when not in test mode

n13: Fixed

runtime/tests/common.rs
```diff
@@ -50,28 +50,16 @@ impl TestCommons {
...
-			use quantus_runtime::governance::definitions::GlobalTrackConfig;
-
-			// Set global fast timing for ALL governance tracks (Community, Treasury, Tech
-			// Collective)
-			GlobalTrackConfig::set_fast_test_timing(); // Sets 2 blocks for all periods
-
-			println!("Fast governance test config activated: All tracks use 2-block periods");
-			Self::new_test_ext()
-		}
+		#[cfg(feature = "fast-governance")]
```
illuzen, line 58: seems like we don't need the println or feature gating

illuzen (approved):

runtime/tests/common.rs
n13: This is just in tests, should be in there in case someone runs the tests with production timing and wonders why nothing is happening...

n13: it's testing code anyway

PR #566, opened 2026-05-21
Fix multi arch docker image building

dewabisma:
## Summary
- Detect arch on the fly instead of hard coded for x86_64
- Update action version for docker build and push to v6
- Add linker and cross compile to arch64 linux

---

> [!NOTE]
> **Low Risk**
> Low risk: small GitHub Actions workflow change limited to the `aarch64-unknown-linux-gnu` hotfix build configuration.
> 
> **Overview**
> Ensures the hotfix release workflow can reliably cross-compile for `aarch64-unknown-linux-gnu` by exporting `CC_aarch64_unknown_linux_gnu` and `CXX_aarch64_unknown_linux_gnu` in addition to `CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5bbfa176e57fe9d04428a43cc4114bfafa58a74d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/ubuntu/action.yml
+++ b/.github/actions/ubuntu/action.yml
@@ -13,6 +13,7 @@ runs:
           clang \
           protobuf-compiler \
           gcc-aarch64-linux-gnu \
+          g++-aarch64-linux-gnu \
           libc6-dev-arm64-cross
         rustup update
         rustup target add wasm32-unknown-unknown
--- a/.github/actions/ubuntu/action.yml
+++ b/.github/actions/ubuntu/action.yml
@@ -11,8 +11,9 @@ runs:
         sudo apt-get install -yqq --no-install-recommends \
           libclang-dev \
           clang \
-          protobuf-compiler
-          gcc-aarch64-linux-gnu
+          protobuf-compiler \
+          gcc-aarch64-linux-gnu \
+          libc6-dev-arm64-cross
         rustup update
         rustup target add wasm32-unknown-unknown
         rustup target add aarch64-unknown-linux-gnu
--- a/.github/actions/ubuntu/action.yml
+++ b/.github/actions/ubuntu/action.yml
@@ -12,8 +12,10 @@ runs:
           libclang-dev \
           clang \
           protobuf-compiler
+          gcc-aarch64-linux-gnu
         rustup update
         rustup target add wasm32-unknown-unknown
+        rustup target add aarch64-unknown-linux-gnu
         rustup component add rustfmt --toolchain nightly
         rustup component add clippy rust-src
       shell: bash
--- a/.github/workflows/quantus-docker-image.yml
+++ b/.github/workflows/quantus-docker-image.yml
@@ -103,7 +103,7 @@ jobs:
           path: source_code_for_build
 
       - name: Build and push Docker image from Dockerfile
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v6
         with:
           context: ./source_code_for_build
           file: ./Dockerfile
--- a/.github/workflows/quantus-hotfix-release.yml
+++ b/.github/workflows/quantus-hotfix-release.yml
@@ -65,6 +65,8 @@ jobs:
         include:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
+          - target: aarch64-unknown-linux-gnu
+            os: ubuntu-latest
           - target: aarch64-apple-darwin
             os: macos-latest
           - target: x86_64-apple-darwin
@@ -88,6 +90,10 @@ jobs:
       - name: Setup Windows
         if: runner.os == 'Windows'
         uses: ./.github/actions/windows
+
+      - name: Set ARM64 cross-linker
+        if: matrix.target == 'aarch64-unknown-linux-gnu'
+        run: echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc" >> $GITHUB_ENV
 
       - name: Build quantus-node binary
         shell: bash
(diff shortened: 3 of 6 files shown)


n13:
Here's my review of [PR #566 — Fix multi arch docker image building](https://github.com/Quantus-Network/chain/pull/566).

## TL;DR
Direction is good and the `Dockerfile.local` cleanup is a nice independent fix, but the runtime `Dockerfile` change ships a build that will 404 on `linux/arm64`. The Cursor Bugbot's "ARM64 release asset missing" call is correct and is a **blocker**. The docker image workflow only runs on `workflow_dispatch`, so CI on this PR will not catch it — it needs a manual dry run before merge.

---

## 🚨 Blocker: no `aarch64-unknown-linux-gnu` asset is published

In the new `Dockerfile` the case maps `arm64 → aarch64-unknown-linux-gnu`:

```16:28:Dockerfile
ARG VERSION_ARG # Expecting format like vX.Y.Z
RUN set -eux; \
    DPKG_ARCH="$(dpkg --print-architecture)"; \
    case "$DPKG_ARCH" in \
        amd64)   ARCH="x86_64-unknown-linux-gnu" ;; \
        arm64)   ARCH="aarch64-unknown-linux-gnu" ;; \
        *) echo "Unsupported architecture: $DPKG_ARCH" && exit 1 ;; \
    esac; \
    echo "Downloading version: ${VERSION_ARG} for architecture: ${ARCH}"; \
    curl -fsSL "https://github.com/Quantus-Network/chain/releases/download/${VERSION_ARG}/quantus-node-${VERSION_ARG}-${ARCH}.tar.gz" \
        | tar -xzC /usr/local/bin/; \
    chmod +x /usr/local/bin/quantus-node
```

…but the release workflow only builds these targets:

```71:80:.github/workflows/quantus-release.yml
    strategy:
      matrix:
        include:
          - target: x86_64-unknown-linux-gnu
            os: ubuntu-latest
          - target: aarch64-apple-darwin
            os: macos-latest
          - target: x86_64-apple-darwin
            os: macos-15-intel
          - target: x86_64-pc-windows-msvc
            os: windows-latest
```

There is no `aarch64-unknown-linux-gnu` entry, so `https://github.com/Quantus-Network/chain/releases/download/${VERSION}/quantus-node-${VERSION}-aarch64-unknown-linux-gnu.tar.gz` doesn't exist. Combined with the docker workflow's platform list `linux/amd64,linux/arm64` ([line 111](https://github.com/Quantus-Network/chain/blob/main/.github/workflows/quantus-docker-image.yml#L111)), the `linux/arm64` leg will 404 on `curl -fsSL` and break the multi-arch publish — which is the exact thing this PR claims to fix.

### Fix options (pick one)

1. **Preferred — produce the asset.** Add an `aarch64-unknown-linux-gnu` row to the `build-and-release` matrix in `quantus-release.yml`. Either use a native ARM runner (e.g. `ubuntu-24.04-arm`) or cross-compile from `ubuntu-latest` (install `gcc-aarch64-linux-gnu`, set `CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER`, etc.). Once that asset ships in a release, this Dockerfile change works as intended.
2. **Stopgap — narrow the platforms.** In `quantus-docker-image.yml`, drop `linux/arm64` from `platforms:` until ARM64 binaries are published. Then this PR becomes effectively a no-op for the runtime `Dockerfile` (which is fine to land for the `Dockerfile.local` and v6 bump).

If you go with option 1, please also run the `Quantus - Docker Image` workflow manually against the first release that includes the new asset to confirm both legs succeed before relying on it for `:latest`.

---

## ✅ Good

- **`Dockerfile.local` `LIBCLANG_PATH`**: `/usr/lib/llvm-14/lib` is the right, architecture-agnostic path on `rust:bookworm` (libclang-dev installs the symlink there for both amd64 and arm64). The inline comment explaining the rationale is helpful.
- **`set -eux`** in the download `RUN` gives much better failure diagnostics than the previous `&&`-chain.
- **`docker/build-push-action@v5 → v6`** is a safe bump for this workflow. Heads-up: v6 enables provenance/SBOM attestations by default. For GHCR that's fine, but if you want to keep manifests slim, add `provenance: false` (and/or `sbom: false`) under `with:`.

---

## Minor nits (non-blocking)

- Both `Dockerfile` and `Dockerfile.local` lost their trailing newline at EOF (the diff shows `\ No newline at end of file`). Worth restoring — POSIX text-file convention and avoids the noisy diff next time someone edits them.
- The download URL hard-codes `…-${ARCH}.tar.gz` where `ARCH` is the Rust triple. That's consistent with current release naming, just be aware that if you ever rename release assets (e.g. swap to `…-linux-arm64.tar.gz`), this case statement has to move in lockstep.
- The `ARG VERSION_ARG # Expecting format like vX.Y.Z` line — the `#` portion is parsed by Docker as part of the default-value/comment grammar in some versions; it works today on BuildKit but a separate `# comment` line above the `ARG` is more portable. Pre-existing, just flagging.
- The `case` already has an explicit `*) … exit 1` branch — nice, matches the project's "fail early" stance.

---

## What I'd ask before approving

1. Confirm the plan for publishing `aarch64-unknown-linux-gnu` (option 1) or temporarily narrow the docker platforms (option 2).
2. Manually trigger `Quantus - Docker Image` against a known-good tag and paste the run link, since regular PR CI does not exercise this path.

Happy to push a follow-up commit on this branch that either adds the ARM64 matrix entry to `quantus-release.yml` or trims `platforms:` to `linux/amd64` — let me know which direction you want.

dewabisma:
I have tested the updated CI/CD.

I also tested to run the generated docker image in my mac m4 air. It works fine!

The POC fork is there https://github.com/Quantus-Network/chain-docker-macos/pkgs/container/quantus-node-test

## Screenshot
- Running image generated by CI/CD
image

n13 (approved):
LGTM

PR #577, opened 2026-06-02
RPC Review

illuzen:
- get basic peer info now async
- zktree rpc marked as blocking

### DIFF
--- a/node/src/rpc.rs
+++ b/node/src/rpc.rs
@@ -7,7 +7,7 @@
 
 use std::sync::Arc;
 
-use jsonrpsee::{core::RpcResult, proc_macros::rpc, RpcModule};
+use jsonrpsee::{core::{async_trait, RpcResult}, proc_macros::rpc, RpcModule};
 use quantus_runtime::{opaque::Block, AccountId, Balance, Nonce};
 use sc_network::service::traits::NetworkService;
 use sc_transaction_pool_api::TransactionPool;
@@ -36,7 +36,7 @@ pub struct PeerInfo {
 pub trait PeerApi {
 	/// Get basic peer information
 	#[method(name = "peer_getBasicInfo")]
-	fn get_basic_info(&self) -> RpcResult<PeerInfo>;
+	async fn get_basic_info(&self) -> RpcResult<PeerInfo>;
 }
 
 /// Peer RPC implementation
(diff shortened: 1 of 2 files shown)


n13 (approved):
Reviewed [PR #577 "RPC Review"](https://github.com/Quantus-Network/chain/pull/577) by illuzen in a throwaway worktree (now cleaned up). It's a small, focused fix (+17/-13, 2 files) for [issue #513](https://github.com/Quantus-Network/chain/issues/513): avoid blocking the async executor in RPC handlers.

## Verdict: LGTM — approve

Both changes are the textbook-correct treatments, consistent with existing patterns in the crate and Substrate conventions. CI is fully green and it's `MERGEABLE`, only blocked on a required review.

## Change 1 — `node/src/rpc.rs`: `peer_getBasicInfo` made async

```61:69:node/src/rpc.rs
	async fn get_basic_info(&self) -> RpcResult<PeerInfo> {
		if let Some(network) = &self.network {
			let network_state = network.network_state().await.map_err(|_| {
```

Previously this used `futures::executor::block_on(network.network_state())`, which blocks a tokio worker thread until the network worker replies — exactly the executor-blocking anti-pattern #513 targets. Awaiting instead yields the thread. Verified correct because:
- `network_state(&self) -> Result<NetworkState, ()>` is genuinely `async fn` in the local fork (`client/network/src/service/traits.rs:406`), so `.await` is right.
- The `#[async_trait]` + `use jsonrpsee::core::async_trait` pattern is identical to the already-working `TxWatchApiServer` impl in the same crate.
- No internal Rust callers exist; only `scripts/query_peers.sh` calls it over JSON-RPC, which is unaffected by the Rust signature change.

## Change 2 — `node/src/zktree_rpc.rs`: two methods marked `blocking`

`zkTree_getState` and `zkTree_getMerkleProof` are synchronous `fn`s that call into the runtime API (`runtime_api().get_root` / `get_leaf_count` / `get_depth` / `get_merkle_proof`) — synchronous, CPU-bound WASM execution. Adding `blocking` dispatches them to jsonrpsee's dedicated blocking thread pool instead of running on the async executor. This is the standard Substrate convention for runtime-API-backed RPCs.

## Why the two different approaches are both correct

This shows good judgment rather than a blind find-replace:
- `network_state()` is I/O-ish with an async API → `.await` so the thread yields.
- The zktree calls are synchronous CPU-bound with no async API → `blocking` to move them off the executor.

## Verification
- CI all green: Format, Build & Test (macOS + Ubuntu), Clippy & Doc — so it compiles, lints clean, and tests pass on both platforms.

## Minor, non-blocking notes
- `get_basic_info`'s `.map_err(|_| …)` discards the error, but the source error type is `()`, so there's genuinely nothing to log — fine as-is (the zktree handlers already include `{:?}` of the error). 
- Coverage is complete: the other registered RPCs (`System`, `TransactionPayment`) already handle blocking via Substrate, and `TxWatch` is already async/subscription-based — no other handler is left with the `block_on` / sync-runtime-call pattern.
- The third commit is just `fmt`; could be squash-merged, but trivial.

PR #569, opened 2026-05-26
7 leaf proofs

illuzen:
this also needs to update to latest qp-zk-circuits after release

---

> [!NOTE]
> **Medium Risk**
> Changes compile-time wormhole circuit parameters and proof batch size, which affects on-chain aggregated proof verification compatibility until fixtures and downstream qp-zk-circuits versions are aligned.
> 
> **Overview**
> This PR lowers the default wormhole aggregation batch from **16 to 7 leaf proofs** when `build.rs` generates embedded ZK circuit binaries (`QP_NUM_LEAF_PROOFS` still overridable via env). The comment documents the tradeoff: 7 leaves stay at `degree_bits=15` (~1.5 GB peak) for mobile; 8+ needs a larger circuit (~2.5 GB).
> 
> Workspace and `pallet-wormhole` **dev-dependencies** add `qp-wormhole-aggregator` and `qp-wormhole-inputs` (plus circuit-builder in tests) to support an **ignored** `regenerate_test_fixture` test that rebuilds circuits at 7 leaves, proves/aggregates a sample transfer, and rewrites `test-data/aggregated.hex`. A local `compute_zk_leaf_hash` helper keeps fixture generation aligned with the circuit.
> 
> **Note for reviewers:** after this lands, run the ignored fixture generator (or refresh `aggregated.hex` manually) so tests/benchmarks that `include_str!` the hex proof match the new binaries; the PR description also calls for bumping to the latest `qp-zk-circuits` release when available.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8110b33d1c4a526a233192a6347d398c35b9f96d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #578, opened 2026-06-06
Update zk-circuits, poseidon, plonky versions

illuzen:


n13:
## Review: PR #578 — "Update zk-circuits, poseidon, plonky versions"

Reviewed in a throwaway worktree (`/tmp/pr-578-review` on `pr-578-review`); your `main` checkout is untouched.

### Verdict: Request changes — CI is red and the PR is not mergeable as-is

### What the PR does
A dependency bump of the ZK stack plus the minimal code adaptation to compile against the new poseidon API:

- `Cargo.toml`: `qp-plonky2` 1.4.1→1.5.1, `qp-poseidon-core` 2.0.2→2.1.0, `qp-wormhole-*` 2.0.1→**3.0.0**, `qp-zk-circuits-common` 2.0.1→**3.0.0**.
- `primitives/header/`: drops the direct `p3-field`/`p3-goldilocks` deps and instead pulls `Goldilocks` from `qp_poseidon_core`; swaps `Goldilocks::from_int(..)`→`Goldilocks::new(..)` and `bytes_to_digest::<Goldilocks>(..)`→`bytes_to_digest(..)`.

### Blocking issue: stale test fixture → 7 failing tests
CI status: Format, Clippy & Doc pass; **Build & Test Matrix fails on both ubuntu and macOS**. The `compile` step succeeds — the failure is in the `test` step (ubuntu was just fail-fast-cancelled after macOS failed).

7 tests in `pallet-wormhole`'s `aggregated_proof_tests` fail, all rooted at `tests.rs:306`:

```302:307:pallets/wormhole/src/tests.rs
	fn deserialize_test_proof() -> ProofWithPublicInputs<F, C, D> {
		let proof_bytes = get_test_proof_bytes();
		let verifier = crate::get_aggregated_verifier().expect("Verifier should be available");
		ProofWithPublicInputs::<F, C, D>::from_bytes(proof_bytes, &verifier.circuit_data.common)
			.expect("Proof should deserialize")
	}
```

Root cause: the fixture `pallets/wormhole/test-data/aggregated.hex` is a real aggregated proof that was last regenerated in #569 ("7 leaf proofs") and is **not touched by this PR**. The `qp-wormhole-*` major bump (2.x→3.0.0) changed the proof/circuit format, so the old proof no longer deserializes against the new verifier. This is corroborated by the macOS log: the pallet returns `AggregatedProofDeserializationFailed` where `test_..._block_not_found` expects `BlockNotFound` (`tests.rs:375`). Tellingly, the only two `verify_aggregated_proof` tests that *pass* (`wrong_origin`, `invalid_bytes`) are precisely the ones that never need to deserialize the real proof.

### Fix
Regenerate and commit the fixture using the helper the repo already ships for exactly this case:

```bash
cargo test -p pallet-wormhole --lib -- regenerate_test_fixture --nocapture --ignored
```

It writes a fresh proof to `pallets/wormhole/test-data/aggregated.hex` (`tests.rs:654`). Commit that file. (This matches the historical pattern — the fixture was previously refreshed in #509 and #569, both version bumps.)

### Secondary notes
- The `primitives/header` changes look correct and are compile-verified by CI. `Goldilocks::new(..)` vs `from_int(..)` is semantically equivalent here since the input (`number.as_u32() as u64`) is always `< 2^32`, well under the Goldilocks modulus, so no reduction difference applies.
- The lockfile still carries `qp-poseidon-core 1.4.0` and `qp-poseidon-constants` (with `p3-*`) transitively via other crates — this multi-version situation is pre-existing, not introduced here, so not a blocker.

### One question worth confirming before merge
`qp-poseidon-core` 2.0.2→2.1.0 (minor) suggests hash outputs are unchanged, but poseidon is used in consensus-critical paths (`primitives/header`, `pallets/zk-tree/src/tree.rs`, `qpow`). Can you confirm the bump does **not** alter on-chain hash outputs / header hashing, so this stays a clean runtime upgrade and doesn't fork state? The wormhole verifier key changing is expected (that's runtime-internal), but a poseidon output change would be consensus-breaking.

Want me to regenerate the fixture in the worktree and verify the full `pallet-wormhole` suite goes green? I won't push or commit anything without your go-ahead.

n13:
## Updated Review: PR #578 — after fixture regeneration

The blocking issue from my first review is resolved. Here's the updated picture, verified against the current state of the PR branch.

### What changed since last review
- New commit `482b556a "update test"` regenerates the fixture, and it's **already pushed** (`refs/pull/578/head` → `482b556a`), so the PR now has 3 commits.
- The commit touches **only** `pallets/wormhole/test-data/aggregated.hex` (1 line, the hex blob) — no stray code changes snuck in.
- The fixture was regenerated against the **correct new circuit versions** (`qp-wormhole-circuit` 3.0.0, `qp-plonky2` 1.5.1, `qp-poseidon-core` 2.1.0), confirmed via `Cargo.toml`/`Cargo.lock` on the branch. The proof shrank 362,546 → 302,200 hex chars, consistent with the new format.

### Test results — green
Your full local run passes, including all 7 previously-failing tests:

```
test result: ok. 20 passed; 0 failed; 1 ignored ...   (pallet_wormhole)
```

`test_proof_deserialization_succeeds`, `test_parse_aggregated_public_inputs_succeeds`, `..._block_not_found`, `..._nullifier_already_used`, `..._wrong_block_hash`, `..._succeeds_with_valid_state`, and `..._cannot_replay` all pass now.

### Consensus-safety concern — resolved
My earlier open question was whether the `qp-poseidon-core` 2.0.2→2.1.0 bump alters on-chain hash output. It does **not**, and there's a direct guard proving it:

```405:413:primitives/header/src/lib.rs
		// Known good hash for this header - ensures hash algorithm stability
		let expected_hash = "b7dbfd398bcef7d1c91821eb105c52e87dab78e822779bfd0dc7ab132d659a55";
		let actual_hash: [u8; 32] = header.hash().into();

		assert_eq!(
			hex::encode(actual_hash),
			expected_hash,
			"Header hash changed! This would break consensus."
		);
```

`qp_header::poseidon_header_hash_is_stable` asserts a hardcoded header hash and **passed**, so poseidon header hashing output is byte-identical across the bump. This is a clean, hash-stable runtime upgrade — the only changing key material is the wormhole verifier (runtime-internal, expected).

### Remaining (non-blocking) notes
- `Cargo.lock` still carries `qp-poseidon-core 1.4.0` + `qp-poseidon-constants` (with `p3-*`) transitively alongside 2.1.0. Pre-existing multi-version situation, not introduced here.
- GitHub CI is currently re-running on `482b556a` (`config` passed, Format pending, Build & Test Matrix queued). Worth letting it finish to confirm remote green, but local evidence is conclusive.

### Verdict: Approve (pending CI confirmation)
The dependency bump is correct, the code adaptation in `primitives/header` is sound and compile-verified, the fixture is properly regenerated against the new circuits, hashing is proven stable, and the full suite is green. I'd just wait for the GitHub Build & Test Matrix to come back green before merging.

Want me to keep an eye on the re-running CI and report when the Build & Test Matrix finishes? Also, FYI: your main checkout is currently sitting on `illuzen/zk-versions` (not `main`), and there's a now-stale review worktree at `/tmp/pr-578-review` on the old commit — I can clean that up if you'd like.

n13 (approved):
Fixed the test

Now it's good to go

I am still investigating the old versions in the lock file but that can be an additional PR

PR #579, opened 2026-06-07
feat: regenerate weights

dewabisma:
- create regenerate weight script
- fix failing benchmarking
- regenerate weights

---

> [!NOTE]
> **Medium Risk**
> Updated extrinsic weights (especially wormhole aggregated proof and multisig claim_deposits) directly affect block limits and fees; reviewers should confirm the new wormhole weight still covers full dispatch storage work.
> 
> **Overview**
> Adds **`scripts/regenerate_weights.sh`** to benchmark six pallets against the release runtime WASM (`--genesis-builder=runtime`, compiled WASM, shared template) and run `cargo +nightly fmt`.
> 
> **Benchmark / genesis fixes:** `pallet_scheduler`’s `service_timestamp_agendas_base` verify now expects **bucket-normalized** `LastProcessedTimestamp` (aligned with runtime behavior). Under `runtime-benchmarks`, dev genesis strips the chain-spec-only `tech_collective_seed_members` JSON key before deserializing `RuntimeGenesisConfig`.
> 
> **Regenerated weights** reflect the current runtime: transfer paths that used **`Wormhole::TransferProof`** now account for **`ZkTree`** storage (mining rewards, reversible `execute_transfer`); reversible **`schedule_asset_transfer`** drops hand-tuned estimates for measured weights including **`AssetsHolder::BalancesOnHold`**; multisig **`execute`** includes a high-security read, **`claim_deposits`** component ranges/DB writes updated; treasury **`set_treasury_account`** includes a read; wormhole **`verify_aggregated_proof`** is re-benchmarked without the previous manual read/write breakdown (ZK-only ~16s vs prior ~45s with explicit DB weights).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7fa37070bbd5a9cc6467dbb317cc868861357496. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13:
Review complete. I fetched PR #579 into a throwaway worktree, analyzed the diff against `main`, and traced each change back to the actual extrinsic implementations and benchmark definitions.

## Review: PR #579 — `feat: regenerate weights`

### Verdict
The script, scheduler, genesis, and most regenerated weights are sound and actually *improve* accounting. But there is **one real under-weighting regression** (wormhole) and **one confirmed benchmark bug exposed** (multisig). Both Bugbot flags are valid; the wormhole one is the blocker.

---

### 🔴 1. `wormhole::verify_aggregated_proof` drops all DB weight (regression — should block)

On `main`, this weight was hand-augmented because the benchmark *only* measures ZK verification:

```68:73:pallets/wormhole/src/weights.rs
		Weight::from_parts(45_155_000_000, 200000)
			.saturating_add(T::DbWeight::get().reads(100_u64))
			.saturating_add(T::DbWeight::get().writes(130_u64))
```

The regeneration overwrote that with the raw benchmark output — **0 reads, 0 writes, proof size 0**:

```70:72:pallets/wormhole/src/weights.rs
		// Minimum execution time: 15_336_000_000 picoseconds.
		Weight::from_parts(15_976_000_000, 0)
```

The benchmark deliberately wraps only `verifier.verify(...)` in its `#[block]` and even says so: *"This only benchmarks the ZK verification itself, not the full extrinsic (which includes state writes that depend on proof contents)."* But the extrinsic, on the success path, declares `verify_aggregated_proof()` and returns `actual_weight: None`, so the **declared weight is the only charge**. Meanwhile the dispatch does substantial worst-case storage work (up to ~32 `UsedNullifiers` writes, ~32 account mints, ~32 `record_transfer` → ZkTree `Leaves`/`Root`/`Depth`/`LeafCount` + `TransferCount`, total-issuance r/w, miner-fee r/w). All of it is now uncharged in both ref_time and PoV.

The dropped DB weight (~100 reads + 130 writes ≈ 15ms RocksDB) was roughly equal to the ZK time itself, so the extrinsic is under-weighted by ~50% — a block-fill / DoS vector. **Fix:** re-add the manual `DbWeight` reads/writes on top of the regenerated ZK time (this is unavoidably a hand-augmented weight given the benchmark design).

### 🟠 2. `multisig::propose_high_security` no longer charges call length (benchmark bug)

The regenerated signature dropped the `c` term:

```fn propose_high_security(_c: u32, ) -> Weight```

while `propose` keeps `+ 333 * c`. Root cause is in the (unchanged) benchmark — the component is bound and then thrown away:

```225:240:pallets/multisig/src/benchmarking.rs
	fn propose_high_security(
		c: Linear<0, { T::MaxCallSize::get().saturating_sub(100) }>,
	) -> Result<(), BenchmarkError> {
		let _ = c;
		...
		let whitelisted_call =
			pallet_reversible_transfers::Call::<T>::cancel { tx_id: Default::default() };
```

`propose` builds a size-`c` call (`remark { remark: vec![99u8; c as usize] }`), so its regression finds the per-byte cost; `propose_high_security` uses a fixed tiny call, so the regression finds nothing. The HS path still stores/decodes the full bounded call at runtime, so it's under-weighted. This is lower severity (bounded by `MaxCallSize`, ~µs scale) but it's a correctness bug, not noise. **Fix:** mirror `propose` — pad the stored call to size `c` and remove `let _ = c;`. This pre-exists on `main`'s benchmark but this PR is what bakes it into the committed weights.

---

### ✅ Changes that are correct / good
- **Scheduler benchmark fix** is correct. `normalize()` returns the *next* bucket start (`(t/p)*p + p`), so the runtime sets `LastProcessedTimestamp = 3*bucket` for `current_time = 2*bucket`; the old assertion expecting raw `current_time` was genuinely wrong. New assertion derives the expected value from `normalize()` itself — robust.
- **Genesis fix** is correctly gated under `#[cfg(feature = "runtime-benchmarks")]` and strips the chain-spec-only `tech_collective_seed_members` key before deserializing `RuntimeGenesisConfig`, matching the existing `prepare_genesis_build_input` pattern.
- **mining-rewards / reversible-transfers / treasury** weights all *add* previously-missing storage accounting (ZkTree `Depth`/`LeafCount`/`Leaves`/`Root`, `AssetsHolder::BalancesOnHold`, treasury read), replacing stale `Wormhole::TransferProof` and hand-tuned `// 6 base + 1` estimates with measured values. These are improvements.
- **`claim_deposits`** writes now scale with `r` (removed count) instead of `i` (total), consistent with the benchmark's restricted ranges (`r <= i`). Informational, looks intentional.

### Nits
- `scripts/regenerate_weights.sh` uses `pallet-wormhole` (hyphen) but underscores for the other five. It worked, but the inconsistency is worth normalizing.
- The genesis `.remove(...)` duplicates the removal already in `prepare_genesis_build_input`; trivial one-liner, acceptable but could share a helper.

I did **not** post anything to GitHub (per your rules) and removed the throwaway worktree. Want me to draft the two weight/benchmark fixes (re-add wormhole DB weight + fix the `propose_high_security` benchmark)?

n13 (commented):
## Finding
Low: `claim_deposits` is still called outside the generated benchmark ranges. The generated weight declares `i: [100, 200]` and `r: [0, 100]`, but pre-dispatch still charges with `r = MaxTotalProposalsInStorage`, which is `200` in runtime config, and post-dispatch can pass `i < 100` when there are fewer proposals.

This likely overcharges rather than undercharges, but it relies on extrapolated benchmark coefficients. I’d align the benchmark domain with the actual call-site domain, or adjust the call-site to the proven max cleaned-per-signer invariant.

## Checks
Reviewed in a detached throwaway worktree. `git diff --check` passed, and `SKIP_WASM_BUILD=1 cargo check -p quantus-runtime --features runtime-benchmarks` passed.

GitHub CI still had build/test and clippy/doc pending when checked; format and config were passing.

illuzen (approved):

PR #581, opened 2026-06-09
Update rusty crystals

illuzen:
upgraded some other packages to reduce (not eliminate) various cargo audit issues

---

> [!NOTE]
> **Medium Risk**
> Touches P2P DNS resolution and the miner QUIC/TLS path; dependency major/minor bumps could affect connectivity or certificate handling at runtime.
> 
> **Overview**
> Bumps **Quantus rusty-crystals** crates (`qp-rusty-crystals-hdwallet` 2.4.0, dilithium 2.5.0 in the lockfile) and aligns related crypto/DNS deps (`qp-poseidon-core` 2.1.0 only, `x509-parser` 0.18.1, `hickory-resolver` 0.26.1).
> 
> **litep2p** is updated for Hickory 0.26: resolver construction drops `TokioConnectionProvider`, treats `TokioResolver::build()` as fallible, and broadens `CannotReadSystemDnsConfig` to a boxed error. Tests mirror the same builder pattern.
> 
> The **node miner QUIC server** moves to **quinn 0.11** with **rustls 0.23** and **rcgen 0.13**—self-signed TLS material uses `rustls::pki_types`, and the endpoint is created via `QuicServerConfig::try_from`. Workspace **quinn** is pinned at 0.11.9 and older quinn/rustls 0.21 stacks are removed from the lockfile to reduce `cargo audit` noise.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c82dcb3536eca16b1d211bf336dc0866a4310ea. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #586, opened 2026-06-10
fix: use system DNS resolver in litep2p backend

n13:
hickory-resolver 0.26 (bumped in #581) changed `ResolverConfig::default()` to an empty config with no nameservers, so all `/dns/...` bootnode and telemetry dials fail silently with `NoConnections` (0 peers). Use the system DNS config instead; startup now fails loudly if it can't be read.

### DIFF
--- a/client/network/src/litep2p/mod.rs
+++ b/client/network/src/litep2p/mod.rs
@@ -324,6 +324,7 @@ impl Litep2pNetworkBackend {
 			.unzip();
 
 		config_builder
+			.with_system_resolver()
 			.with_websocket(WebSocketTransportConfig {
 				listen_addresses: websocket.into_iter().flatten().map(Into::into).collect(),
 				yamux_config: litep2p::yamux::Config::default(),


n13:
## Review

Verified the root cause against both resolver versions in `Cargo.lock`:

- **hickory-resolver 0.24.4**: `impl Default for ResolverConfig` explicitly returned Google's nameservers (`8.8.8.8`, `8.8.4.4`, …).
- **hickory-resolver 0.26.1**: `ResolverConfig` is `#[derive(Default)]`, so `Default::default()` has an **empty** `name_servers` list.

The vendored litep2p builds its resolver from `Default::default()` unless `use_system_dns_config` is set, so after #581 every `/dns/...` lookup had zero nameservers to query. Enabling `with_system_resolver()` here is the right fix.

Correctness checks:

- `Litep2p::new` calls `hickory_resolver::system_conf::read_system_conf()` and propagates failure as `Error::CannotReadSystemDnsConfig` — startup fails loudly as described.
- The resolver is built once and shared by both the TCP and WebSocket transports, so both dial paths are covered.
- hickory's `system-config` feature is in its default feature set (unix resolv.conf, windows ipconfig).
- This is the only production `Litep2pConfigBuilder` call site; the rest are tests using IP listen addresses.

Notes:

1. **Residual footgun**: the default path in the vendored litep2p still silently builds a resolver with zero nameservers; any future caller that forgets `.with_system_resolver()` reinherits this failure. Filed #587 to make system DNS config unconditional in the vendored crate.
2. **Telemetry claim in the description is likely inaccurate**: telemetry goes through `sc-telemetry` → `libp2p-dns`, which pins hickory-resolver 0.24.4 and never touches litep2p's resolver, so telemetry dials are unaffected by this change.
3. **Ops/release-notes note**: nodes in minimal containers without `/etc/resolv.conf` previously worked via the implicit Google DNS fallback (pre-#581) but will now fail at startup with `CannotReadSystemDnsConfig` even when no DNS multiaddrs are configured. Intended fail-loud trade-off, worth a mention in the v0.7.0 release notes.

LGTM otherwise.

n13:
GPT review: 
No issues found in PR #586.

The one-line change in `client/network/src/litep2p/mod.rs` correctly enables `with_system_resolver()`. In `client/litep2p/src/lib.rs`, that flag makes startup read the system DNS config and propagate `CannotReadSystemDnsConfig` on read/build failure, so the PR matches its “fail loudly” intent.

Verified with:

`cargo check -p sc-network`

Residual risk: I did not run an end-to-end node boot/dial test with `/dns/...` bootnodes, so runtime DNS behavior is still worth smoke-testing in the target deployment environment. I did not post anything to GitHub.

PR #590, opened 2026-06-12
Governance settings update

n13:
Almost — one correction: **support is 60%, not 61%**. That's deliberate: support is measured against all 5 members, so 3/5 = exactly 60%. If you set support to 61%, 3 ayes would fail quorum and you'd need 4/5 to execute. Approval (ayes vs nays among voters) is the one at 61%, which is what makes 2 nays block (3a/2n = 60% < 61%).

Also "1 day minimum confirmation" is the right phrasing — it's 24h of *continuous* passing; if nays abort confirmation, the clock restarts.

Here's a PR description:

---

**Title:** `runtime: tighten tech collective governance track`

## Summary

Tightens the Tech Collective referenda track for a 5-member collective:

| Parameter | Before | After |
|---|---|---|
| `min_support` (quorum) | 0% | **60%** (constant) |
| `min_approval` | 100% → 50% (decaying) | **61%** (constant) |
| `confirm_period` | 20 blocks (~4 min) | **1 day** |
| `min_enactment_period` | 20 blocks (~4 min) | **1 day** |

`decision_period` (1 day) and `prepare_period` (20 blocks) are unchanged. The `fast-governance` test feature still collapses all timings to 2 blocks.

## Security properties (5-member collective)

- **3/5 ayes required to pass** — support quorum of 60% closes the hole where 2 ayes could approve a referendum
- **2 nays always block** — 3 ayes / 2 nays = 60% approval, below the 61% threshold
- **1 nay cannot block** — 3 ayes / 1 nay = 75% approval, passes
- **Compromising 3 members is not enough** — against 2 honest nays, an attacker needs 4 aye votes to clear 61% approval
- **Dissent always has a 24h window** — approval cannot conclude before 24h of continuous passing; any nay that drops approval below threshold aborts confirmation and restarts the clock
- **24h enactment delay** — approved proposals execute no earlier than 1 day after approval

## Notes

- Thresholds are constant over the decision period (no decay)
- Support is computed against the live member count: with 5 members, 60% = 3 ayes
- Takes effect via runtime upgrade in the next release

---

### DIFF
--- a/docs/TECH_COLLECTIVE_GOVERNANCE_TUNING.md
+++ b/docs/TECH_COLLECTIVE_GOVERNANCE_TUNING.md
@@ -6,27 +6,27 @@ Scope: the runtime-upgrade track (`TechReferenda` = `pallet_referenda::Pallet<Ru
 
 All periods are in blocks. `runtime/src/lib.rs:84-89`: `TARGET_BLOCK_TIME_MS = 12_000`, so `MINUTES = 5`, `HOURS = 300`, `DAYS = 7200` blocks.
 
-Track definition: `runtime/src/governance/definitions.rs:165-188` (`TechCollectiveTracksInfo::create_tech_collective_tracks`):
+Track definition: `runtime/src/governance/definitions.rs:165-192` (`TechCollectiveTracksInfo::create_tech_collective_tracks`):
 
 ```rust
 max_deciding: 1,
 decision_deposit: 1000 * UNIT,
 prepare_period: 20,
 decision_period: DAYS,
-confirm_period: 20,
-min_enactment_period: 20,
+confirm_period: DAYS,
+min_enactment_period: DAYS,
 ```
 
 | Parameter | Current (blocks) | Wall clock | Meaning |
 |---|---|---|---|
 | `prepare_period` | 20 | 4 min | Delay between submission and decision start |
 | `decision_period` | 7200 (`DAYS`) | 24 h | Window in which the referendum must reach passing state |
-| `confirm_period` | 20 | 4 min | Must remain continuously passing this long to be approved |
-| `min_enactment_period` | 20 | 4 min | Min delay between approval and dispatch of the upgrade |
+| `confirm_period` | 7200 (`DAYS`) | 24 h | Must remain continuously passing this long to be approved |
+| `min_enactment_period` | 7200 (`DAYS`) | 24 h | Min delay between approval and dispatch of the upgrade |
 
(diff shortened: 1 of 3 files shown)


n13:
## Review

**Verdict: Changes requested** — the parameter changes themselves are correct and well-documented, but CI is red and the PR description needs cleanup. (Evaluation only, not a formal review request.)

### Blocking

1. **Format check fails, so the Build & Test jobs were skipped** — the "fix governance tests" commit was never actually verified by CI. rustfmt wants the comment at `runtime/tests/governance/tech_collective.rs:686` reflowed:

   ```
   -  // 2. A referendum is approved when 3 of 5 members vote AYE and no one votes against
   -  //    (meets the 60% support and 61% approval thresholds).
   +  // 2. A referendum is approved when 3 of 5 members vote AYE and no one votes against (meets
   +  //    the 60% support and 61% approval thresholds).
   ```

   `cargo +nightly fmt --all` and push.

2. **PR description contains leftover conversation text** ("Almost — one correction: …", "Here's a PR description:"). Please replace the body with just the intended description, and consider the title it suggests (`runtime: tighten tech collective governance track`) over the current generic one.

### Verified correct

- **Threshold math checks out against the pallet sources.** `passing` is inclusive (`y >= threshold`, pallet-referenda `types.rs:637`), and `Perbill::from_percent(61)` = 610,000,000 vs `from_rational(3,5)` = 600,000,000, so 3 ayes / 2 nays (60% approval) fails while 3/5 support (exactly 60%) passes. With all members at rank 0 (`VoteWeight = Linear`, `PromoteOrigin = Never`), weighted ayes equal bare ayes. All claimed properties hold: 3/5 ayes pass, 2 nays always block, 1 nay cannot block, and overriding 2 honest nays requires 4 ayes.
- **Tests pass locally on this branch** (since CI never ran them): full governance suite, 33 passed / 0 failed / 1 pre-existing ignored.
- **The new doc is accurate.** Spot-checked timing constants, the `UndecidingTimeout = 45 * DAYS` value (including the stale "90 days" comment it calls out), Cancel/Kill origins, and tally semantics against the runtime and crate sources.

### Follow-up worth its own issue

The doc's §4 finding is important: `RemoveOrigin`/`AddOrigin` = `RootOrMemberForCollectiveOrigin` means **any single member can unilaterally remove the other four** (or stuff membership up to 13), which invalidates every threshold guarantee this PR establishes. That should be tracked and fixed before these parameters are relied on for mainnet security.

illuzen (approved):

PR #593, opened 2026-06-12
Simplify tx watch ready-queue lookup and log decode failures

n13:
## Summary
- Remove the fallback scan over the ready queue in `TxWatch`; rely solely on `pool.ready_transaction()` for lookup
- Add warn-level logging when stripping the length prefix or decoding a ready-queue transaction fails, instead of silently skipping

### DIFF
--- a/node/src/txwatch.rs
+++ b/node/src/txwatch.rs
@@ -88,24 +88,18 @@ impl TxWatch {
 					continue;
 				}
 
-				let encoded = if let Some(in_pool) = pool.ready_transaction(&tx_hash) {
-					Encode::encode(in_pool.data())
-				} else {
-					let found = pool
-						.ready()
-						.find(|in_pool| *in_pool.hash() == tx_hash)
-						.map(|in_pool| Encode::encode(in_pool.data()));
-					let Some(data) = found else {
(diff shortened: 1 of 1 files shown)


n13:
## Review

**Verdict: LGTM** (evaluation only, not a formal approval)

### What I checked

- **The removed fallback was dead code.** The node pins `TransactionPoolType::SingleState` (`node/src/service.rs:625`). In the single-state pool, `ready_transaction()` is a direct hash lookup into the validated pool's ready set (`validated_pool().ready_by_hash`), and `ready()` iterates that same set — so the linear scan could never find a transaction the direct lookup missed. Even under the fork-aware pool, both methods read the same most-recent view (`view_store.ready()` / `view_store.ready_transaction()`), so the conclusion holds regardless of pool type.
- **Warn-level logs on decode failure** replace silent `continue`s — decode failures here would indicate a real bug (a ready-pool extrinsic that doesn't decode as `UncheckedExtrinsic`), so surfacing them is right. The not-in-ready-queue case correctly stays at trace, since that's an expected race with block inclusion.
- **Local verification:** branch builds and all 17 txwatch tests pass.

### Nit (non-blocking)

- Commit message is "minor tx watch update" — the PR title is the better description; worth squash-merging with the title.
- Build & Test Matrix was still running at review time; format and clippy are green.

illuzen (approved):

PR #594, opened 2026-06-12
fix: report connected peers in litep2p network_state

### LINKED ISSUE #592: peer_getBasicInfo RPC always returns peer_count: 0 since q-day-2 (#577 async refactor)
## Summary

Since v0.7.1-q-day-2, the custom `peer_getBasicInfo` RPC always reports `peer_count: 0` and an empty `connected_peers` list, even when the node is connected and syncing. Likely a regression from the basic-peer-info async refactor in #577.

## Reproduction

Node `0.7.1-q-day-2-5ea94cb1e53` (release binary, macOS aarch64) on planck, actively syncing with peers (informant shows them, `system_health` agrees). Both methods queried on the same node within the same second:

```
$ curl -s 127.0.0.1:9933 -H 'Content-Type: application/json' \
    -d '{"jsonrpc":"2.0","id":1,"method":"peer_getBasicInfo"}'
{"jsonrpc":"2.0","id":1,"result":{"peer_id":"QmYU89…","peer_count":0,"connected_peers":[],
 "external_addresses":["/ip4/…/tcp/56627/p2p/QmYU89…"],"listen_addresses":[…]}}

$ curl -s 127.0.0.1:9933 -H 'Content-Type: application/json' \
    -d '{"jsonrpc":"2.0","id":1,"method":"system_health"}'
{"jsonrpc":"2.0","id":1,"result":{"peers":10,"isSyncing":true,"shouldHavePeers":true}}
```

`peer_id`, `external_addresses`, and `listen_addresses` in the response are correct — only `peer_count` / `connected_peers` are wrong.

## Root cause

`peer_getBasicInfo` counts `network_state().connected_peers` (`node/src/rpc.rs`), but the litep2p backend's `network_state()` hardcodes it to an empty map:

```rust
// client/network/src/litep2p/service.rs
async fn network_state(&self) -> Result<NetworkState, ()> {
    Ok(NetworkState {
        peer_id: self.local_peer_id.to_base58(),
        listened_addresses: ...,   // populated correctly
        external_addresses: ...,   // populated correctly
        connected_peers: HashMap::new(),
        not_connected_peers: HashMap::new(),
        // TODO: Check what info we can include here.
        //       Issue reference: https://github.com/paritytech/substrate/issues/14160.
        peerset: serde_json::json!("Unimplemented. ..."),
    })
}
```

This is an inherited upstream TODO (paritytech/substrate#14160): the libp2p backend populated `connected_peers`, litep2p never did. The regression surfaced in q-day-2 because #574 made litep2p the only backend. Note this is unrelated to `--enable-peer-sharing` — with the flag off the RPC errors with "Peer sharing is not enabled"; with it on, it succeeds but counts the hardcoded empty map.

Fix would be to populate `connected_peers` in `litep2p/service.rs::network_state()` (e.g. from the peerset/transport manager), or have `peer_getBasicInfo` use the same source as `system_health`.

## Impact

The GUI miner used `peer_getBasicInfo` for its peer-count display, so it showed 0 peers while the node was healthily syncing. We've switched the app to `system_health` as a workaround (Quantus-Network/quantus-apps#522 follow-up), but the custom RPC should either report real values or be removed.

n13:
## Summary

Fixes #592 — `peer_getBasicInfo` always returned `peer_count: 0` and an empty `connected_peers` list since v0.7.1-q-day-2, because the litep2p backend's `network_state()` hardcoded `connected_peers` to an empty map (an inherited upstream TODO, surfaced when litep2p became the only backend).

- Add a `PeersetCommand::GetConnectedPeers` command that returns the peers currently in `PeerState::Connected` along with their substream direction.
- `Litep2pNetworkService::network_state()` now queries the block-announce peerset — the same source `system_health`'s peer count is derived from — and populates `connected_peers` with the result. Connection-level addresses are not exposed by litep2p, so endpoint info only reflects substream direction; `peer_id`, peer count, and peer IDs are now correct.
- This also fixes the standard `system_networkState` RPC, which reads the same struct.

## Test plan

- [x] New unit test `get_connected_peers_returns_only_connected_peers` verifying only peers in `Connected` state are returned
- [x] `cargo test -p sc-network peerset` — all 15 tests pass
- [x] Manual check on a syncing planck node: `peer_getBasicInfo` reports `peer_count: 3` with the connected peer IDs listed, matching `system_health`'s `peers: 3` queried moments apart:

```
$ curl -s 127.0.0.1:9933 -H 'Content-Type: application/json' \
    -d '{"jsonrpc":"2.0","id":1,"method":"peer_getBasicInfo"}'
{"jsonrpc":"2.0","id":1,"result":{"peer_id":"QmbBmA…","peer_count":3,
 "connected_peers":["QmbctL…","QmZT5L…","QmczRB…"], ...}}

$ curl -s 127.0.0.1:9933 -H 'Content-Type: application/json' \
    -d '{"jsonrpc":"2.0","id":1,"method":"system_health"}'
{"jsonrpc":"2.0","id":1,"result":{"peers":3,"isSyncing":true,"shouldHavePeers":true}}
```

### DIFF
--- a/client/network/src/litep2p/service.rs
+++ b/client/network/src/litep2p/service.rs
@@ -24,11 +24,14 @@ use crate::{
 		notification::{config::ProtocolControlHandle, peerset::PeersetCommand},
 		request_response::OutboundRequest,
 	},
-	network_state::NetworkState,
+	network_state::{
+		Endpoint as NetworkStateEndpoint, NetworkState, Peer as NetworkStatePeer,
+		PeerEndpoint as NetworkStatePeerEndpoint,
+	},
 	peer_store::PeerStoreProvider,
 	service::{
 		out_events,
-		traits::{IfDisconnected, OutboundFailure, RequestFailure},
+		traits::{Direction, IfDisconnected, OutboundFailure, RequestFailure},
 	},
 	Event, NetworkDHTProvider, NetworkEventStream, NetworkPeers, NetworkRequest, NetworkSigner,
 	NetworkStateInfo, NetworkStatus, NetworkStatusProvider, ProtocolName, Signature,
@@ -343,6 +346,16 @@ impl NetworkStatusProvider for Litep2pNetworkService {
 	}
 
 	async fn network_state(&self) -> Result<NetworkState, ()> {
+		let Some(handle) = self.peerset_handles.get(&self.block_announce_protocol) else {
+			return Err(())
+		};
+		let (tx, rx) = oneshot::channel();
+		handle
+			.tx
+			.unbounded_send(PeersetCommand::GetConnectedPeers { tx })
+			.map_err(|_| ())?;
+		let connected_peers = rx.await.map_err(|_| ())?;
+
 		Ok(NetworkState {
 			peer_id: self.local_peer_id.to_base58(),
 			listened_addresses: self
@@ -358,7 +371,33 @@ impl NetworkStatusProvider for Litep2pNetworkService {
 				.into_iter()
 				.map(|a| Multiaddr::from(a).into())
 				.collect(),
-			connected_peers: HashMap::new(),
+			connected_peers: connected_peers
+				.into_iter()
+				.map(|(peer, direction)| {
+					// Connection-level addresses are not exposed by litep2p, only the
+					// substream direction is known.
+					let endpoint = match direction {
+						Direction::Inbound => NetworkStatePeerEndpoint::Listening {
+							local_addr: Multiaddr::empty(),
+							send_back_addr: Multiaddr::empty(),
+						},
+						Direction::Outbound => NetworkStatePeerEndpoint::Dialing(
+							Multiaddr::empty(),
+							NetworkStateEndpoint::Dialer,
+						),
+					};
+
+					(
+						peer.to_base58(),
+						NetworkStatePeer {
+							endpoint,
+							version_string: None,
+							latest_ping_time: None,
+							known_addresses: HashSet::new(),
+						},
+					)
+				})
+				.collect(),
 			not_connected_peers: HashMap::new(),
 			// TODO: Check what info we can include here.
 			//       Issue reference: https://github.com/paritytech/substrate/issues/14160.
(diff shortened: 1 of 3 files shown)


n13:
Ran local node, it works

Live verification passed. I ran your freshly built release binary on planck with --enable-peer-sharing (using a temp base path so it wouldn't touch your data), let it connect, and queried both endpoints:

peer_getBasicInfo returned peer_count: 3 with all three connected peer IDs — the exact field that was stuck at 0 before the fix.
system_health returned peers: 3 at the same moment, so the two sources now agree.

n13:
## Review

**Verdict: LGTM** (evaluation only, not a formal approval)

### What I checked

- **Root cause is correctly identified.** `network_state()` in the litep2p backend hardcoded `connected_peers: HashMap::new()`, so `peer_getBasicInfo` (which derives `peer_count` from `connected_peers.len()` in `node/src/rpc.rs`) always reported 0.
- **Implementation follows the existing pattern.** The new `PeersetCommand::GetConnectedPeers` mirrors `GetReservedPeers` exactly, and the `network_state()` query uses the same handle-lookup + oneshot structure as `reserved_peers()` in the same file. No new conventions introduced.
- **Source of truth is consistent with `system_health`.** Filtering on `PeerState::Connected` matches the semantics of the `connected_peers` atomic counter that `sync_num_connected()` reads (incremented on the `Opening → Connected` transition, decremented on close), so the two RPCs now agree by construction. `Closing`-state peers are correctly excluded.
- **The `Direction → traits::Direction` conversion** already existed in `peerset.rs`; no duplication.
- **Local test run:** all 15 `sc-network` peerset tests pass, including the new `get_connected_peers_returns_only_connected_peers`. The test correctly verifies that peers with merely-opening substreams are excluded.

### Notes (non-blocking)

- Behavior change: `network_state()` can now return `Err` (surfaced as RPC error 5001) if the block-announce peerset handle is missing or the peerset loop has shut down, instead of silently returning empty data. This is an improvement (fail loudly) and matches `reserved_peers()`.
- Endpoint addresses are `Multiaddr::empty()` since litep2p doesn't expose connection-level addresses — acknowledged in the PR; only substream direction is meaningful in the `endpoint` field.
- `not_connected_peers` remains empty (pre-existing upstream TODO, out of scope).
- Build & Test Matrix was still running at review time; format and clippy are green.

illuzen (approved):

PR #595, opened 2026-06-13
Inline txpool service

illuzen:
- inline sc-transaction-pool
- inline sc-service
- fix metric bug in sc-transaction-pool

inlining these will help with security audit as well as trimming dependencies

---

> [!NOTE]
> **High Risk**
> Large vendored changes to node startup, networking, RPC, and the transaction pool—core infrastructure where regressions affect sync, submission, and observability; risk is mitigated if this is mostly upstream Substrate parity with minimal local edits.
> 
> **Overview**
> **Vendors upstream Substrate node plumbing** by adding in-repo `client/service` (`sc-service`) and `client/transaction-pool` (`sc-transaction-pool`), registering them as workspace members, and **patching `crates-io`** so the node builds against these paths instead of published crates.
> 
> The new **`sc-service`** crate carries full-node assembly: client/backend construction, **`spawn_tasks`** (RPC v1/v2, telemetry, Prometheus, tx-pool hooks), and **`build_network`** / sync engine wiring. **`chain_ops`** (import/export blocks, raw state export, revert, check block) and the large **`Client`** implementation are included as vendored sources.
> 
> Workspace **`Cargo.toml` / `Cargo.lock`** pick up supporting deps (`directories`, `exit-future`, `linked-hash-map`, `tracing-futures`, extra `sc-*` RPC/network crates, `thiserror` 2 for patched crates). The PR description also notes a **transaction-pool Prometheus metrics fix** in the inlined pool code.
> 
> **Intent:** smaller, auditable dependency surface for security review rather than changing chain consensus rules directly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 936f3ea5da6362a91ca761a421e0820b5549445c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (commented):
## Review: vendored crates vs. upstream

**Verdict: LGTM with caveats.** This is a faithful vendoring of the exact published crates (`sc-transaction-pool` 44.1.0, `sc-service` 0.56.0), and the claimed txpool metrics fix is present and correct. The title undersells the scope though — beyond formatting + the metrics fix, the PR also removes panics/unwraps (which changes failure behavior on "impossible" paths), changes two public `sc-service` function signatures, and adds a few **silent-failure** paths. None are blockers, but a couple deserve a second look before merge.

### How this was verified
Downloaded the exact `.crate` tarballs from crates.io for both versions, then formatted **both** upstream and vendored sources with this repo's `.rustfmt.toml` + the same nightly rustfmt, so the diff shows only real (non-formatting) changes.

- ✅ **Identical source-file sets** — no `src/` files added/removed (only the upstream top-level `tests/` integration dir is dropped).
- ✅ Most files are **byte-identical** to upstream after formatting.
- ✅ Library `[dependencies]` match upstream (one manifest exception, see below).

### The claimed fix — present & correct
`single_state_txpool.rs`: `submitted_transactions` is now incremented only *after* a successful submit and counts only `Ok` results. Matches the description. Note it's applied to the **single-state pool only**, not the fork-aware pool.

### Differences from upstream
1. **Cosmetic** (no behavior change): the bulk of the diff is `return x` → `return x;`, `break;`/`continue;`, a match-arm reformat, and `&[h.clone()]` → `std::slice::from_ref(h)`.
2. **Panic/unwrap removal** (clippy commit): `expect`/`unwrap`/panic → graceful error/skip. Equivalent on the happy path; differs only if a "qed" invariant is violated (upstream panics loudly, here it returns an error or continues). e.g. `fork_aware_txpool` initial-view `TreeRoute` now `return self` on error; `graph/future.rs`/`ready.rs` `continue` instead of `expect`; `client.rs` proof recorder / finalize header; `builder.rs` genesis hash.
3. **Public API signature changes** (`sc-service/src/config.rs`):
   - `PrometheusConfig::new_with_default_registry` → `Option<Self>` (was `Self`)
   - `BasePath::from_project` → `Option<BasePath>` (was `BasePath`)
   These required the accompanying `client/cli` edits.
4. **Manifest deviation:** vendored `sc-service/Cargo.toml` adds `sc-client-db.default-features = true`, whereas upstream relied on the workspace default (`default-features = false`).
5. **⚠️ Silent failures** (worth reconsidering given our "fail early, no silent failures" convention): in `fork_aware_txpool/tx_mem_pool.rs` the sync-bridge helpers no longer panic when the worker channel is broken — `remove_transactions_sync` / `update_transaction_priority_sync` do `let _ = response.recv();`, and `extend_unwatched_sync` does `.unwrap_or_default()` (returns an empty Vec). `try_insert_with_replacement_sync` is the good example — it propagates an error. Suggest at least logging an error before degrading in the other three.

### On the tests
Nearly all upstream unit tests depend on monorepo-internal path crates (`substrate-test-runtime`, `substrate-test-runtime-client`, and for the `tests/` dir, `zombienet`/`txtesttool`) that aren't available standalone — so they can't compile outside polkadot-sdk. The PR gates exactly those behind a `test-helpers` feature that **can't be satisfied** (those crates aren't vendored), which is reasonable. The only test module that survives is `sc-service`'s `task_manager/tests.rs` (self-contained — only `TaskManager` + `futures`/`std`), which still runs under `cargo test`. Net effect: the inlined crates ship with upstream's own test suites effectively disabled, so regressions in the vendored copies won't be caught by their tests.

*Source-level audit only — I did not run a full build/test of the vendored crates.*

n13 (approved):
LGTM but plz review why we have silent failures now that weren't there before - intentional?

PR #598, opened 2026-06-17
Implement turnstile and migration

illuzen:
# Wormhole soundness turnstile

## Summary

Adds a Zcash-style turnstile to the zk-wormhole so the chain can never mint more
value out of the wormhole than could possibly have been deposited into it. The
invariant enforced on every exit is:

```
TotalWormholeExits <= PotentialWormholeBalance
```

If a ZK soundness bug ever let a forged proof through, this backstop caps the
damage at the currently-available margin and rejects the exit instead of minting
unbacked tokens.

## How it works

- **`PotentialWormholeBalance`** — running total of value held by *ambiguous*
  addresses (accounts with `nonce == 0`, i.e. that have never signed a
  transaction). These are the addresses that could be wormhole deposits.
- **`TotalWormholeExits`** — cumulative value minted via wormhole exits.
- **Deposit tracking**: `record_transfer` adds to the pool whenever a native
  transfer lands in an ambiguous address (covers ordinary transfers, genesis
  endowments, mining rewards, and exits that flow back into the wormhole).
- **Reveal**: the moment an account first signs a transaction it is no longer
  ambiguous, so `WormholeProofRecorderExtension::validate()` subtracts its
  balance from the pool.
- **Enforcement**: `verify_aggregated_proof` checks `exits_after <= potential_balance`
  *before* minting any exit output, and rejects with `SoundnessInvariantViolation`
  otherwise.

A shared `is_ambiguous_account()` helper is the single source of truth for the
`nonce == 0` heuristic, used by both the deposit-tracking and reveal paths.

## Seeding an already-running chain

On a chain that predates this tracking, wormhole deposits made earlier aren't
reflected in `PotentialWormholeBalance` (it would default to `0`), so the first
legitimate exit would trip the invariant and brick the wormhole.

A one-shot versioned migration (`migrations::v1::InitSoundnessCounters`, wired as
`MigrateV0ToV1`) seeds:

```
PotentialWormholeBalance = total_issuance()
```

Every balance in an ambiguous address is backed by issued tokens, so total
issuance is a safe upper bound on what could legitimately be exited — seeding to
it can never block a valid exit. The pool then tightens toward the true ambiguous
sum as accounts reveal themselves. (We intentionally avoid an extra configurable
buffer: it could only loosen the bound, never make it safer.) On a fresh chain
the migration is skipped and the pool is seeded by the block-1 genesis-endowment
`record_transfer` calls instead.

## Changes

- `pallets/wormhole/src/lib.rs` — `PotentialWormholeBalance` / `TotalWormholeExits`
  storage, `SoundnessInvariantViolation` error, `STORAGE_VERSION = 1`,
  deposit tracking in `record_transfer`, invariant enforcement in
  `verify_aggregated_proof`, and the `is_ambiguous_account()` helper.
- `pallets/wormhole/src/migrations.rs` — v0 -> v1 seeding migration (with
  try-runtime pre/post checks).
- `runtime/src/lib.rs` — wire `MigrateV0ToV1` into `Executive` and bump
  `spec_version` 131 -> 132.
- `runtime/src/transaction_extensions.rs` — reveal logic in the proof-recorder
  extension's `validate()`, weight accounting, and extensive integration tests.
- `Cargo.toml`s — `try-runtime` feature plumbing.
- `docs/wormhole-soundness-detection-plan.md` — design write-up.

## Testing

`pallet-wormhole` (25 pass) and the runtime `transaction_extensions` suite
(27 pass) are green. Coverage includes:

- Deposits to ambiguous vs. revealed addresses.
- The full sender/recipient nonce matrix through a complete transaction
  lifecycle (validate -> prepare -> dispatch -> post_dispatch), including batch
  transfers (sender revealed once, each ambiguous recipient counted).
- Mining rewards: a mined block's full emission flows into the pool (miner and
  treasury are both ambiguous).
- Migration seeds the pool to total issuance.
- **Soundness enforcement**: an exit is rejected with
  `SoundnessInvariantViolation` when the margin is exhausted (`P == E`), even when
  the exit targets an ambiguous address (exiting into another wormhole is valid,
  but must still respect the margin), plus the cold-start `P == 0` case.

## Notes

- The `nonce == 0` heuristic is a deliberate over-approximation of "wormhole
  deposit address": unused accounts are also counted, which only makes the bound
  looser (safe). Revealing tightens it back down.
- Exiting into another wormhole address (ambiguous) is a supported operation; the
  exit output legitimately re-counts toward the pool because it's still inside the
  wormhole system. Margin is only consumed when value leaves to a revealed account.

---

> [!NOTE]
> **High Risk**
> Changes wormhole exit minting and supply accounting with a runtime migration; a bug in counters or reveal timing could block legitimate exits or weaken the backstop.
> 
> **Overview**
> Adds an on-chain **wormhole soundness turnstile** so cumulative ZK exits cannot exceed value tracked as possibly wormhole-backed. The enforced invariant is `TotalWormholeExits <= PotentialWormholeBalance`.
> 
> **`PotentialWormholeBalance`** grows when native transfers (via `record_transfer`) land on *ambiguous* accounts (`nonce == 0`, via `is_ambiguous_account`). **`TotalWormholeExits`** increases only after successful aggregated exits. **`verify_aggregated_proof`** checks the invariant *before* minting and fails with **`SoundnessInvariantViolation`** otherwise.
> 
> First-time signers *reveal* as non-wormhole: **`WormholeProofRecorderExtension::validate`** subtracts the signer’s free balance from the pool (with updated **weight** accounting). Pallet storage version bumps to **v1**; runtime **`spec_version` 132** wires **`MigrateV0ToV1`**, which one-shot seeds **`PotentialWormholeBalance = total_issuance()`** so pre-tracking deposits are not bricked on live chains.
> 
> Includes design doc, try-runtime feature plumbing, and broad pallet/runtime tests (deposit/reveal matrix, mining rewards, margin-exhausted exits).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5b4e863a7b1774d871faee0d6453f07ae644745b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/wormhole-soundness-detection-plan.md
+++ b/docs/wormhole-soundness-detection-plan.md
@@ -23,6 +23,13 @@ On-chain, wormhole addresses are indistinguishable from regular dilithium addres
 
 When an address signs its first transaction, it "reveals" itself as a regular dilithium address (not a wormhole address), and its balance should be subtracted from the potential wormhole pool.
 
+### Accounts that never reveal
+
+The `nonce == 0` heuristic over-counts two kinds of accounts that have a zero nonce but are known *not* to be wormhole deposits. These are excluded via the `NonWormholeAccounts: Contains<AccountId>` config (so they never add to the pool), and the runtime populates it as follows:
+
+- **Multisig accounts** spend through their signatories, so the multisig account itself never signs and never reveals. The runtime excludes any address registered in `pallet_multisig` (`is_multisig`). Because funds can be sent to a *pre-computed* multisig address before it is created, the multisig pallet also calls `TransferProofRecorder::reveal_address` on creation, which subtracts the address's balance from the pool (the multisig analog of a normal account's first-signature reveal). Together these make a multisig net zero into the pool over its lifetime.
+- **Keyless accounts** that can never sign at all: the treasury account (a keyless governance account that receives a block reward every block), the `PalletId`-derived pallet accounts, and the sentinel minting addresses used as the `from` side of minted transfers.
+
 ## Storage Items
 
 Add to `pallets/wormhole/src/lib.rs`:
@@ -219,7 +226,8 @@ Wormhole exits are unsigned (`ensure_none`). They don't have a signer and won't 
 1. **`pallets/wormhole/src/lib.rs`**
    - Bumped `STORAGE_VERSION` to 1 (`#[pallet::storage_version]`)
    - Added `PotentialWormholeBalance` and `TotalWormholeExits` storage items (with getters)
-   - Added `is_ambiguous_account()` helper (single source of truth for the `nonce == 0` heuristic)
+   - Added `is_ambiguous_account()` helper (single source of truth for the `nonce == 0` heuristic), now also excluding `NonWormholeAccounts`
+   - Added `NonWormholeAccounts: Contains<AccountId>` config and `reveal_account()` helper (the deduction side, shared by the first-signature reveal and multisig creation)
    - Added `SoundnessInvariantViolation` error variant
    - Updated `record_transfer()` to track deposits to ambiguous addresses
    - Updated `verify_aggregated_proof()` to check the invariant and update `TotalWormholeExits`
@@ -227,10 +235,20 @@ Wormhole exits are unsigned (`ensure_none`). They don't have a signer and won't 
 2. **`pallets/wormhole/src/migrations.rs`** (new)
    - `v1::InitSoundnessCounters` + `MigrateV0ToV1` versioned migration to seed the pool
 
-3. **`runtime/src/transaction_extensions.rs`**
-   - Updated `WormholeProofRecorderExtension::validate()` to detect first-time signers and subtract their balance (and accounted for it in `weight()`)
+3. **`primitives/wormhole/src/lib.rs`**
+   - Added `reveal_address()` to the `TransferProofRecorder` trait (the cross-pallet wormhole hook), implemented by the wormhole pallet
 
-4. **`runtime/src/lib.rs`**
+4. **`pallets/multisig/src/lib.rs`**
+   - Added `is_multisig()` helper (registry lookup)
+   - Added a `ProofRecorder: TransferProofRecorder` config and a `reveal_address` call on multisig creation, so a multisig is revealed to the soundness counter when created
+
+5. **`runtime/src/transaction_extensions.rs`**
+   - Updated `WormholeProofRecorderExtension::validate()` to detect first-time signers and reveal them (and accounted for it in `weight()`)
+
+6. **`runtime/src/configs/mod.rs`**
+   - Implemented `NonWormholeAccounts` (registered multisigs + treasury + `PalletId`/sentinel keyless accounts) and wired multisig's `ProofRecorder = Wormhole`
+
+7. **`runtime/src/lib.rs`**
    - Added the `Migrations` tuple (`MigrateV0ToV1`) to `Executive` and bumped `spec_version` to 132
 
 ## Testing
@@ -238,11 +256,18 @@ Wormhole exits are unsigned (`ensure_none`). They don't have a signer and won't 
 1. **Unit tests in wormhole pallet:**
    - Test that transfers to nonce-0 addresses increment `PotentialWormholeBalance`
    - Test that transfers to nonce>0 addresses don't affect `PotentialWormholeBalance`
+   - Test that transfers to `NonWormholeAccounts` don't affect `PotentialWormholeBalance`
+   - Test that `reveal_account()` subtracts an account's balance from the pool
    - Test that exits increment `TotalWormholeExits`
    - Test that exits fail with `SoundnessInvariantViolation` when invariant would be violated
 
-2. **Integration tests in runtime:**
+2. **Multisig pallet test:**
+   - Test that creating a multisig reveals its address to the wormhole soundness counter
+
+3. **Integration tests in runtime:**
    - Test that signing a first transaction subtracts balance from `PotentialWormholeBalance`
+   - Test that creating a multisig deducts a pre-funded (pre-computed) address's balance and excludes it afterward
+   - Test that mining rewards count only the miner's (ambiguous) portion, not the excluded treasury portion
    - Test full flow: deposit -> exit -> verify counters
    - Test reveal flow: deposit -> sign transaction -> verify balance subtracted
 
--- a/docs/wormhole-soundness-detection-plan.md
+++ b/docs/wormhole-soundness-detection-plan.md
@@ -1,0 +1,253 @@
+# Wormhole Soundness Bug Detection Plan
+
+## Overview
+
+This document describes a mechanism to detect potential soundness bugs in the wormhole system (ZK proof-of-burn minting). The core invariant we want to enforce:
+
+```
+total_wormhole_exits <= potential_wormhole_balance
+```
+
+Where:
+- `potential_wormhole_balance` = sum of balances held by "ambiguous" addresses (addresses that have never signed a dilithium transaction)
+- `total_wormhole_exits` = sum of all tokens minted via wormhole exits
+
+If this invariant is ever violated, it indicates a soundness bug — more tokens are being exited than could possibly have been deposited into wormhole addresses.
+
+## Key Insight: Nonce-Based Address Classification
+
+On-chain, wormhole addresses are indistinguishable from regular dilithium addresses. However, we can use a heuristic:
+
+- **Ambiguous address**: `account_nonce == 0` (has never signed a transaction)
+- **Revealed address**: `account_nonce > 0` (has signed at least one transaction with dilithium)
+
+When an address signs its first transaction, it "reveals" itself as a regular dilithium address (not a wormhole address), and its balance should be subtracted from the potential wormhole pool.
+
+## Storage Items
+
+Add to `pallets/wormhole/src/lib.rs`:
+
+```rust
+/// Sum of balances held by ambiguous addresses (nonce == 0, might be wormhole).
+/// This represents the maximum possible value that could be legitimately exited.
+#[pallet::storage]
+pub type PotentialWormholeBalance<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
+
+/// Total value of all successful wormhole exits (mints to exit accounts).
+#[pallet::storage]
+pub type TotalWormholeExits<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
+```
+
+## Implementation
+
+### 1. Track Deposits on Every Transfer
+
+Location: `record_transfer()` in `pallets/wormhole/src/lib.rs` (around line 603)
+
+After recording the transfer in the ZK tree, check if the recipient is ambiguous:
+
+```rust
+pub fn record_transfer(
+    asset_id: T::AssetId,
+    from: &<T as Config>::WormholeAccountId,
+    to: &<T as Config>::WormholeAccountId,
+    amount: BalanceOf<T>,
+) {
+    // ... existing ZK tree recording code ...
+
+    // Soundness tracking: if recipient has never signed (nonce == 0),
+    // they might be a wormhole address, so add to potential balance
+    if asset_id == T::AssetId::default() {
+        // Native token only for now
+        let to_account: <T as frame_system::Config>::AccountId = to.clone().into();
+        let recipient_nonce = frame_system::Pallet::<T>::account_nonce(&to_account);
+        
+        if recipient_nonce.is_zero() {
+            PotentialWormholeBalance::<T>::mutate(|total| {
+                *total = total.saturating_add(amount);
+            });
+        }
+    }
+    
+    // TODO: Add similar tracking for asset transfers when asset wormhole is enabled
+
+    // ... existing event emission code ...
+}
+```
+
+### 2. Subtract Balance on First Signature (Reveal)
+
+Location: `WormholeProofRecorderExtension::validate()` in `runtime/src/transaction_extensions.rs` (around line 231)
+
+When an address signs its first transaction, subtract its balance from the potential pool:
+
+```rust
+fn validate(
+    &self,
+    origin: sp_runtime::traits::DispatchOriginOf<RuntimeCall>,
+    _call: &RuntimeCall,
+    _info: &DispatchInfoOf<RuntimeCall>,
+    _len: usize,
+    _self_implicit: Self::Implicit,
+    _inherited_implication: &impl sp_runtime::traits::Implication,
+    _source: frame_support::pallet_prelude::TransactionSource,
+) -> sp_runtime::traits::ValidateResult<Self::Val, RuntimeCall> {
+    // Soundness tracking: detect first-time signers and subtract their balance
+    // Note: We're in validate(), so nonce has NOT been incremented yet by CheckNonce
+    if let Ok(signer) = frame_system::ensure_signed(origin.clone()) {
+        let nonce = frame_system::Pallet::<Runtime>::account_nonce(&signer);
+        
+        if nonce.is_zero() {
+            // First transaction from this address — they're revealing as dilithium signer
+            // Subtract their current balance from potential wormhole balance
+            let balance = <Runtime as pallet_wormhole::Config>::Currency::free_balance(&signer);
+            
+            pallet_wormhole::PotentialWormholeBalance::<Runtime>::mutate(|total| {
+                *total = total.saturating_sub(balance);
+            });
+        }
+    }
+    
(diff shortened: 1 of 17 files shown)


n13:
## Review: Wormhole soundness turnstile

Reviewed the full diff plus surrounding code (`pallets/wormhole/src/lib.rs`, `migrations.rs`, `runtime/src/transaction_extensions.rs`, the fee/mining-reward path in `pallet-mining-rewards`, and the runtime wiring). Static review only (didn't compile — ZK deps make a build slow).

### Verdict

The core idea is sound and the implementation is careful and well-tested. The invariant `TotalWormholeExits <= PotentialWormholeBalance` is algebraically just **conservation of the ambiguous pool** (cumulative outflows <= cumulative inflows), so the accounting is coherent and the failure direction is safe — a tripped invariant rejects an exit, it never over-mints. **Mergeable as a backstop**, but there's one design limitation that meaningfully weakens its security value over time, plus an encapsulation issue worth fixing. None are hard blockers.

### Strengths

- `is_ambiguous_account()` as the single `nonce == 0` source of truth — good DRY.
- `VersionedMigration<0,1,…>` is the right pattern; fresh chains skip it and seed via the block-1 `on_initialize` genesis-endowment calls, existing chains seed once.
- Seed = `total_issuance()` is genuinely a safe upper bound — can't block a valid exit.
- Enforcement is **check-before-mint**, and `TotalWormholeExits` is committed only after every mint succeeds.
- Reveal timing reasoning is correct: all extensions' `validate` run before any `prepare`, and `CheckNonce` increments in `prepare`, so `nonce == 0` in `validate` reliably means "first signature."
- `try-runtime` feature plumbing is valid — `pallet-balances`/`pallet-zk-tree` are real (non-dev) deps, so the new entries compile.

### Findings

#### 1. The backstop loosens monotonically — keyless accounts never "reveal" (most important)

The PR claims the pool "tightens toward the true ambiguous sum as accounts reveal themselves." That's only true for accounts that can sign. **Keyless/derived accounts never sign, so their inflows enter `PotentialWormholeBalance` and are never released.**

Concrete, every block: `pallet-mining-rewards::on_finalize` mints `treasury_reward` to the treasury account and calls `record_transfer_proof`. The treasury is a `PalletId`-derived account with `nonce == 0` forever -> ambiguous -> `P += treasury_reward`. When the treasury later disburses, the reveal path never fires (treasury never signs), so `P` is not decremented. Same for multisig accounts, pure proxies, and sovereign/pallet accounts that move funds without their own nonce incrementing.

Net effect: `P` accumulates a permanent, monotonically growing term, so the margin `P - E` grows without bound from treasury emission alone. Safe for liveness (only makes false positives *less* likely), but it steadily inflates the headroom a forged exit could consume before tripping — i.e. it erodes the very backstop this PR exists to provide. The added test `counter_mining_rewards_increase_potential_balance` even bakes this in.

Suggestion: decrement `P` when value leaves known keyless accounts (treasury payouts), or exclude the treasury portion from the pool, or at minimum document that the bound is loose-and-growing rather than tightening.

#### 2. Add-side and subtract-side use different sources of truth

Deposits add the transferred `amount` (via `record_transfer`), but reveal subtracts `Currency::free_balance(signer)`. Different code paths in different crates, not guaranteed to net to zero per account:

- **Untracked credits** inflate `free_balance` without incrementing `P`, so the eventual reveal over-subtracts (pushes `P` down — the dangerous direction for liveness). Clearest example: `force_set_balance` is counted in `count_transfers()` (weight) but emits `BalanceSet`, which `record_proofs_from_events_since` does **not** scan (it only matches `Balances::Transfer` and `Balances::Minted`). So a sudo balance bump isn't tracked, yet a later reveal subtracts the full balance.
- **Reserved balance** is excluded by `free_balance`, so an account holding a reserve at reveal time subtracts less than was added.

The huge seed + `saturating_sub` absorb this, so unlikely to bite, but the coupling is fragile. Worth reconciling (track per-account contribution, or subtract total balance and ensure every credit path the reveal will later subtract is actually tracked).

#### 3. Counter is mutated from two crates (encapsulation / DRY)

The deposit increment lives in the pallet (`record_transfer`), but the reveal decrement reaches directly into pallet storage from the runtime:

```rust
pallet_wormhole::PotentialWormholeBalance::<Runtime>::mutate(|total| {
    *total = total.saturating_sub(balance);
});
```

The counter's invariant is split across the crate boundary. Cleaner to expose one pallet method that encapsulates read-balance-and-subtract, e.g. `pallet_wormhole::Pallet::<T>::on_account_revealed(&who)`, so the extension just calls it. Keeps all mutations of the counter (and the ambiguity check) in one place, consistent with the deposit side.

#### Nits

- **Enforcement is placed after `UsedNullifiers` are marked.** Correct because `ensure!` -> `Err` rolls back the whole dispatch (transactional-by-default, which existing minting/`?` paths already rely on), so a rejected exit doesn't burn its nullifiers. Slightly cleaner to check the invariant before marking nullifiers. The new tests assert end-state counters but don't independently assert nullifiers are *not* consumed on the rejection path — adding that assertion would lock in the rollback guarantee the safety story depends on.
- **Mixed token APIs:** seed/deposit use the fungible traits (`total_issuance`, `NativeBalance`) while reveal uses legacy `Currency::free_balance`. Works for native, but inconsistent.
- **Duplicate branch:** `record_transfer` now has two adjacent `if asset_id == T::AssetId::default()` blocks (new tracking + existing event). Merge them.
- **Reveal weight** (`reads_writes(2,1)`) is added unconditionally to every extrinsic, including unsigned and already-revealed signers where the real work is one nonce read. Conservative/safe, just slightly over-charged.
- **False-positive edge case:** if a genuine deposit address ever signs (reveals) before its ZK exit, the reveal subtracts its balance and the later legit exit could trip `SoundnessInvariantViolation` until other deposits grow `P`. Unusual for the privacy model, and only a contingent block (not a loss), but worth a docs note.

n13:
## Review: Wormhole soundness turnstile — verdict

Reviewed the full diff plus surrounding code (`pallets/wormhole/src/lib.rs`, `migrations.rs`, `runtime/src/transaction_extensions.rs`, the fee/mining-reward path in `pallet-mining-rewards`, and the runtime wiring). Static review only — didn't compile (ZK deps make a build slow).

### Verdict

The design is **sound and safe in the failure direction**: the invariant `TotalWormholeExits <= PotentialWormholeBalance` is just conservation of the ambiguous pool (cumulative outflows ≤ cumulative inflows), and a tripped invariant *rejects* an exit — it never over-mints. The implementation is careful and well-tested. **Mergeable as a backstop**, but it ships with one security limitation that grows over time plus a couple of correctness/encapsulation issues worth addressing first. None are hard blockers.

### On the existing reviews

- **Bugbot ("Migration omits historical exit total", High):** the symptom is real but the framing is off. `TotalWormholeExits` never existed pre-upgrade, so "seed E too" is impossible — you can't recover historical cumulative exits. The real issue is that seeding `P = total_issuance()` with `E = 0` makes the post-migration cap roughly a *full issuance* of headroom. The fix is a **tighter P seed** (below), not seeding E. I'd treat this as Medium, not High: it only loosens the bound, it never blocks a valid exit.
- **@n13's review:** accurate; I independently confirmed every substantive point against the code (see findings).

### Findings

**1. Backstop loosens monotonically — keyless accounts never reveal (most important).**
`pallet-mining-rewards::on_finalize` mints `treasury_reward` to the treasury (a `PalletId`-derived, never-signing account → ambiguous forever) and the miner, then calls `record_transfer_proof`. Both land in `P` every block, and the treasury portion is **never** decremented (treasury never signs → reveal never fires). Same for multisig/pure-proxy/sovereign accounts that move funds without their own nonce incrementing. `P` therefore accrues a permanent, monotonically growing term, inflating the margin `P − E` from emission alone and steadily eroding the very backstop this PR provides. The `counter_mining_rewards_increase_potential_balance` test bakes this in. The PR text claims the pool "tightens toward the true ambiguous sum"; that is not true for keyless accounts. *Suggestion:* decrement `P` on known keyless payouts (treasury disbursements), exclude the treasury portion, or at minimum document that the bound is loose-and-growing rather than tightening.

**2. Add-side and subtract-side use different sources of truth.**
Deposits add the transferred `amount` (via `record_transfer`); reveal subtracts `Currency::free_balance(signer)`. These aren't guaranteed to net to zero per account. `record_proofs_from_events_since` only matches `Balances::Transfer/Minted` and `Assets::Transferred/Issued` — **not** `Balances::BalanceSet`. Yet `count_transfers` counts `force_set_balance` (for weight). So a sudo `force_set_balance` raises a balance that was never added to `P`, but a later reveal subtracts the full balance → over-subtraction (the liveness-dangerous direction). Reserved balance is likewise excluded by `free_balance`. The huge migration seed + `saturating_sub` hide this on an existing chain, but on a fresh chain (tight seed) it could wrongly trip `SoundnessInvariantViolation` and block a legit exit. Root-gated, so low probability — but it's a silent over-subtraction. *Suggestion:* track per-account contribution, or ensure every credit path the reveal later subtracts is actually tracked.

**3. Storage mutation inside `validate()` (new — not raised yet).**
The reveal performs a storage write in `validate()`:

```rust
if let Ok(signer) = ensure_signed(origin.clone()) {
    if pallet_wormhole::Pallet::<Runtime>::is_ambiguous_account(&signer) {
        let balance = <Balances as Currency<AccountId>>::free_balance(&signer);
        if balance > 0 {
            pallet_wormhole::PotentialWormholeBalance::<Runtime>::mutate(|t| *t = t.saturating_sub(balance));
        }
    }
}
```

`validate()` is the pool-validation entry point and is intended to be side-effect-free (it can be called repeatedly against discarded overlays). The idiomatic place for committed pre-dispatch state changes is `prepare()`. It happens to work today (pool runs are discarded; the committed write occurs once during block execution), but it's fragile and inconsistent with the rest of this very extension, which mutates in `post_dispatch`. Cleaner: detect ambiguity + capture the balance in `validate`, return it via `Val`, and apply the subtraction in `prepare`.

Related: the new integration tests are good but they call `validate → prepare → dispatch → post_dispatch` directly, not through `frame_executive::apply_extrinsic`. So the two things the safety story leans on — that validate-phase writes commit in the real pipeline, and that all extensions' `validate` run before any `prepare` (relative to `CheckNonce`) — are assumed by construction, not exercised end-to-end.

**4. Counter is mutated from two crates (encapsulation / DRY).**
The deposit increment lives in the pallet (`record_transfer`), but the reveal decrement reaches into pallet storage from the runtime (`pallet_wormhole::PotentialWormholeBalance::<Runtime>::mutate(...)`). The counter's invariant is split across the crate boundary. Cleaner to expose one pallet method, e.g. `Pallet::<T>::on_account_revealed(&who)`, so the extension just calls it — keeping all mutations (and the ambiguity check) in one place, consistent with the deposit side.

### Concrete suggestions (ranked)

1. **Tighten the migration seed.** Seed `P` to the sum of balances over `nonce == 0` accounts instead of `total_issuance()`. That's the true current ambiguous pool, makes `E = 0` actually correct, gives a meaningful cap immediately, and resolves the Bugbot concern. Cost is account iteration — fine for a young chain, or use a multi-block migration. If you keep `total_issuance()`, document that the post-migration cap is ~full issuance.
2. **Address the keyless/treasury monotonic growth** (finding 1).
3. **Encapsulate the counter** via `on_account_revealed`, and move the mutation into `prepare()` (findings 3 + 4).
4. **Reconcile add vs subtract sources** (finding 2), notably `force_set_balance`/`BalanceSet` and reserved balance.

### Nits

- **Duplicate adjacent `if asset_id == T::AssetId::default()` blocks** in `record_transfer` — merge them.
- **Soundness check is placed after `UsedNullifiers` are marked.** Correct only because a returned `Err` rolls back the whole transactional dispatch; slightly cleaner to check before marking. The rejection-path tests should also assert nullifiers are **not** consumed, to lock in the rollback guarantee the safety story relies on.
- **Reveal weight** (`reads_writes(2,1)`) is added unconditionally to every extrinsic, including unsigned/already-revealed — conservative but over-charged.
- **Mixed token APIs:** migration/deposit use the fungible traits; reveal uses legacy `Currency::free_balance`. Works for native, but inconsistent.
- **False-positive edge case:** if a genuine deposit address ever signs (reveals) before its ZK exit, the reveal subtracts its balance and the later legit exit could trip `SoundnessInvariantViolation` until other deposits grow `P`. Contingent block, not a loss — worth a docs note.

### Verified correct

- `VersionedMigration<0,1,…>`, `STORAGE_VERSION = 1`; fresh chains skip the migration (genesis writes version 1) and seed via block-1 `record_transfer` calls; existing chains seed once.
- Check-before-mint; `TotalWormholeExits` committed only after all mints succeed.
- Exiting into an ambiguous address re-adds to `P` and is balanced by the reveal subtraction — traced the full path; it does **not** allow unbounded extraction to revealed accounts (margin is correctly paid back on reveal).
- `try-runtime` plumbing valid: `pallet-balances` and `pallet-zk-tree` are real (non-dev) deps of `pallet-wormhole`.

n13:
If I read this correctly
- We're missing treasury events - we need to "decrement P when value leaves a known keyless account" or "exclude the treasury portion"
- Missing force_set_balance (sudo, I don't think we should call this, but still, should also not screw up the turnstile)
- validate() is the wrong place to modify the counter... see (3) above, we should fix this

n13:
## Re-review (commits `27010d28`, `2c902192`, `d94f2f14`)

Thanks for the fast turnaround — all three findings from the first pass are addressed, and the `validate` side-effect concern was fixed properly too. Re-reviewed statically against the rest of the runtime (didn't rebuild).

### Verdict: LGTM / approve

The turnstile is now materially tighter and the bookkeeping is cleanly encapsulated. Two low-severity, **non-blocking** follow-ups below.

### Findings addressed

- **Keyless accounts loosening the backstop** — fixed via `NonWormholeAccounts: Contains<AccountId>` (excludes registered multisigs, the configured treasury, and the keyless `PalletId`/sentinel accounts), so their balances no longer inflate the pool. The mining test now asserts only the miner's ambiguous portion enters the pool and that it's strictly less than the full emission — exactly the treasury case I flagged. 
- **Multisig pre-funding gap** (the subtle one this surfaces) — handled correctly: a pre-computed multisig address looks ambiguous, so a pre-fund is counted, and `create_multisig` calls `reveal_address` *after* registering the address, deducting its balance so it nets zero, and `is_multisig` excludes it thereafter. Nice catch and good test (`counter_multisig_creation_reveals_prefunded_address`). Ordering (insert → reveal) is correct.
- **free vs total balance** — now subtracts `Currency::total_balance` (free + reserved); the "ambiguous accounts can't spend, so total == sum of credits" justification is sound. `force_set_balance` removed from `count_transfers` (it emits `BalanceSet`, never recorded) and the over-subtraction caveat is documented on `reduce_potential_balance`. 
- **Encapsulation / DRY** — `reveal_account` / `reduce_potential_balance` pallet methods + the `TransferProofRecorder::reveal_address` trait hook mean the runtime and multisig pallet no longer poke pallet storage directly. 
- **Bonus:** moving the mutation out of `validate` (capture balance in `Val`, commit in `prepare`) is exactly right — `validate` is now side-effect free, and capturing the pre-tx balance keeps the subtraction correct regardless of fee-charge ordering.

### Remaining (non-blocking)

1. **The `total_balance` reveal over-subtracts for *any* untracked credit path, not just `force_set_balance`.** The invariant the reveal relies on is "every credit the account received was added via `record_transfer`." That also breaks for governance-scheduled transfers (the scheduler is `EnsureRoot`, dispatches in `on_initialize` with no transaction extension, so the `Transfer`/`Minted` event is never scanned → pool not incremented), and for any future pallet-internal credit. All such paths are privileged today, so the risk is bounded and in the safe-ish (liveness, not safety) direction, mitigated by the conservative seed. Worth either generalizing the caveat doc beyond `force_set_balance`, or — for true robustness — tracking a per-account contributed amount and subtracting that instead of `total_balance`.

2. **Weight under-counts the new storage reads.** `is_ambiguous_account` now also reads `Multisigs` (`is_multisig`) and `TreasuryAccount` (`treasury_account()`) on every call — i.e. once per recorded transfer (deposit side) and once per signed tx (reveal detection). `reveal_weight = reads_writes(2, 1)` and the per-transfer `count_transfers` weight don't include these extra ~2 reads (nor the conditional pool write in `record_transfer`). Minor, but worth tightening so block weight stays a true upper bound.

### Nits

- `NonWormholeAccounts::contains` rebuilds the 5-element keyless array (with `into_account_truncating` hashing) on every call; could hoist to a `parameter_types!`/const.
- The treasury is excluded twice — once via the configured `treasury_account()` getter (the real one) and once via `TreasuryPalletId::…into_account_truncating()` (a different, likely-unused derived account). Harmless, but the comment calling the latter "the treasury" is slightly misleading since the treasury account is configurable in this fork.

n13 (approved):

PR #600, opened 2026-06-19
Inline pallets

illuzen:
Just vendor / inline some pallets so we can have our trusted computing base fully under our control.

---

> [!NOTE]
> **High Risk**
> Inlining pallet-balances and related core FRAME pallets changes how account balances and issuance are built and reviewed; any drift from upstream Substrate behavior would affect the entire chain economy.
> 
> **Overview**
> Moves several **polkadot-sdk** pallets from **crates.io** into the repo so the runtime’s trusted computing base is built from local sources.
> 
> **`Cargo.toml`** now resolves **`pallet-balances`**, **`pallet-preimage`**, **`pallet-recovery`**, **`pallet-timestamp`**, and **`pallet-utility`** via **`path = "./pallets/..."`**, adds those crates to the workspace members, and wires a workspace **`frame`** alias for **`polkadot-sdk-frame`**. **`Cargo.lock`** is updated for path-based packages and dev-test dependencies (e.g. **`pallet-utility`** tests pulling in **`pallet-balances`**, **`pallet-collective`**, **`pallet-root-testing`**, **`pallet-timestamp`**).
> 
> The diff includes a full in-tree **`pallet-balances`** (v46.0.0): lib, currency/fungible impls, migrations, benchmarks, weights, and extensive unit tests—intended as a vendored copy rather than new balance semantics in this PR.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fdfa5b18fec393d2bfa31b7ada795ece8bb9a54c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13:
# Review: PR #600 "Inline pallets" ([link](https://github.com/Quantus-Network/chain/pull/600))

## What it does
Vendors 5 polkadot-sdk pallets from crates.io into `pallets/`, switching the workspace deps from `version = "..."` to `path = "./pallets/..."`:

| Pallet | Version | Notes |
|---|---|---|
| `pallet-timestamp` | 44.0.0 | |
| `pallet-balances` | 46.0.0 | |
| `pallet-utility` | 45.0.0 | |
| `pallet-preimage` | 45.0.0 | |
| `pallet-recovery` | 45.0.0 | pulls in `frame` (`polkadot-sdk-frame` 0.14.0) umbrella crate |

## Fidelity — verified clean
I downloaded each exact upstream version from crates.io and diffed against the vendored copies. **Every `src/` file is byte-for-byte identical to upstream.** The only differences are the expected, correct ones:
- Omitted crates.io packaging artifacts (`.cargo_vcs_info.json`, `Cargo.lock`, `Cargo.toml.orig`) — correct.
- `Cargo.toml` rewritten to use workspace deps. Versions line up with the workspace (`frame-support 45.1.0` satisfies `^45.0.0`, `sp-core 39.0.0`, `sp-runtime 45.0.0`, `sp-io 44.0.0`, etc.), so no resolved-version drift vs. the previous crates.io deps.

So the trusted-computing-base goal is met with no behavioral change. 

The dev-dependencies were correctly *reconstructed* (crates.io strips them): `pallet-balances` for recovery; `pallet-collective 45.0.0` + `pallet-root-testing 21.0.0` + `pallet-timestamp` for utility. `Cargo.lock` changes are limited to the path switch plus those two new dev-deps. Scope is clean — nothing outside `Cargo.{toml,lock}` and `pallets/`.

## Blocking issue — Clippy CI is failing ❌
`🤖 Analysis (Clippy & Doc)` fails: `could not compile pallet-balances (lib) due to 13 errors`.

Root cause: each vendored `Cargo.toml` uses

```toml
[lints]
workspace = true
```

which inherits the workspace's strict production lints:

```45:51:Cargo.toml
[workspace.lints.clippy]
# Deny unwrap and panic in production code (runtime & pallets)
# Test code can opt-out using: #![allow(clippy::unwrap_used, clippy::panic)]
panic = "deny"
unwrap_used = "deny"
# Also deny expect_used for consistency
expect_used = "deny"
```

Upstream uses `unwrap()`/`expect()`/`panic!` throughout (lib + benchmarks + tests), and its real manifest sets `[lints.clippy.all] = "allow"`, which is why it passes upstream CI. Applying Quantus's deny lints to verbatim third-party code breaks the build.

**Recommended fix (preserves verbatim vendoring):** replace `[lints] workspace = true` in each vendored `Cargo.toml` with upstream's own permissive lint table (the `allow`-all clippy config from `Cargo.toml.orig`). Editing the `.rs` files to add `#![allow(...)]` would work too but breaks the "identical to upstream" property and complicates future syncs, so I'd avoid it.

## Other notes
- `Build & Test Matrix` (ubuntu/macos) is still **pending** — compilation/tests not yet confirmed. Worth waiting on before merge.
- Adding the `polkadot-sdk-frame` umbrella (for recovery) is a heavyweight dependency, but it's exactly what upstream recovery requires — acceptable.

## Verdict
Vendoring is faithful and the wiring is correct — this is a clean, low-risk-of-drift change despite the "High Risk" label. **Don't merge until the Clippy lint config on the vendored crates is fixed and the build matrix goes green.**

Want me to apply the lint-config fix (swap `[lints] workspace = true` for the upstream permissive table in each vendored `Cargo.toml`)?

n13:
I think we need to change the lint settings in the vendored crates to be less strict, as suggested above

PR #602, opened 2026-06-23
Inline frame core

illuzen:
Inline all frame-* crates

---

> [!NOTE]
> **High Risk**
> This replaces consensus-critical FRAME glue (executive, support, proc macros) that compiles into runtime WASM; any drift from upstream or patch resolution mistakes could affect block execution and metadata across the chain.
> 
> **Overview**
> **Vendors the FRAME runtime stack** (`frame-support`, proc-macro crates, `frame-executive`, metadata hash extension, benchmarking, try-runtime, system RPC APIs, etc.) under `frame/` and wires the workspace to **path dependencies** plus **`[patch.crates-io]`** so the runtime and transitive deps resolve to these copies instead of crates.io.
> 
> **`docs/RUNTIME_SURFACE.md`** is updated to describe the new in-tree FRAME section and to label many pallets as **inlined** rather than upstream-only.
> 
> **`primitives/high-security`** is added to the workspace members/deps alongside expanded workspace keys for FRAME’s macro/benchmark toolchain (`pretty_assertions`, `aquamarine`, additional `sp-*` crates).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c423cbf154089b74c16a2d5d8f05d7326cc2bb67. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (commented):
## Review: Inline frame core

Reviewed by checking out the branch and diffing the non-vendored config against `main`, verifying `Cargo.lock` resolution, and — for this consensus-critical code — **diffing every vendored crate's `src/` against the actual upstream crates downloaded from crates.io**.

### Verdict

**Strong, low-risk vendoring PR.** This is a mechanical vendor-and-patch of the FRAME runtime stack, done correctly. The ~82k additions are almost entirely byte-faithful upstream copies; the meaningful surface (workspace `Cargo.toml`, `[patch.crates-io]`, `Cargo.lock`, doc) is small and correct. A few cosmetic nits, and one thing to wait on (the WASM build matrix is still pending). Recommend merge once Build & Test goes green.

### What I verified

**Source drift (the critical check) — clean.** Diffed `src/` for all 11 vendored crates against the published crates.io versions:

| Crate | Claimed ver | Source vs upstream |
| --- | --- | --- |
| frame-support | 45.1.0 | identical |
| frame-executive | 45.0.1 | identical |
| frame-benchmarking | 45.0.3 | identical |
| frame-support-procedural (+tools, +tools-derive) | 36.0.0 / 13.0.1 / 12.0.0 | identical |
| frame-system-benchmarking | 45.0.0 | identical |
| frame-system-rpc-runtime-api | 40.0.0 | identical |
| frame-try-runtime | 0.51.0 | identical |
| frame-metadata-hash-extension | 0.13.0 | identical |
| frame-metadata | 23.0.1 | **rustfmt-only** diffs |

The only crate that differs is `frame-metadata` (it ships from a different upstream repo with a different rustfmt config, then got `cargo fmt`'d here). Stripping comments + whitespace confirmed the differences are purely `imports_granularity` grouping and trailing-comma removal — **no semantic change**.

- **`[features]` wiring matches upstream** for frame-support/executive/benchmarking (only tab-vs-space indentation differs, a cargo `.crate` normalization artifact). Important since feature flags drive what compiles into WASM.
- **Lockfile resolves correctly.** Every vendored crate dropped its `source`/`checksum` in `Cargo.lock` (patches applied), with **no duplicate crates.io copies** and no version conflicts. `[patch.crates-io]` correctly forces transitive dependents to the in-tree copies, including the `pallet-transaction-payment` family.
- **Lints handled deliberately.** No vendored crate opts into the workspace `unwrap_used/expect_used = deny` (no `[lints] workspace = true`), with explanatory comments — which is why clippy passes.
- **CI:** Clippy & Doc, Format, Bugbot all green; Build & Test matrix still pending at review time.

### Findings (all minor)

1. **Dead `[workspace.dependencies]` keys.** `derive-syn-parse`, `expander`, and `proc-macro-warning` were added to workspace deps, but their only consumer (`frame/support-procedural`) uses direct version literals instead of `.workspace = true`, so the keys are unused. Either wire `frame-support-procedural` to `.workspace = true`, or drop the three keys. (`macro_magic` is similarly split — could be `{ workspace = true, features = ["proc_support"] }`.)
2. **`default-features = false` in `[patch.crates-io]` is a no-op.** Cargo ignores `features`/`default-features`/`optional` in patch entries — only the path matters. The new entries carry it (misleading, and inconsistent with the existing `frame-system`/`litep2p` patch entries that omit it). Cosmetic.
3. **Vendored manifests dropped upstream's `[lints.rust.unexpected_cfgs]` `check-cfg`.** Harmless here — none of these crates' sources use `substrate_runtime`/`build_opt_level`/`build_profile` cfgs, and there's no global `-D warnings`. Worth knowing for future re-syncs.

### Recommendations

- **Wait for the WASM Build & Test matrix before merging** — std clippy + identical source + clean lock make the no_std/wasm risk low, but that's the build that actually compiles the consensus runtime.
- **Maintenance:** vendoring this stack means owning upstream security/bug fixes with no auto-sync. The per-commit version notes are good; consider documenting the exact upstream tag/commit per crate and a re-sync procedure so future drift stays auditable.

<sub>🤖 AI-assisted review (Cursor).</sub>

n13 (approved):
Looks good!

Need to fix CI then GTG

PR #605, opened 2026-06-24
Tech collective: clean up stale votes after member removal (F-91265)

n13:
## Summary
Adds the substantive fix from the v12 suggestions (authored by `illuzen`) that closes audit finding **F-91265** (a removed member's vote stays in an ongoing poll's tally) in the surviving tech-collective lane.

- **Cherry-picks** `allow removing votes from poll if member removed` (illuzen): adds a permissionless `remove_vote_for_non_member(who, poll)` extrinsic to `pallet-ranked-collective` (`call_index(7)`) that clears a non-member's stale vote from an **Ongoing** poll and corrects the tally. Guarded to only act on accounts that are **not** current members (`StillMember`), that actually voted (`NotVoted`), on polls still ongoing (`NotPolling`). Fee waived (cleanup op). Ships 6 unit tests.
- Bumps runtime `spec_version` 133 → 134 (new extrinsic).
- Documents the call in `docs/TECH_REFERENDA_FEATURES.md`.

## The second suggested commit was intentionally dropped
The other suggested commit (`fix zero-denominator issue` in `pallets/conviction-voting`, F-91141) does **not** apply here: `pallet-conviction-voting` was removed entirely in #604, so the bug is moot. The surviving tech-lane tally already guards the zero-denominator case:

```
Perbill::from_rational(self.ayes, 1.max(self.ayes + self.nays))
```

## Test plan
- [x] `cargo test -p pallet-ranked-collective` → 21 passed (incl. 6 new: `remove_vote_for_non_member_works`, `..._removes_nay_vote`, `..._fails_if_not_voted`, `..._fails_on_completed_poll`, `..._fails_if_still_member`, `remove_member_cleanup_works`)
- [x] `cargo check -p quantus-runtime` → clean; extrinsic exposed via `TechCollective`

## Notes
- New extrinsic at a fresh `call_index(7)` is additive/backward-compatible → `transaction_version` left unchanged.
- Weight reuses `WeightInfo::vote()` (O(1), comparable work) → no new benchmark required.

### DIFF
--- a/docs/TECH_REFERENDA_FEATURES.md
+++ b/docs/TECH_REFERENDA_FEATURES.md
@@ -47,6 +47,7 @@ Referenda's `Tally = pallet_ranked_collective::TallyOf<Runtime>` and
 | `exchange_member` | `NeverEnsureOrigin` → **disabled** | Account swap unreachable |
 | `vote(poll, aye)` | member | Rank-weighted aye/nay; re-votable while Ongoing |
 | `cleanup_poll` | signed | GC vote records after a poll ends |
+| `remove_vote_for_non_member` | **permissionless** | Clear a removed member's stale vote from an Ongoing poll's tally (#91265) |
 
 ### `TechReferenda` (referenda `Instance1`)
 | Call | Origin | Notes |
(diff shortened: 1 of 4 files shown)


n13:
The edge case here is

What if a poll is ongoing, and a tech collective member is removed while the poll is ongoing.

In our current setting only 1 poll can be active at a time

Therefore you can't remove a member while the poll is ongoing (needs a 2nd poll)

also I don't see why this is a better solution - the member was part of the tech collective when the voting began, then voted...

Removing them if they're no longer eligible is a choice, but keeping them is equally valid...

I don't think we should do this

Claude opened a PR for it already, but I think it just adds some complications that then also have other impacts potentially, less code is more

PR #606, opened 2026-06-24
Streamline CI

illuzen:
- parallel tests
- download taplo binary instead of building from source
- don't cargo clean before tests
- cache build artifacts with rusty-cache

---

> [!NOTE]
> **Low Risk**
> Workflow-only changes; test execution still uses the same locked workspace and feature flags, with only the runner and caching strategy changed.
> 
> **Overview**
> **CI is tuned for faster feedback** by caching Rust builds, compiling through **sccache**, and running tests with **cargo nextest** instead of plain `cargo test`.
> 
> The **build-and-test** and **analysis** jobs now set `SCCACHE_GHA_ENABLED` / `RUSTC_WRAPPER`, add **mozilla-actions/sccache-action**, and **Swatinem/rust-cache** with per-job `shared-key` values (including OS-specific keys on the matrix). The pre-test **`cargo clean`** step is removed so the compile step’s artifacts can be reused for tests.
> 
> **Fast checks** install **taplo** from a released Linux binary via `curl` rather than `cargo install taplo-cli`, avoiding a source build on every run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f5ca2605cc94c66b62d45507eb0e91a2a58de90. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,8 @@ env:
   CARGO_TERM_COLOR: always
   CARGO_PROFILE_DEV_DEBUG: 1
   CARGO_PROFILE_TEST_DEBUG: 1
+  SCCACHE_GHA_ENABLED: "true"
+  RUSTC_WRAPPER: "sccache"
 
 jobs:
   fast-checks:
@@ -35,7 +37,10 @@ jobs:
       - name: Install required components
         run: rustup component add rustfmt --toolchain nightly
       - name: Install taplo
-        run: cargo install taplo-cli --locked
+        run: |
+          curl -fsSL https://github.com/tamasfe/taplo/releases/latest/download/taplo-linux-x86_64.gz | gunzip > taplo
+          chmod +x taplo
+          sudo mv taplo /usr/local/bin/
       - name: Run format checks
         run: |
           taplo format --check --config taplo.toml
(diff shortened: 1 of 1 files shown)


n13 (approved):
## Review — Approve

Workflow-only change, and all checks (fast-checks, build & test matrix on ubuntu + macos, analysis) are green on the head commit. The one Bugbot finding (nextest doesn't run doctests) was correctly resolved by the `run doctests` commit, which adds a dedicated `cargo test --doc` step. LGTM.

### What's good
- `sccache` + `Swatinem/rust-cache` + dropping `cargo clean` should meaningfully cut warm-cache build times.
- `cargo nextest run` for parallel test execution.
- Downloading the prebuilt `taplo` binary instead of `cargo install taplo-cli --locked` avoids a from-source build each run (verified the release asset URL resolves).
- Per-job `shared-key` values (including OS in the matrix) keep the caches correctly partitioned.

### Non-blocking suggestions
1. **Pin `taplo` instead of `latest`.** `releases/latest/download/...` is non-reproducible — a new taplo release can silently change formatting rules (or asset layout) and break CI without any change on our side. Pin to an explicit version tag to keep formatting deterministic.
2. **Watch disk usage from removing `cargo clean`.** The old step was labeled "to free space," and now `target/` is also cached. The matrix builds the workspace with `runtime-benchmarks,try-runtime` (compile) and again with `fast-governance` (test + doctest), so on the standard hosted runners this could get tight. It passed this run (ubuntu ~3h16m), so fine for now — just flagging it as a fragility to keep an eye on.
3. **sccache + rust-cache is somewhat redundant.** rust-cache already caches `target/` and the registry; sccache adds a separate object cache on top. Harmless, but the marginal benefit may be small — and sccache won't cache the `clippy-driver` invocations in the analysis job, where rust-cache does the heavy lifting.
4. **Branch is named `beefier-runners` but `runs-on` is still `ubuntu-latest`/`macos-latest`.** If bumping to larger runners was part of the intent, it's not in this diff — was that deferred intentionally?

None of these block merge.

PR #610, opened 2026-06-29
Fix: origin-rewriting wrappers bypass high-security restrictions

n13:
## Summary

High-security (HS) accounts are meant to be limited to whitelisted calls (e.g. reversible `schedule_transfer`/`cancel`). They could be bypassed: `Recovery::as_recovered` and `Utility::as_derivative` synthesize a fresh `Signed` origin **after** top-level transaction validation, so a non-HS outer signer could dispatch a non-whitelisted call *as* an HS account — including when nested under `batch`/`batch_all`/`force_batch`/`if_else`.

This PR enforces the HS whitelist at the **effective dispatched origin**, entirely at the runtime/extension layer:

- `HighSecurityConfig::call_allowed_for` recursively resolves origin-rewriting wrappers (`as_recovered`/`as_derivative`) and origin-preserving combinators (`batch*`/`if_else`), re-checking the whitelist at each effective signer. Root-only `dispatch_as`/`with_weight` and the scheduler are intentionally not traversed.
- Shared via a new `HighSecurityInspector::is_call_allowed` provided method, so `ReversibleTransactionExtension` and `pallet-multisig` use one implementation (no duplicated logic).
- Recursion is **bounded and fail-closed** (`MAX_CALL_DEPTH = 16`, well below `MAX_EXTRINSIC_DEPTH = 256`): over-nested calls are rejected rather than allowed to escape the whitelist, so validation work can't be exhausted.
- The extension `weight()` scales with the number of nodes traversed (one `is_high_security` read per node).
- `pallet-recovery` and `pallet-utility` are left **unchanged from upstream Substrate**.

## Test plan

- [ ] `cargo test -p quantus-runtime --lib transaction_extensions::tests` — origin-rewriter bypass blocked, whitelisted inner call allowed, nested-in-batch bypass blocked, non-HS origin unaffected, over-nested call rejected fail-closed
- [ ] `cargo test -p pallet-multisig`
- [ ] `cargo test -p pallet-reversible-transfers`
- [ ] Runtime integration tests (recovery / reversible)

---

> [!NOTE]
> **High Risk**
> This is a security fix for account restriction enforcement; incorrect recursion or depth handling could still allow HS bypass or reject valid transactions.
> 
> **Overview**
> Closes a bypass where **high-security (HS)** call limits could be evaded by wrapping disallowed inner calls in **`Recovery::as_recovered`** or **`Utility::as_derivative`** (and **`batch*` / `if_else`**), because those pallets rewrite the signed origin after top-level validation.
> 
> **`HighSecurityInspector::is_call_allowed`** is added with a default that only checks the top-level call; the runtime overrides it via **`HighSecurityConfig::call_allowed_for`**, which recursively walks origin-rewriting and batching wrappers, re-applies the HS whitelist at each effective signer, and **rejects over-deep nesting** (`MAX_CALL_DEPTH = 16`, fail-closed). **`ReversibleTransactionExtension`** now uses this helper instead of a shallow HS + whitelist check, and its **`weight()`** scales with **`high_security_read_count`**. **`pallet-multisig` `execute`** re-checks with **`is_call_allowed`** instead of only testing the outer call against the whitelist.
> 
> Runtime tests cover blocked/allowed paths for **`as_recovered`**, **`as_derivative`**, batch-wrapped bypasses, and excessive nesting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32c521abf8e0ae3737c178a0cfe920c2aa3f8536. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6642,6 +6642,7 @@ dependencies = [
  "pallet-balances",
  "parity-scale-codec",
  "polkadot-sdk-frame",
+ "qp-high-security",
  "scale-info",
 ]
 
@@ -6813,6 +6814,7 @@ dependencies = [
  "pallet-root-testing",
  "pallet-timestamp",
  "parity-scale-codec",
+ "qp-high-security",
  "scale-info",
  "sp-core",
  "sp-io",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8079,6 +8079,9 @@ dependencies = [
 [[package]]
 name = "qp-high-security"
 version = "0.1.0"
+dependencies = [
+ "parity-scale-codec",
+]
 
 [[package]]
 name = "qp-plonky2"
--- a/pallets/multisig/src/lib.rs
+++ b/pallets/multisig/src/lib.rs
@@ -1130,9 +1130,7 @@ pub mod pallet {
 			// or the whitelist may have been updated via runtime upgrade.
 			// This prevents bypassing HS restrictions by proposing before enabling HS.
 			// After decode + get_dispatch_info, don't refund - burn the full reserved weight.
-			if T::HighSecurity::is_high_security(&multisig_address) &&
-				!T::HighSecurity::is_whitelisted(&call)
-			{
+			if !T::HighSecurity::is_call_allowed(&multisig_address, &call) {
 				return Self::err_burn_full(Error::<T>::CallNotAllowedForHighSecurityMultisig);
 			}
 
--- a/pallets/multisig/src/lib.rs
+++ b/pallets/multisig/src/lib.rs
@@ -621,7 +621,8 @@ pub mod pallet {
 			// ===== PHASE 4: High-security whitelist check (if applicable) =====
 			// (additional read: HighSecurityAccounts)
 			let is_high_security = T::HighSecurity::is_high_security(&multisig_address);
-			if is_high_security && !T::HighSecurity::is_whitelisted(&decoded_call) {
+			// Use the shared `is_call_allowed` policy so `propose` and `execute` stay consistent.
+			if !T::HighSecurity::is_call_allowed(&multisig_address, &decoded_call) {
 				// Don't refund after decode - same reasoning as above.
 				return Self::err_burn_full(Error::<T>::CallNotAllowedForHighSecurityMultisig);
 			}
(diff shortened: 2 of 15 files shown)


n13 (commented):
Verdict: Approve — no remaining blockers from this review.

Update: the only blocker from my earlier review (the failing Format check) is resolved. Commit `e196524a` ("format") makes the `Fast Checks (Format)` job pass, and `git show -w` confirms it is whitespace-only (rustfmt line-reflow in `runtime/src/transaction_extensions.rs` — `weight()`, `validate_with()`, and two test constructors), so the logic and tests validated earlier are unaffected. (Build & Test / Clippy & Doc were still running at review time — merge once they go green.)

Effective-origin high-security enforcement (reconfirmed):
- The recursive resolver re-checks the whitelist at the effective dispatched origin, covering the origin-rewriting wrappers (`Recovery::as_recovered`, `Utility::as_derivative`) and the origin-preserving combinators (`batch` / `batch_all` / `force_batch` / `if_else`).
- The remaining `pallet-utility` origin-altering calls — `dispatch_as`, `with_weight`, `dispatch_as_fallible` — are all `ensure_root`, so they are not an unprivileged bypass and correctly are not traversed. The scheduler's extrinsics are disabled.
- `ReversibleTransactionExtension` and `pallet-multisig` share the single `HighSecurityInspector::is_call_allowed` implementation (no duplicated logic); multisig re-checks at execute time against the multisig address.
- Recursion is bounded (`MAX_CALL_DEPTH = 16`) and fails closed, well under the `MAX_EXTRINSIC_DEPTH = 256` decode limit; `weight()` scales with the number of traversed nodes.
- Recovery address handling is safe: this runtime sets `type Lookup = AccountIdLookup`, which only resolves `MultiAddress::Id`, so the non-`Id` arm cannot dispatch and is not an alternate-encoding bypass.

No security or logic blocker found.

Local tests (still valid — the format commit is whitespace-only):
- `cargo test -p quantus-runtime --lib transaction_extensions::tests` passed
- `cargo test -p pallet-multisig` passed
- `cargo test -p pallet-reversible-transfers` passed

n13 (commented):
Updated verdict after the format commit: previous blocker cleared.

The new `format` commit fixed the issue I flagged; GitHub's Fast Checks (Format) job is now passing. My code-review verdict is approve/merge-ready from the effective-origin high-security enforcement perspective.

Remaining note: the build/test matrix and clippy/doc jobs are still running on the latest commit, so I would wait for those checks to finish green before merging.

illuzen (approved):

PR #617, opened 2026-07-03
Frame fixes medium low

illuzen:
# Harden FRAME and runtime against V12 medium-severity audit findings

## Summary

This PR addresses **27 Medium-severity findings** from the V12 audit (`frame-medium-low.md`). Each item was triaged: **25 were confirmed and fixed**, **2 were acknowledged and documented** (no code change). Fixes are intentionally minimal — targeted at the specific failure mode, with regression tests where appropriate.

The branch stacks on earlier `illuzen/frame-fixes` work (zk root validation, genesis bounds, bounded migrations, token accounting, etc.). Against `main` this is **41 commits** across **67 files** (~3k insertions).

### Disposition

| Status | Count |
|--------|-------|
| Valid (fixed) | 25 |
| Acknowledged (documented) | 2 |
| Invalid | 0 |

---

### Macro, metadata, and codegen

- **Unknown module name panics** (#93203) — `get_call_names` returns an empty slice for unknown modules instead of panicking.
- **Eager task enumeration exhausts memory** (#93241) — task enumeration streams lazily instead of collecting all tasks into memory up front.
- **New-prefix migrations skip** (#93290) — new-pallet storage version is initialized *after* migrations, not before.
- **Trailing Weight Tokens Ignored** (#93294) — trailing tokens in `#[pallet::weight(...)]` are rejected at compile time instead of silently ignored.
- **Unmetered view-function decoding** (#93295) — view-function inputs are decoded with a memory limit and a full-consumption check.
- **Ambiguous Genesis Field Retention** (#93308) — genesis field retention uses serde-exact camelCase comparison.
- **Opaque Wrapper Breaks Bounds** (#93357) — `WrapperKeepOpaque` enforces `MaxEncodedLen` on the wrapped type.
- **Unbounded runtime metadata decoding** (#93479) — `META_RESERVED` is validated before decoding metadata payloads.
- **Counted N-map counter prefixes bypass collision checks** (#93481) — `CountedStorageNMap` counter prefixes are included in the procedural duplicate-prefix check.

### Storage, migrations, and counted maps

- **Corrupt Takes Persist** (#93324) — undecodable explicit child-storage entries are cleared in `take()`.
- **Overlay Work Misreported** (#93325) — `child::clear_storage` counters account for overlay-only removals.
- **Overlapping Prefix Migration** (#93334) — `move_prefix` rejects overlapping `from`/`to` prefixes to avoid self-reprocessing.
- **Default-Bound Append Bypass** (#93339) — `CountedStorageMap::try_append` enforces bounds against the effective `OnEmpty` default, not just stored length.
- **Set Append Stores Duplicates** (#93340) — `BTreeSet` storage appends deduplicate entries.
- **Prefix Clear Resets Counter** (#93347) — partial `clear_prefix` subtracts only the removed entries instead of killing the global counter.
- **Prefix clear fallback fabricates progress** (#93482) — resumed `unhashed::clear_prefix` walks from the supplied cursor and makes real progress.
- **Counted storage helpers corrupt key cardinality counters** (#93483) — `try_append`, `migrate_key`/`migrate_keys`, and `CountedStorageNMap::set` reconcile counters against actual key existence.
- **Drainable counted-storage iterators desync counters** (#93484) — all drain-capable counted map/N-map iterators (`iter_keys`, `iter_prefix_values`, etc.) wire through `OnRemovalCounterUpdate`; `KeyPrefixIterator` gains an `OnRemoval` hook.

### Scheduler, preimage, and tickets

- **Preimages Persist After Failed Scheduling** (#93360) — noted preimages are dropped when scheduling fails.
- **Stale Tickets Clear Locks** (#93404) — zero-cost lone freeze/hold consideration tickets are rejected.

### Token / fungible traits

- **Held transfers bypass accounts** (#93409) — `OnHold` transfers enforce destination account existence.

### Frame-system and block limits

- **Derived extrinsic caps exceed block capacity** (#93453) — derived `max_extrinsic` caps subtract `base_block`.
- **Upgrade Weight Fails Drain** (#93455) — runtime-upgrade weight is actually drained from the block budget.

### Runtime / wormhole

- **Derivative accounts inflate wormhole soundness pool** (#93477) — first use of an `as_derivative` pseudonym reveals it to the soundness counter (subtracting its balance from `PotentialWormholeBalance`).
- **Transfer-proof weight misses wrapped/internal dispatch** (#93491) — `as_derivative`-wrapped transfers are statically counted; `post_dispatch` registers extra block weight for proof-recording work that static analysis cannot see (e.g. `Multisig::execute`, `recover_funds`).

### Acknowledged (documented, not changed)

- **Callbacks are upgrade-unstable** (#93370) — `Callback` stores `(pallet_index, call_index)` and resolves against the call layout at `curry` time. No in-repo consumer persists callbacks across upgrades; constraint documented for future oracle work.
- **Signer length underweights validation** (#93448) — `check_nonce` / `check_non_zero_sender` use constant weights. Sound when `AccountId` has a fixed `max_encoded_len()` (true for Quantus); documented so future variable-length account types regenerate weights.

---

## Earlier fixes on this branch (from `illuzen/frame-fixes`)

These predate the medium batch but are included when merging this branch to `main`:

- Validate `zk_tree_root` on block import (`frame-executive`)
- Cap genesis `dev_accounts` derivation (`pallet-balances`)
- Bound `RemovePallet` / `RemoveStorage` deletion; try-runtime tolerates intentional remainders
- Add `move_prefix_bounded` for staged prefix moves
- `Balanced::deposit` checks issuance headroom before crediting
- `CheckNonce` rejects `MAX` nonce instead of wrapping to zero
- `ensure_frozen` / `decrease_frozen` idempotency for permitted over-freezes
- `DecodeAll` for `Callback::curry` and NFT `typed_attribute`
- `Imbalance::ration` overflow handling; `SplitTwoWays` compile-time validation
- Transfer dust accounting (`fungibles::transfer` + `do_transfer_approved`)
- `dynamic_params` inner-attribute precedence

---

## Test plan

- [x] `cargo test -p frame-support --lib` (228 tests)
- [x] `cargo test -p quantus-runtime --lib` (36 tests, including wormhole extension lifecycle tests)
- [x] `cargo check --workspace --all-targets`
- [ ] Full CI on PR
- [ ] `cargo test --release` doctests (defensive trait docs fixed to `no_run`)

### Focused regression areas

- Counted storage: `try_append`, `migrate_key`, `set`, `clear_prefix`, iterator drains
- `unhashed::clear_prefix` cursor resumption
- Wormhole: `as_derivative` weight counting, post-dispatch weight reconciliation, derivative reveal
- Migration helpers: bounded deletion, overlapping prefix rejection, partial clear counter accounting

---

> [!NOTE]
> **High Risk**
> Touches block import (ZK header field), runtime upgrades/migrations, counted storage invariants, and fungible issuance/transfer/nonce paths—errors could brick sync, skew counters, or break accounting; breadth across executive, support, balances, assets, and frame-system warrants careful CI and runtime upgrade testing.
> 
> **Overview**
> Hardens FRAME and related pallets against a batch of **V12 medium-severity audit** issues: bounded decoding/enumeration, storage counter correctness, migration safety, and token/accounting edge cases, plus **ZK tree root** validation on block import.
> 
> **Block import & headers:** `frame-executive` re-derives and asserts `zk_tree_root` on import (with a regression test for forged roots). `qp-header` is wired in as a normal dependency.
> 
> **Codegen & metadata:** Unknown `get_call_names` modules return `[]` instead of panicking. Runtime/pallet **task** iterators are lazy. `#[pallet::weight]` and related attributes reject trailing tokens. New-pallet **storage version** is seeded after migrations, not in `before_all_runtime_migrations`. Genesis field matching uses serde-exact camelCase. `RuntimeMetadataPrefixed` checks `META_RESERVED` before decoding payloads. `WrapperKeepOpaque` enforces `MaxEncodedLen` on decode/construct. Counted **N-map** counter prefixes participate in duplicate-prefix checks. `dynamic_params` only injects outer args when the inner attribute has none.
> 
> **Storage & migrations:** Child `take` clears corrupt entries; `clear_storage` / `unhashed::clear_prefix` fix overlay accounting and cursor resumption. `move_prefix` is bounded via `move_prefix_bounded` and panics on overlapping prefixes. `RemovePallet` / `RemoveStorage` take a per-upgrade key **Limit**. Counted map/N-map: `try_append` respects `OnEmpty` defaults and key cardinality; `set`/`migrate_*`/partial `clear_prefix` and drain iterators keep counters in sync via `KeyPrefixIterator` **OnRemoval** hooks. `BTreeSet::append` deduplicates; bounded `try_append` uses effective absent length.
> 
> **Tokens & system:** `OnHold` transfers require an existing destination. `deposit` checks issuance headroom; expendable **transfer** credits reaped dust (assets `transfer_approved` uses `burn_dust`). Freeze/hold **consideration** rejects zero-footprint tickets; `ensure_frozen` / `decrease_frozen` idempotency fixes. `CheckNonce` rejects max nonce (no wrap). `max_extrinsic` validation uses derived caps. View functions decode with a **16 MiB** mem limit and full input consumption. NFT/callback paths use `DecodeAll` where prefix decode was unsafe. `MAX_DEV_ACCOUNTS` caps genesis dev derivation.
> 
> **Docs only (acknowledged):** `Callback` upgrade instability and constant-weight signer extensions are documented, not behavior-changed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8fc3a0195b2183c8bbe00eb3735af0a163424c2e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6815,6 +6815,7 @@ dependencies = [
  "pallet-timestamp",
  "parity-scale-codec",
  "qp-high-security",
+ "qp-wormhole",
  "scale-info",
  "sp-core",
  "sp-io",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -543,27 +543,11 @@ dependencies = [
 
 [[package]]
 name = "asn1-rs"
-version = "0.6.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5493c3bedbacf7fd7382c6346bbd66687d12bbaad3a89a2d2c303ee6cf20b048"
-dependencies = [
- "asn1-rs-derive 0.5.1",
- "asn1-rs-impl",
- "displaydoc",
- "nom 7.1.3",
- "num-traits",
- "rusticata-macros",
- "thiserror 1.0.69",
- "time",
-]
-
-[[package]]
-name = "asn1-rs"
 version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "56624a96882bb8c26d61312ae18cb45868e5a9992ea73c58e45c3101e56a1e60"
 dependencies = [
- "asn1-rs-derive 0.6.0",
+ "asn1-rs-derive",
  "asn1-rs-impl",
  "displaydoc",
  "nom 7.1.3",
@@ -571,18 +555,6 @@ dependencies = [
  "rusticata-macros",
  "thiserror 2.0.18",
  "time",
-]
-
-[[package]]
-name = "asn1-rs-derive"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "965c2d33e53cb6b267e148a4cb0760bc01f4904c1cd4bb4002a085bb016d1490"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.106",
- "synstructure 0.13.2",
 ]
 
 [[package]]
@@ -792,17 +764,6 @@ name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
-
-[[package]]
-name = "attohttpc"
-version = "0.24.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8d9a9bf8b79a749ee0b911b91b671cc2b6c670bdbc7e3dfd537576ddc94bb2a2"
-dependencies = [
- "http 0.2.12",
- "log",
- "url",
-]
 
 [[package]]
 name = "autocfg"
@@ -2337,25 +2298,11 @@ dependencies = [
 
 [[package]]
 name = "der-parser"
-version = "9.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5cd0a5c643689626bec213c4d8bd4d96acc8ffdb4ad4bb6bc16abf27d5f4b553"
-dependencies = [
- "asn1-rs 0.6.2",
- "displaydoc",
- "nom 7.1.3",
- "num-bigint",
- "num-traits",
- "rusticata-macros",
-]
-
-[[package]]
-name = "der-parser"
 version = "10.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "07da5016415d5a3c4dd39b11ed26f915f52fc4e0dc197d87908bc916e51bc1a6"
 dependencies = [
- "asn1-rs 0.7.1",
+ "asn1-rs",
  "displaydoc",
  "nom 7.1.3",
  "num-bigint",
@@ -2740,18 +2687,6 @@ name = "encode_unicode"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34aa73646ffb006b8f5147f3dc182bd4bcb190227ce861fc4a4844bf8e3cb2c0"
-
-[[package]]
-name = "enum-as-inner"
-version = "0.6.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1e6a265c649f3f5979b601d26f1d05ada116434c87741c9493cb56218f76cbc"
-dependencies = [
- "heck 0.5.0",
- "proc-macro2",
- "quote",
- "syn 2.0.106",
-]
 
 [[package]]
 name = "enum-display"
@@ -3512,17 +3447,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "futures-rustls"
-version = "0.26.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a8f2f12607f92c69b12ed746fabf9ca4f5c482cba46679c1a75b874ed7c26adb"
-dependencies = [
- "futures-io",
- "rustls",
- "rustls-pki-types",
-]
-
-[[package]]
 name = "futures-sink"
 version = "0.3.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3748,25 +3672,6 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.3.27"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0beca50380b1fc32983fc1cb4587bfa4bb9e78fc259aad4a0032d2080309222d"
-dependencies = [
- "bytes 1.11.1",
- "fnv",
- "futures-core",
- "futures-sink",
- "futures-util",
- "http 0.2.12",
- "indexmap",
- "slab",
- "tokio 1.47.1",
- "tokio-util",
- "tracing",
-]
-
-[[package]]
-name = "h2"
 version = "0.4.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f3c0b69cfcb4e1b9f1bf2f53f95f766e4661169728ec61cd3fe5a0166f2d1386"
@@ -3776,7 +3681,7 @@ dependencies = [
  "fnv",
  "futures-core",
  "futures-sink",
- "http 1.3.1",
+ "http",
  "indexmap",
  "slab",
  "tokio 1.47.1",
@@ -3937,37 +3842,12 @@ dependencies = [
  "futures-channel",
  "futures-io",
  "futures-util",
- "hickory-proto 0.26.1",
+ "hickory-proto",
  "idna",
  "ipnet",
  "jni 0.22.4",
  "rand 0.10.1",
  "thiserror 2.0.18",
- "tinyvec",
- "tokio 1.47.1",
- "tracing",
- "url",
-]
-
-[[package]]
-name = "hickory-proto"
-version = "0.24.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "92652067c9ce6f66ce53cc38d1169daa36e6e7eb7dd3b63b5103bd9d97117248"
-dependencies = [
- "async-trait",
- "cfg-if",
- "data-encoding",
- "enum-as-inner",
- "futures-channel",
- "futures-io",
- "futures-util",
- "idna",
- "ipnet",
- "once_cell",
- "rand 0.8.5",
- "socket2 0.5.10",
- "thiserror 1.0.69",
  "tinyvec",
  "tokio 1.47.1",
  "tracing",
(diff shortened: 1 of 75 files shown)


n13 (commented):
## Review — V12 medium-severity audit fixes

Went through all 67 files. Overall this is careful, well-tested work: each fix is narrowly targeted at the audited failure mode, and nearly every behavioral change ships with a regression test that would fail without the fix. Issues below, ordered by importance.

### Blocking

**1. CI is red — `cargo fmt` failures skip the entire test matrix.**
The Fast Checks (Format) job fails on `pallets/scheduler/src/tests.rs`, `pallets/wormhole/src/lib.rs`, and `runtime/src/transaction_extensions.rs`, so Build & Test and Clippy never ran — nothing in this PR has CI test coverage yet. Needs a fmt pass and re-push.

### Should fix / discuss

**2. `child::clear_storage` now walks the whole child trie twice** (`frame/support/src/storage/child.rs`).
The new `key_count` helper enumerates every key via `next_key` both before and after the kill to attribute overlay-only removals. A call with `maybe_limit = Some(small)` now does O(total keys) host calls — the same "user-growable state forces unbounded work" pattern this audit batch fixes elsewhere (e.g. `RemovePallet`). No in-repo callers today, so it's latent, but consider bounding the count or at least documenting the cost on the function.

**3. Wormhole weight model not updated for the new `KnownDerivatives` read** (`runtime/src/transaction_extensions.rs` / `runtime/src/configs/mod.rs`).
`NonWormholeAccounts::contains` now also reads `Utility::KnownDerivatives` (via `is_derivative`), so `is_ambiguous_account` costs one more read than the model assumes. Both `per_transfer_weight()` (4 reads) and the signer reveal weight (`reads_writes(4, 1)`) should become 5 reads, or the comments should say why the extra read is ignored.

### Minor / questions

4. **`KeyPrefixIterator` drain does an extra `get_raw` per key even when `OnRemoval = ()`** (`frame/support/src/storage/mod.rs`). Every plain `iter_keys().drain()` on a non-counted map now pays a value read per key that only exists to feed the counter hook. Fine for correctness; could be specialized away if it ever matters somewhere hot.
5. **New-pallet detection changed from "no keys under the pallet prefix" to "no storage-version key"** (`hooks.rs`). A pallet holding data that never wrote `:__STORAGE_VERSION__:` now gets its in-code version seeded after migrations, where before it was left untouched. I believe that's fine (arguably more correct) for this chain — just confirming it's intentional.
6. `BTreeSet` append is now decode-insert-rewrite, i.e. O(n) per append instead of an O(1) raw append. Correct fix for the dedup issue; worth remembering for large sets.
7. `RemovePallet`/`RemoveStorage` gained a mandatory `Limit` type param — breaking for downstream users of these helpers (none in-repo).

### Checked and found sound

- `zk_tree_root` re-derivation on import (both the `execute_block` state-root-gated path and `final_checks`), with the forged-root regression test.
- Storage-version seeding order: in `Executive::execute_on_runtime_upgrade`, custom migrations (`COnRuntimeUpgrade`) run before `AllPalletsWithSystem::on_runtime_upgrade`, so version-gated migrations targeting a new prefix correctly observe version 0, and the pallet hook doesn't clobber a version a migration set.
- Scheduler preimage ownership: the scheduler already drops the preimage after execution and on cancel, so dropping on failed placement (including the referenda v3-shim path, where the proposal is handed off at approval) is consistent and closes the leak without double-drops.
- Counted map/N-map counter reconciliation (`try_append`, `migrate_key(s)`, `set`, partial `clear_prefix`, all drain iterators), including the hasher-collapse cases.
- Fungible(s) `transfer` crediting the actual debit: matches assets' reap semantics (`decrease_balance` can return amount + dust there); for pallet-balances dust still routes through `handle_dust`, so no double count. `transfer_approved` burning reap dust instead of over-crediting the delegate's destination is the right call.
- `Balanced::deposit` issuance-headroom check mirrors `mint_into`; `Exact` fails cleanly, `BestEffort` caps at headroom, and the debt drop no longer silently saturates.
- `set_code` weight drain via `register_extra_weight_unchecked` — no double counting with `CheckWeight`'s refund since `actual_weight` is capped at the static weight.
- `unhashed::clear_prefix` cursor resumption, `move_prefix_bounded` cursor + disjointness assert (equal prefixes still early-return before the assert), nonce-MAX rejection, `max_extrinsic` base_block subtraction, view-function 16 MiB decode bound, serde-exact camelCase genesis retention, and trailing-token rejection in pallet attributes.
- The two acknowledged items (#93370 Callback upgrade-instability, #93448 constant-weight signer extensions) are reasonable as documentation-only given in-repo usage.

### Verdict

**Approve after fixes.** The substance is sound and test coverage is strong, but this shouldn't merge until (1) fmt is fixed so the full CI matrix actually runs green, and ideally (2) the wormhole read count is bumped for the `KnownDerivatives` lookup. Item 2 (`child::clear_storage` cost) can be a follow-up since the function currently has no callers, but a doc warning now would be cheap.

n13 (approved):
## Re-review

All feedback from the previous round has been addressed, and CI is now fully green (Format, Build & Test on Linux + macOS, Clippy & Doc). Verified each follow-up commit:

**1. fmt fixed** (`daa52b4f`, `3255146d`) — the Fast Checks job passes and the full test matrix actually ran this time.

**2. `child::clear_storage` counting is now bounded** (`f5179c48`) — both pre/post key walks cap at `limit + 1024`. I checked the saturation arithmetic: `counted_before − counted_after` is always ≤ the number of keys actually deleted (when `before` saturates and `after` doesn't, the difference is strictly less than the true deletion count), so overlay attribution can only *under*-report, never inflate `unique`/`loops` — exactly as the new doc comment claims. The `backend` count stays exact, and the 1,500-key regression test pins the degraded-to-zero behavior.

**3. Wormhole weight model updated** (`1d611e03`) — `per_transfer_weight` and the signer reveal weight both went from 4 to 5 reads to account for the `KnownDerivatives` lookup in `NonWormholeAccounts::contains`, with comments and tests updated consistently.

**4. `KeyPrefixIterator` drain read eliminated when unused** (`5a2981d3`) — the `NEEDS_VALUE: bool` associated const is a clean compile-time specialization: `()` and both `OnRemovalCounterUpdate` hooks opt out, so plain and counted key-drains skip the per-key value read, while the default (`true`) keeps the safe behavior for any future value-inspecting hook.

**Merge with main** (`612f33e0`) — this also shrank the PR from 67 to 47 files since the stacked `illuzen/frame-fixes` base (zk-root validation, bounded `RemovePallet`/`RemoveStorage`, transfer dust accounting, nonce-MAX rejection, NFT `DecodeAll`, etc.) has landed on main; what remains is the medium-severity batch itself, which I reviewed in full last round. The test duplicated by the merge was correctly removed in `893b1507` (the surviving copy on main is identical).

Remaining notes from last round (`BTreeSet` append now O(n), storage-version seeding keyed on the version key rather than pallet-prefix keys, `Limit` param being a breaking change for downstream `RemovePallet` users) are acknowledged trade-offs, not defects.

### Verdict

**Approve.** The fixes are minimal and targeted, every behavioral change has a regression test, all review feedback was resolved properly rather than papered over, and CI is green across the board. Good to merge.

PR #618, opened 2026-07-07
Wormhole Aggregation Level 2

illuzen:
## Summary

Teaches `pallet-wormhole` to verify second-layer ("public batch") aggregation proofs.
A public batch bundles M private-batch proofs into one transaction and names an
`aggregator_address`; as an incentive to aggregate, part of the fee that would
normally be burned is redirected to that address. Verification is segment-aware:
individual private-batch segments can be denied (replayed nullifiers) or skipped
(dummy padding) without rejecting the whole proof.

### New extrinsic

- `verify_public_batch(proof)` (unsigned, call index 3), alongside the renamed
  `verify_private_batch` (formerly `verify_aggregated_proof`).

### Segment-based exit bundles

- Validation and execution are refactored around `ExitBundle` / `ExitSegment`:
  both extrinsics parse their proof's public inputs into a normalized bundle where
  each segment corresponds to one private batch and is accepted or denied atomically.
- **Segment-level denial**: a segment whose nullifiers are already used (or collide
  with another segment in the same bundle) is denied; the rest of the bundle still
  executes. This also closes the double-mint that would otherwise arise from
  including the same private-batch proof in a public batch twice, since the circuit
  forwards nullifiers without deduplication.
- **Inert (dummy) segments**: dummy-padded segments (all-zero nullifiers) are
  skipped entirely — no mints, no fee accounting, no nullifier storage. Zero
  nullifiers are never written to `UsedNullifiers`.

### Aggregator incentive

- New config item `VolumeFeesAggregatorRate` (runtime: 50%). For public batches,
  that share of the **burn** portion of the 0.1% volume fee is paid to the proof's
  `aggregator_address` instead of being burned. The miner share is unchanged, so
  aggregation is strictly fee-positive for the aggregator and strictly
  burn-reducing for users.

### Transaction pool

- `validate_unsigned` handles both call variants; priority is computed from the
  total value of *valid, non-inert* segments only, so a mostly-replayed bundle
  can't jump the queue.

### Build / infra

- `build.rs` generates and embeds the public-batch verifier binaries and exposes
  circuit sizing constants (`NUM_LEAF_PROOFS`, `NUM_PRIVATE_BATCH_PROOFS`,
  overridable via `QP_NUM_PRIVATE_BATCH_PROOFS`) for public-input parsing.
- `[patch.crates-io]` temporarily points the `qp-wormhole-*` /
  `qp-zk-circuits-common` crates at the local `../qp-zk-circuits` checkout.
  **Must be switched to published crate versions before merge.**
- Placeholder weight for `verify_public_batch` (25s ref-time upper bound);
  real benchmarking to follow.

## Test plan

- `cargo test -p pallet-wormhole --release`: new `public_batch_proof_tests` cover
  PI parsing, wrong origin, malformed bytes, replay prevention, and a success path
  asserting the aggregator rebate is paid and dummy segments are skipped.
- Fixtures regenerated (`test-data/private_batch.hex`, `test-data/public_batch.hex`)
  via the `regenerate_*_fixture` ignored tests.
- End-to-end against `--dev` node via `quantus-cli wormhole multiround --public`.

---

> [!NOTE]
> **High Risk**
> Changes ZK proof verification, minting, fee burn/miner/aggregator splits, and nullifier handling on a security-critical pallet; segment denial and temporary local circuit patches add integration and supply-accounting risk until published crates and real benchmarks land.
> 
> **Overview**
> Adds **second-layer wormhole aggregation** on-chain: **`verify_public_batch`** (call index 3) alongside **`verify_private_batch`**, which replaces **`verify_aggregated_proof`**. Naming and ZK crates move from layer-0/layer-1 to **private batch / public batch**, with **`[patch.crates-io]`** temporarily pointing `qp-wormhole-*` at a local circuits checkout until **3.1.0** ships.
> 
> **`build.rs`** now generates and embeds **public-batch** verifier binaries, passes **`QP_NUM_PRIVATE_BATCH_PROOFS`** (default 4), and emits **`NUM_LEAF_PROOFS` / `NUM_PRIVATE_BATCH_PROOFS`** for PI parsing at runtime.
> 
> Validation and execution are refactored around **`ExitBundle` / `ExitSegment`**: both extrinsics normalize proofs into per–private-batch segments. **Segment-level denial** skips whole segments when nullifiers are already used or collide within the bundle (partial success for public batches); **inert** dummy segments are ignored. **`SegmentsDenied`** is emitted when some segments are dropped. Pool **`validate_unsigned`** priority counts only valid, non-inert segment value.
> 
> **`VolumeFeesAggregatorRate`** redirects part of the volume-fee **burn** bucket to the proof’s **`aggregator_address`** on public batches (miner share unchanged). Errors and weights are generalized (`VerifierNotAvailable`, `verify_public_batch` placeholder ~25s ref-time). Tests and fixtures move to **`private_batch.hex`** / **`public_batch.hex`** with new public-batch coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7b043f722cd3fe0a6d5c04b0da755cbbd42ab8a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (commented):
## Review: Wormhole Aggregation Level 2

**Verdict: Solid design and implementation — approve after closing one test gap and confirming the release/versioning items below.** The segment-based refactor is well thought out (denial granularity is correctly justified by the circuit's exit-summing), validity is recomputed at dispatch, dispatch is transactional so partial state can't leak, and the PI parser strictly validates lengths so the segment slicing in `ExitBundle::from_public_batch` can't panic. Nice supporting hardening too (`derive_wormhole_address` canonicality, documented lossy encodings in `zk-tree`/header).

### Should fix before merge

**1. The headline feature — partial segment denial — has no test coverage.**
The public-batch fixture contains exactly one real segment, so no test exercises:
- a bundle where one segment is denied and the rest still execute (`SegmentsDenied` event path, `denied_segments` accounting),
- the cross-segment `claimed`-set dedup in `segment_validity` (the stated double-mint fix for including the same private batch twice in one bundle),
- fee/rebate math when `total_exit_amount` excludes a denied segment's value.

`test_verify_public_batch_cannot_replay` only covers the all-denied → reject path. Given this is the mechanism that prevents double-minting on a mint-authorized pallet, a two-real-segment fixture (or a unit test driving `segment_validity`/`process_exit_bundle` directly with synthetic bundles) seems worth it before this ships.

**2. Batch verifier loading bypasses the audit-era artifact checks** (`pallets/wormhole/src/lib.rs`, `load_batch_verifier_from_bytes`).
`WormholeVerifier::new_from_bytes` (hardened in the #616-era audit work) pins artifacts to keccak256 commitments and runs `ensure_loaded_matches_canonical_leaf_profile` (config equality + `MIN_LEAF_SECURITY_BITS`). The new loader constructs `VerifierCircuitData` from raw deserialization with no integrity or config/security-bits check at all. The bytes are build-time-generated from the pinned crate, so the trust story is mostly intact, but the cheap defense-in-depth (at minimum: `common.config == standard_recursion_config()` / security-bits floor) should be retained — ideally as a `new_batch_from_bytes` in `qp-wormhole-verifier` so the check lives next to the pinned leaf path rather than being reimplemented (and weakened) in the pallet.

**3. `spec_version` is unchanged (134 on both main and this branch)** despite a new call, new event, and renamed errors/calls in metadata. If the release automation bumps it, fine — just confirm this doesn't ship as a runtime upgrade under the same version.

### Worth noting / smaller items

- **Aggregator rebate can revert the whole bundle.** `increase_balance(…, Precision::Exact)` for the aggregator fails if the account doesn't exist and the rebate is below ED — the entire public batch (users' exits included) then fails at dispatch, *after* passing full ZK verification in `validate_unsigned`, with `Pays::No`. It's self-inflicted by the aggregator, but it drags user exits down with it and consumes block space for free. Consider requiring/documenting a pre-existing aggregator account, or tolerating the dust (e.g. fall back to burning the rebate on failure).
- **Weight placeholder units:** the PR body says "25s ref-time"; `25_000_000_000` picoseconds is 25 **ms** (vs ~16.7 ms for the private batch). Presumably the value is what's intended and the body has a typo — worth double-checking against the real benchmark when it lands, since a wrapper proof over 4 inner proofs may not fit in 25 ms. Also, the public-batch early-failure path returns `pre_validate_proof()` weight, which was calibrated for private-batch pre-validation (33 reads); public pre-validation reads up to 64 nullifier keys.
- **PR body is stale on `[patch.crates-io]`:** the diff no longer contains the local-path patch — all `qp-*` deps point at published 3.1.0/1.5.4 crates. So the "must be switched before merge" blocker appears already resolved; please confirm that was intentional and the published 3.1.0 matches the checkout the fixtures were generated from.
- **bip39 swap (parity-bip39 → upstream 2.2.2):** chain-side usage is only `Mnemonic::generate`/`words()` in `client/cli/generate.rs`, so this looks safe, but worth a one-time sanity check that no mnemonic→key path in the wallet stack depended on parity-fork behavior.
- Nit: in `validate_unsigned` for `verify_public_batch`, the `!segment_is_inert(segment)` filter is redundant — inert segments already get `valid = false` from `segment_validity`.
- Nit: an all-dummy bundle is rejected with `NullifierAlreadyUsed`, which is misleading (the doc comment on the error acknowledges this; a dedicated `NoValidSegments`-style variant would read better in explorers).

### Things I checked that are fine

- Cross-segment nullifier collision handling is sound: first valid segment claims, later colliders are denied; a nullifier already on-chain invalidates every segment carrying it. No double-mint path found.
- Zero nullifiers are never written to `UsedNullifiers`, in both validity and processing passes.
- Dispatch errors are transactional, so nullifier inserts/mints roll back on late failures (min-transfer, soundness invariant, decode errors).
- The aggregator rebate stays inside the burn bucket (miner share and the supply-accounting comments remain consistent with the existing `increase_balance`/`set_total_issuance` scheme).
- Pool priority now counts only valid, non-inert segment value — a mostly-replayed bundle can't jump the queue.
- Weight `storage_tail` scaling and its unit test are consistent with the `NUM_LEAF_PROOFS`/`NUM_PRIVATE_BATCH_PROOFS` build constants.

CI note: Clippy/format pass; the Linux and macOS build+test matrix jobs were still pending at review time.

n13 (approved):
## Re-review (commits `4702b6c8..640cd6ec`)

**Verdict: Approve.** All substantive findings from my previous review are addressed, CI is fully green (both build+test matrices included), and the new code holds up under scrutiny.

### Findings from last review — resolution status

**1. Partial segment denial untested → resolved (`4702b6c8`).**
The new `exit_bundle_tests` module drives `segment_validity` / `process_exit_bundle` directly with synthetic multi-segment bundles and covers exactly the gaps called out: partial denial with `SegmentsDenied` + correct per-segment nullifier accounting (denied segment's fresh nullifiers stay unconsumed — important, verified), the cross-segment claimed-set dedup including the subtle "denied segment claims nothing" case (`[1,2],[2,3],[3]` → `[true,false,true]`), the same-private-batch-twice double-mint scenario, zero-nullifier exemption, and fee/rebate math computed on valid segments only (with a guard assert that the test would actually distinguish the two totals — nice touch).

**2. Verifier loading bypassed audit-era checks → resolved (`b4270ac3`).**
`ensure_batch_verifier_profile` restores the defense-in-depth: config equality against the canonical batch configs, the `MIN_LEAF_SECURITY_BITS` floor, and a PI-count check against the compiled batch dimensions (which the leaf path didn't even have). The config duplication forced by `qp-zk-circuits-common`'s std constraint is properly mitigated by the `batch_configs_match_circuit_crate` parity test against the source of truth, and `canonical_configs_meet_security_floor` guards against a future upstream config regression. This is a satisfying resolution given the keccak-pin approach genuinely can't work for size-parameterized artifacts.

**3. Below-ED aggregator rebate reverting the bundle → resolved (`e34042b2`).**
The rebate mint now falls back to burning on failure instead of dragging users' exits down, with a test that raises the ED at runtime (mock ED made `static` for this) and asserts both the successful user exit and the exact issuance delta.

**4. Smaller items → resolved.** `NoValidSegments` distinguishes all-dummy bundles from replays (tested); the redundant inert filter in `validate_unsigned` is gone with an explanatory comment; the public-batch early-failure path now gets its own `pre_validate_public_batch_proof` weight scaled to all 64 nullifier reads.

### Still open (non-blocking, mostly acknowledged)

- **`spec_version` is still 134.** Unchanged from main despite new calls/events/errors (and the new `NoValidSegments` variant shifts error indices further). If release automation bumps it, fine — flagging one last time since it's the kind of thing that's silent until an upgrade goes out.
- **`verify_public_batch` real benchmark** — the 25 ms placeholder ref-time and the borrowed 0.685 ms deserialize cost in `pre_validate_public_batch_proof` (public-batch proofs are larger to deserialize) both await the dedicated benchmark, as the code comments acknowledge.
- **PR body is stale**: it still says `[patch.crates-io]` points at a local checkout and "must be switched before merge" — the diff has used published 3.1.0 crates since `5ca0fe96`. Worth editing the description so the merge record doesn't document a blocker that no longer exists.

Nothing new to flag in the delta itself. Nice turnaround on the fixes.

PR #619, opened 2026-07-21
Emit MinerVolumeFeePaid event for wormhole exit volume fees

n13:
## Problem

On `Wormhole.verify_aggregated_proof`, the block author's share of the exit volume fee is minted via `increase_balance` **without** `record_transfer`. Unlike exit payouts (`NativeTransferred`) and mining rewards (`MinerRewarded` + `NativeTransferred`), no wormhole-pallet event carries this amount — indexers only see a bare `Balances.Minted`, which is ambiguous (exit payouts and mining rewards mint through the same event) and therefore intentionally ignored by our subsquid indexer. Result: indexed miner balances drift below RPC state on every wormhole exit (observed gap: 330980981066 planck on one deposit-only miner account).

## Change

New event on `pallet-wormhole`:

```rust
MinerVolumeFeePaid { miner: AccountId, amount: Balance }
```

- Emitted immediately after the `increase_balance` mint in `verify_aggregated_proof`.
- **Appended last** in the `Event` enum, so existing variant indices (`NativeTransferred` = 0, `AssetTransferred` = 1, `ProofVerified` = 2) are unchanged for indexers decoding by metadata.
- The no-author fallback path (fee added to the burn) emits nothing — nothing is minted there.

## Indexer impact

Companion subsquid PR consumes this event and credits `miner.free` directly. Both runtime generations remain indexable: pre-upgrade blocks carry only the ambiguous `Minted` (handled by the heuristic in quantus_subsquid#92), post-upgrade blocks carry this explicit event. `Balances.Minted` emission from `increase_balance` is unchanged, so indexers on the #92 heuristic alone also keep working after the upgrade.

## Tests

- `test_verify_aggregated_proof_emits_miner_volume_fee_paid`: seeds a block author via the pre-runtime digest (QPoW path), verifies a real aggregated proof, asserts the event fires with the exact expected fee (`exit * 10 / 9990`, minus 50% burn) and the author's balance increases by the same amount.
- Existing success test now asserts **no** event is emitted when the digest has no author (fee burned instead).
- `cargo test -p pallet-wormhole`: 28/28 pass. `cargo check -p quantus-runtime`: clean.

No `spec_version` bump — left to the release automation.


pallets/wormhole/Cargo.toml
```diff
@@ -43,6 +43,7 @@ qp-wormhole-inputs.workspace = true
 qp-wormhole-prover.workspace = true
 qp-wormhole-verifier = { workspace = true, default-features = false }
 qp-zk-circuits-common = { workspace = true, default-features = false }
+sp-consensus-qpow = { workspace = true, features = ["std"] }
```
illuzen: This seems wrong, pallets are no-std, why do we need this new dependency?

n13: Good catch — removed in f0c728b6. The dep was only in dev-dependencies (mock.rs needed `POW_ENGINE_ID` for the pre-runtime digest), but it's unnecessary either way: `qp-wormhole` already depends on `sp-consensus-qpow` for `extract_author_from_digest`, so it now re-exports `POW_ENGINE_ID` and the mock uses that.

illuzen (approved):

PR #630, opened 2026-07-24
V12 review node

illuzen:
# Security review fixes

Addresses eight items from the security review. Each item was first reproduced with a failing (red) test where applicable, then fixed, with the test suite green afterwards.

### 1. `txWatch` docs: zero-conf mempool signal, not payment confirmation (`36b5416`)

`txWatch_transfer` notifications are decoded from mempool call data before execution, so they can fire for transfers that never execute (dropped/replaced transactions, failed dispatch, `batch`/`batch_all`/`force_batch` semantics — all attacker-reachable for a transaction fee). This is intentional: the subscription is a zero-confirmation UX signal. The integrator docs now spell out exactly what a notification does and does not mean, and require on-chain verification (finalized inclusion + `Balances.Transfer` / `Assets.Transferred` event) before releasing goods or marking an invoice paid.

**Docs only, no code change.**

### 2. Miner port exposure guidance (`59e2a37`)

The external-miner listen port (`--miner-listen-port`) is unauthenticated: anyone who can reach it can consume mining work or submit results. Documentation now states the port must never be exposed to the public internet and gives concrete network-isolation and firewall guidance.

**Docs only, no code change.**

### 3. `recover_funds` weight undercharging (`27d54f8`)

`pallet_reversible_transfers::recover_funds` iterated all pending transfers of an account (storage reads, release attempts, event emission) but charged weight only for the *successfully cancelled* ones. An attacker could set up many failing releases and make the chain do significant work below the charged weight. The extrinsic now charges `recover_funds(num_processed)` — the number of transfers actually iterated.

### 4. Governance censorship via shared scheduler (`c90566c`)

A single `pallet_scheduler` instance (`MaxScheduledPerBlock = 50`) is shared between referenda enactment and `pallet_reversible_transfers`, whose scheduling surface is permissionless with attacker-chosen target blocks. An attacker could cheaply pre-fill a referendum's deterministic enactment block; the scheduler's `Exhausted` error is only logged in release builds, so an approved referendum would silently fail to enact.

Fix: `push_to_agenda` caps `LOWEST_PRIORITY` tasks at ~80% of `MaxScheduledPerBlock`, reserving headroom for higher-priority tasks, and reversible transfers are now scheduled at `LOWEST_PRIORITY`. Governance enactment (scheduled at higher priority) therefore always finds a slot.

### 5. `txWatch_watchAddress` denial of service (`3509f34`)

Subscriptions were admitted against a node-wide pool of 32 slots — a shared resource one unauthenticated client could exhaust to deny `txWatch` to everyone else. Built-in Substrate subscriptions have no per-method admission control; they are bounded by the server-level `--rpc-max-connections` / `--rpc-max-subscriptions-per-connection` limits. `txWatch` now follows the same model: the bespoke listener pool (and its error codes 5010/5012) is removed, so there is no `txWatch`-specific resource to exhaust.

### 6. Genesis tech-collective seeding panics (`6daf80a`)

`seed_tech_collective` ran the fallible `do_add_member_to_rank` with `.expect(...)` on the caller-supplied `tech_collective_seed_members` list, so duplicate members, overlap with existing members, or over-limit lists turned `build_state` into a WASM trap. Failures now propagate as `Err(String)` through the existing `sp_genesis_builder::Result` channel, naming the offending account and the dispatch error.

### 7. Key-generation secrets on the command line (`d405eea`)

`quantus-node key quantus` accepted the master seed / BIP39 mnemonic as argv values (world-readable via `/proc/<pid>/cmdline`, recorded in shell history, captured by audit/orchestration logs), echoed a caller-supplied mnemonic back to stdout, and printed the derived seed and secret key under `--verbose`.

- `--seed` / `--words` are now bare flags: the secret is read from stdin (no-echo prompt on a terminal, one piped line otherwise, e.g. `--words < mnemonic.txt`).
- Caller-supplied secrets are never echoed back; only newly generated mnemonics are printed (they would otherwise be lost).
- `--verbose` prints public key material only.

**Breaking CLI change:** invocations passing `--seed <value>` / `--words <value>` now fail with a usage error.

### 8. `zkTree_getMerkleProof` deep-archive reads (`9e05c5d`)

The node forces `ArchiveCanonical` state retention, and `at_block` was forwarded unvalidated, so an unauthenticated caller could drive sustained cold trie reads across the entire chain history. However, a wormhole spend proof is only accepted on-chain while the proving block's hash is still in `frame_system::BlockHash` (`BlockHashCount` = 4096 blocks) — older proofs are rejected with `BlockNotFound`, so deep-history queries have no legitimate use.

`at_block` is now validated: unknown hashes return error 9004, blocks more than `BlockHashCount` behind best return error 9005. The window is read from the runtime config, so it cannot drift. Within-window queries keep the platform-standard posture (server-level `--rpc-rate-limit` / `--rpc-max-connections` for public exposure).

### 9. Heisenberg well-known keys are intentional (docs)

Heisenberg genesis endows / authorizes the public Dilithium accounts (`crystal_alice` / `dilithium_bob` / `crystal_charlie`). That looks alarming if Heisenberg is mistaken for mainnet, but it is the **internal integration testnet**: tokens have no monetary value, keys are public so CI and integrators can exercise governance/treasury/transfers without secret distribution, and Planck (the public testnet) already uses distinct live treasury signers. Documented in `node/src/chain_spec.rs`, `runtime/src/genesis_config_presets.rs`, `primitives/dilithium-crypto/src/pair.rs`, and `docs/RUNTIME_SURFACE.md`.

**Docs only, no code change.**

## API / compatibility notes

- **`key quantus`**: `--seed` / `--words` no longer take a value (see item 7).
- **`txWatch_watchAddress`**: error codes 5010 and 5012 no longer exist; capacity is governed by server-level subscription limits.
- **`zkTree_getMerkleProof`**: new error codes 9004 (unknown block hash) and 9005 (block outside the 4096-block proof window).
- **Runtime**: `pallet_scheduler` reserves ~20% of `MaxScheduledPerBlock` for above-lowest-priority tasks; reversible transfers schedule at `LOWEST_PRIORITY`; `build_state` returns structured errors for invalid `tech_collective_seed_members`.

## Test plan

- [x] New unit/integration tests for each code fix (red before, green after):
  - `recover_funds_weight_accounts_for_failed_releases` (reversible-transfers)
  - `lowest_priority_tasks_cannot_crowd_out_high_priority_tasks` (scheduler) and `scheduled_transfers_cannot_crowd_out_high_priority_tasks` (reversible-transfers)
  - `seed_tech_collective_rejects_duplicates_without_panicking` (runtime)
  - `secrets_are_not_accepted_as_command_line_values`, `provided_mnemonic_is_not_echoed_back` (node)
  - `rejects_blocks_older_than_the_proof_window`, `rejects_unknown_block_hashes`, window-boundary acceptance (node)
- [x] Full test suites green: `pallet-scheduler`, `pallet-reversible-transfers`, `pallet-referenda`, `quantus-runtime` (24 passed), `quantus-node` (32 passed)
- [x] Manual smoke test of `key quantus` with piped stdin (known mnemonic produces the expected address; argv secret rejected)

---

> [!NOTE]
> **High Risk**
> Changes touch governance scheduling censorship, fee/weight accounting, secret handling CLI breakage, and unauthenticated RPC resource limits—areas where mistakes affect liveness, funds recovery, or operator security.
> 
> **Overview**
> This PR bundles **security-review fixes** across runtime, node, and documentation.
> 
> **Governance / scheduler:** `pallet_scheduler` now limits `LOWEST_PRIORITY` tasks to ~80% of `MaxScheduledPerBlock`, and `pallet_reversible_transfers` schedules delayed executes at `LOWEST_PRIORITY`, so permissionless scheduling cannot fill a block’s agenda and block referendum enactment at priority 63.
> 
> **Economics / DoS:** `recover_funds` charges weight for every pending transfer processed (`num_processed`), not only successful cancellations, closing undercharge when releases fail.
> 
> **Node — keys (breaking):** `quantus-node key quantus` reads `--seed` / `--words` from stdin via `rpassword` (no inline argv secrets); caller-supplied mnemonics are not echoed; `--verbose` omits seed/secret key output. Tests enforce the new CLI shape.
> 
> **Node — RPC:** `txWatch` drops the global 32-listener cap (errors 5010/5012 removed); capacity follows standard `--rpc-max-*` limits. `zkTree_getMerkleProof` validates `at_block` against `BlockHashCount` (9004 unknown hash, 9005 outside window) to block deep archive trie abuse.
> 
> **Genesis:** `seed_tech_collective` returns `Result<(), String>` instead of panicking on duplicate/invalid members.
> 
> **Docs:** Stronger guidance on unauthenticated miner QUIC (9833), `txWatch` as zero-conf mempool signal (not payment confirmation), Heisenberg’s intentional public test keys, and scheduler priority behavior in `RUNTIME_SURFACE.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3caaf99a8b3c648566870353effaa541814fcaba. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13:
## Review

I checked out the branch and re-ran the suites you list. All green locally, matching your test plan:

| suite | result |
|---|---|
| `pallet-scheduler --lib` | 80 passed |
| `pallet-reversible-transfers --lib` | 51 passed |
| `quantus-node --bins` | 32 passed |
| `quantus-runtime --test mod` (governance) | 24 passed, 1 ignored |

The `cargo fmt --check` diff I see is confined to `client/cli/src/arg_enums.rs` and is pre-existing (needs nightly rustfmt), not from this PR.

### What I verified rather than took on trust

- **`recover_funds` weight (#3)** is sound. `num_processed ≤ MaxPendingPerAccount = 16`, which is exactly the pre-dispatch declaration at `pallets/reversible-transfers/src/lib.rs:549`, so the refund can only ever move downward — no risk of the post-weight exceeding the declared weight. The `assert_eq!(T::MaxPendingPerAccount::get(), 16, ...)` guard in the benchmark keeps `Linear<0, 16>` honest if the config changes. Good.
- **Scheduler headroom (#4)** covers the whole surface. I grepped every `T::Scheduler::schedule*` call site in the workspace: only `pallets/referenda` (63 for enactment, 128 for alarms) and `pallets/reversible-transfers` (now 255). `ScheduleOrigin = EnsureRoot` (`runtime/src/configs/mod.rs:322`) means the caller-chosen-priority extrinsics can't be used to inject arbitrary priorities, and the calls are disabled at the runtime level anyway. So the reservation genuinely does bind the attacker path.
  Also confirmed the new early `Err((Exhausted, what))` return flows into the existing `T::Preimages::drop(&task.call)` cleanup in both `do_schedule` and `do_schedule_named` — no orphaned system-requested preimage on the new rejection path. And the `occupied` scan is over an agenda already in memory, so no extra storage read / PoV cost.
- **Genesis panic (#6)**: `seed_tech_collective` really was the only caller-input-reachable panic in that module. The remaining `.expect()`s are on serialization of internally-constructed values, and `account_from_ss58` (line 428) is only ever called on hardcoded literals.
- **zkTree window (#8)** errs in the safe direction. The RPC compares against `best_number`, while on-chain verification happens at some later block `m ≥ best`, where `frame_system::BlockHash` retains `[m - 4096, m-1]`. So the RPC is always at least as permissive as the chain and can never refuse a proof that would have verified.
- **CLI (#7)**: the new `--seed` help text (128 hex chars / 64 bytes) matches what `generate_quantus_key` actually validates at `node/src/command.rs:121`; the old help text said 64 characters and was simply wrong. Nice incidental fix. In-repo grep finds no remaining `--seed <value>` / `--words <value>` call sites, and the README is updated — but downstream wallet/CLI repos and any operator runbooks will need the same treatment.
- The `RUNTIME_SURFACE.md` treasury correction (30% of genesis supply → no pre-mine) is accurate against `genesis_config_presets.rs:177`. Worth calling out since it's a real doc-vs-code drift fix riding along.

---

### Things I'd like changed

**1. `txWatch`: replacing a shared cap with no cap is a wider swing than the finding needed (#5).**

The actual defect was that the 32 slots were a *shared* pool, so one client could starve everyone. A **per-connection** cap fixes exactly that — a client's budget is its own, so it can't starve others — without giving up the bound. Removing admission control entirely puts capacity at `--rpc-max-connections` (default 100) × `--rpc-max-subscriptions-per-connection` (default 1024) ≈ 100k concurrent subscriptions, each a spawned task that wakes on *every* transfer-bearing pool transaction. The per-message cost is now O(subscriptions × txs) and it's unauthenticated-client-driven. The `state_subscribeStorage` comparison is fair in spirit, but that method's fan-out is filtered server-side per subscription; here every listener wakes for every decoded tx and then linearly scans `decoded.transfers`.

I'd reinstate something like `MAX_LISTENERS_PER_CONNECTION` rather than nothing. If you'd rather not, that's a defensible call — but it should be a deliberate one, not a side effect of deleting the old counter.

**2. `txWatch`: the two INFO logs are now unbounded and unauthenticated-client-driven.** (Small, but I'd fix it in this PR.)

- `node/src/txwatch.rs:147` — `log::info!("Watching address …")`, once per subscription. Subscription churn is now free and uncapped, so this is a log-volume/disk amplifier.
- `node/src/txwatch.rs:174` — `log::info!("Transfer detected …")`, once per *matching listener per transaction*. N listeners on one hot merchant address means N INFO lines per tx.

Both were implicitly bounded by the 32-listener cap and no longer are. `debug!`/`trace!` is the right level for both.

**3. `BlockHashCount` is read from the node's linked runtime crate, not from the chain.**

`node/src/zktree_rpc.rs:86` uses `quantus_runtime::configs::BlockHashCount`, a compile-time constant baked into the node binary. The PR description says "The window is read from the runtime config, so it cannot drift" — it can, and forkless upgrades are precisely the mechanism: a runtime upgrade that changes `BlockHashCount` leaves every already-running node using the old value until operators rebuild. The blast radius is small (a DoS guard becomes slightly wrong in one direction, and the safe direction at that), so I'm not asking you to plumb the metadata lookup. But please either do that or replace the "cannot drift" comment with one that states the compile-time coupling, so the next reader doesn't rely on a guarantee that isn't there.

**4. Stale reference in a test doc comment.**

`pallets/reversible-transfers/src/tests/test_reversible_transfers.rs:2218` cites "the scheduler's reserved high-priority headroom (`ReservedHighPrioritySlots`)". No such item exists anywhere in the tree — the reservation is a hardcoded `max / 5` at `pallets/scheduler/src/lib.rs:728`.

Which raises the better question: should it be a real `Config` item? Two reasons beyond the comment. It lets runtimes tune the split, and it removes a silent failure mode — with `MaxScheduledPerBlock < 5`, `max / 5 == 0`, so `cap == max` and the reservation quietly does nothing. Your mock (`ConstU32<10>`) and the runtime (50) are both fine today, but nothing catches a future config that isn't.

---

### Smaller notes

**5. The `recover_funds` benchmark never exercises the asset path.** The new comment at `pallets/reversible-transfers/src/benchmarking.rs:285` says the all-success setup is "the per-transfer worst case", but the benchmark only creates native transfers via `do_schedule_transfer`, while `recover_funds` also releases *asset* holds through `AssetsHolder::release` — touching `pallet_assets` `Account`/`Asset` storage, which your own new test exercises. The success-vs-failure half of the claim holds and is well argued; the native-vs-asset half isn't established by the benchmark. Pre-existing gap, but the comment now asserts more than the benchmark proves. Either add an asset variant or narrow the wording.

**6. The reservation counts all occupants, not just lowest-priority ones.** `push_to_agenda` refuses a `LOWEST_PRIORITY` task once `occupied >= 40`, even when all 40 occupants are high-priority and 10 slots sit free. Legitimate reversible transfers get refused for a block that has room. It errs safe and I wouldn't change the behaviour, but a one-line comment saying it's deliberate would stop someone "fixing" it later into counting only lowest-priority slots.

**7. `client.number()` backend errors are reported as a bad request.** `node/src/zktree_rpc.rs:78` — `.ok().flatten()` collapses a genuine DB/backend error into error 9004 "unknown block hash". Splitting the error case out would let operators tell a corrupted DB from a client sending garbage.

**8. Secrets still live in non-zeroized `String`s.** `read_secret_from_stdin` returns `String`, and `value.trim().to_string()` leaves the untrimmed copy in freed heap. The argv/history/audit-log exposure — the actual finding — is fixed, and this repo already has `SensitiveBytes32` for the next increment. Not a blocker.

---

### On the description's framing of #4

> Governance enactment (scheduled at higher priority) therefore always finds a slot.

This is stronger than what the change buys. The reservation holds back priority-255 tasks only. Referendum *alarms* are priority 128, and an attacker willing to lock 10 × `ReferendumSubmissionDeposit` (100 UNIT, `runtime/src/configs/mod.rs:232`) can create tasks that legitimately occupy the reserved 10 slots. Meanwhile `schedule_enactment` still only `log::error!`s on `Exhausted` (`pallets/referenda/src/lib.rs:935`), so "approved referendum silently fails to enact" is *mitigated*, not eliminated.

I think that's the right scope for this PR — you closed the cheap, permissionless, attacker-aimed vector, and the residual costs real money and can't be aimed precisely. But it's worth a sentence in the description, and worth opening a follow-up on whether `schedule_enactment` should retry at `when + 1` on `Exhausted` instead of only logging. Belt and braces for the case that actually matters.

Also, item 9 claims a doc change in `primitives/dilithium-crypto/src/pair.rs`, but that file isn't in the diff — either it got dropped or the description should.

---

## Verdict

**Approve, with request #2 (log levels) before merge and a decision on #1 (txWatch cap).**

Eight findings, each aimed at the actual root cause rather than the symptom, each with a test that fails before and passes after — I re-ran them and they hold. The reasoning in the commit messages and code comments is unusually good; the `recover_funds` benchmark comment explaining *why* the all-success path is the worst case, and warning against changing it, is exactly the kind of note that stops a future regression.

The one item I'd genuinely push back on is #5. Going from "a shared cap that one client can exhaust" to "no cap at all" fixes the reported finding by removing the resource rather than bounding it, and the two unbounded INFO logs are a direct consequence. A per-connection cap gets you both properties. Everything else on this list is a comment, a doc sentence, or a follow-up ticket.

n13 (approved):

PR #631, opened 2026-07-28
Security hardening for wormhole / ZK-tree metering and commitment paths, plus related leaf-keying and deposit-tracking fixes.

illuzen:
## Summary

Security hardening for wormhole / ZK-tree metering and commitment paths, plus related leaf-keying and deposit-tracking fixes.

- **Canonical leaf recipients** — key ZK-tree leaves and transfer counts on the Goldilocks-canonical form of the recipient so non-canonical aliases cannot share a nullifier / leave a deposit permanently unexitable.
- **Native-only potential-balance credits** — do not treat `pallet_assets` asset-0 credits as native wormhole deposits for the soundness pool.
- **Private-batch prevalidation weight** — derive `pre_validate_proof` DB reads / PoV from `circuit_config::NUM_LEAF_PROOFS` (and size the benchmark the same way) instead of a hardcoded 32-nullifier assumption that can drift from `QP_NUM_LEAF_PROOFS`.
- **Depth-aware ZK-tree insert metering** — charge `per_transfer_weight` and wormhole exit `storage_tail` using the live tree depth (`insert_leaf` / `update_path` scale with depth), so declared weight tracks real storage work as the tree deepens.
- **Saturating leaf encoding** — commit quantized amount and asset_id as saturating `u32`s instead of truncating wraps, so oversized (e.g. permissionlessly minted) asset transfers cannot alias small amounts in the Merkle leaf.
- **Unsigned exit DoS gate** — move full plonky2 verification off `validate_unsigned` (pool admission) onto `pre_dispatch` (block inclusion); keep pool path at cheap pre-validation; dedup by nullifier set instead of raw proof bytes; stop re-verifying in the dispatch body so charged weight matches actual work.

## Test plan

- [ ] `cargo test -p pallet-zk-tree`
- [ ] `cargo test -p pallet-wormhole`
- [ ] `cargo test -p quantus-runtime transaction_extensions`
- [ ] Confirm new wormhole tests:
  - [ ] `pre_dispatch_rejects_unverifiable_proof`
  - [ ] `validate_unsigned_skips_verify_but_pre_dispatch_enforces_it`
  - [ ] `validate_unsigned_and_pre_dispatch_accept_valid_proof_with_semantic_tag`
- [ ] Confirm zk-tree amount-saturation tests:
  - [ ] `hash_leaf_oversized_amount_does_not_alias_small_amount`
  - [ ] `hash_leaf_saturates_oversized_amounts_at_u32_max`
- [ ] Confirm weights / extension depth tests (wormhole `weights::tests::*`, runtime `per_transfer_weight_scales_with_tree_depth`)
- [ ] Manual / review: unsigned exit with valid PIs but corrupted proof data is pool-admissible but rejected at `pre_dispatch` / block import
- [ ] Manual / review: regenerating wormhole benchmarks after changing `QP_NUM_LEAF_PROOFS` still aligns with the derived nullifier / depth logic

---

> [!NOTE]
> **High Risk**
> Changes wormhole soundness accounting, nullifier/leaf commitment semantics, and where unsigned proofs are cryptographically verified—any bug could affect exits or enable DoS/unbacked native; mitigated by extensive new tests.
> 
> **Overview**
> **Hardens wormhole and ZK-tree deposit/exit correctness** and closes several ways deposits or pool traffic could be abused.
> 
> **Deposit / leaf keying:** `record_transfer` now keys `TransferCount` and ZK-tree leaves on **Goldilocks-canonical recipients** (`canonical_leaf_recipient`), so byte aliases of the same leaf encoding cannot share a transfer-count sequence or collide on nullifiers. **`record_transfer_proof`** treats **`None` as native** and **ignores `Some(0)`** (`pallet_assets` id 0), so asset-0 mints cannot inflate `PotentialWormholeBalance` or insert natively exitable leaves.
> 
> **Unsigned exit traffic:** Pool admission (`validate_unsigned`) runs only **cheap pre-validation**; **full plonky2 verify** runs in **`pre_dispatch`** (block inclusion). Dispatch no longer re-verifies (avoids double ZK cost vs declared weight). Pool **`provides`** tags hash **nullifiers**, not raw proof bytes.
> 
> **Leaf hashing:** `hash_leaf` **saturates** quantized amount and `asset_id` to `u32` (no wrap aliases). `canonicalize_account_bytes` is shared from `pallet-zk-tree`.
> 
> **Weights:** Pre-validate nullifier reads/PoV follow **`NUM_LEAF_PROOFS`**; exit and per-transfer weights add **depth-dependent** ZK-tree insert cost (`insert_leaf_db_ops`). Benchmark `MAX_NULLIFIERS` tracks circuit config.
> 
> Tests wire a real `ZkTree` in the wormhole mock and add regression coverage for the above; genesis docs note asset id 0 is not native.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08bed485c315e9f0cd1c6e961b4f5583de684e45. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pallets/mining-rewards/src/mock.rs
+++ b/pallets/mining-rewards/src/mock.rs
@@ -156,10 +156,11 @@ impl qp_wormhole::TransferProofRecorder<sp_core::crypto::AccountId32, u32, u128>
 		from: sp_core::crypto::AccountId32,
 		to: sp_core::crypto::AccountId32,
 		amount: u128,
-	) {
+	) -> bool {
 		RECORDED_PROOFS.with(|proofs| {
 			proofs.borrow_mut().push(RecordedTransferProof { asset_id, from, to, amount });
 		});
+		true
 	}
 
 	fn reveal_address(_account: sp_core::crypto::AccountId32) {}
--- a/pallets/multisig/src/mock.rs
+++ b/pallets/multisig/src/mock.rs
@@ -249,7 +249,8 @@ impl qp_wormhole::TransferProofRecorder<AccountId, u32, Balance> for MockProofRe
 		_from: AccountId,
 		_to: AccountId,
 		_amount: Balance,
-	) {
+	) -> bool {
+		true
 	}
 
 	fn reveal_address(account: AccountId) {
--- a/pallets/reversible-transfers/src/tests/mock.rs
+++ b/pallets/reversible-transfers/src/tests/mock.rs
@@ -239,10 +239,11 @@ impl qp_wormhole::TransferProofRecorder<AccountId, u32, Balance> for MockProofRe
 		from: AccountId,
 		to: AccountId,
 		amount: Balance,
-	) {
+	) -> bool {
 		RECORDED_PROOFS.with(|proofs| {
 			proofs.borrow_mut().push(RecordedTransferProof { asset_id, from, to, amount });
 		});
+		true
 	}
 
 	fn reveal_address(_account: AccountId) {}
--- a/pallets/wormhole/src/benchmarking.rs
+++ b/pallets/wormhole/src/benchmarking.rs
@@ -12,8 +12,9 @@ use qp_wormhole_verifier::{ProofWithPublicInputs, C, F};
 /// This proof is used to benchmark the actual deserialization and verification cost.
 const PRIVATE_BATCH_PROOF_HEX: &str = include_str!("../test-data/private_batch.hex");
 
-/// Maximum number of nullifiers in an private-batch proof (default aggregation size)
-const MAX_NULLIFIERS: u32 = 32;
+/// Maximum number of nullifiers in a private-batch proof: one per leaf proof, derived
+/// from the compiled circuit so the benchmark tracks the `QP_NUM_LEAF_PROOFS` build knob.
+const MAX_NULLIFIERS: u32 = crate::circuit_config::NUM_LEAF_PROOFS as u32;
 
 /// The D const parameter for plonky2 proofs (extension degree = 2)
 const D: usize = 2;
--- a/pallets/wormhole/src/lib.rs
+++ b/pallets/wormhole/src/lib.rs
@@ -1332,8 +1332,24 @@ pub mod pallet {
 			to: <T as Config>::WormholeAccountId,
 			amount: BalanceOf<T>,
 		) {
-			let asset_id_value = asset_id.unwrap_or_default();
-			Self::record_transfer(asset_id_value, &from, &to, amount);
+			// The wormhole tags native leaves with `asset_id == 0`, but `pallet_assets` uses
+			// id 0 for an unrelated, independently-mintable token. Genuine native reaches us as
+			// `None` (from `Balances` events); a `pallet_assets` asset-0 credit reaches us as
+			// `Some(0)` (from `Assets::Issued`). These must not be conflated: a `Some(0)` credit
+			// is backed by no native, so recording it as a native deposit would inflate the
+			// native potential-balance pool and insert a natively-exitable leaf, letting an
+			// asset-0 issuer mint unbacked native out of the wormhole.
+			match asset_id {
+				// Native token.
+				None => Self::record_transfer(T::AssetId::default(), &from, &to, amount),
+				// A `pallet_assets` asset whose id collides with the reserved native id. The
+				// wormhole only supports native exits, and this is not a native deposit, so it
+				// must not touch native accounting — drop it.
+				Some(id) if id == T::AssetId::default() => {},
+				// A genuine non-native asset: recorded as an inert (non-native, never-exitable)
+				// leaf, preserving the existing behaviour for future asset-wormhole support.
+				Some(id) => Self::record_transfer(id, &from, &to, amount),
+			}
 		}
 
 		fn reveal_address(account: <T as Config>::WormholeAccountId) {
--- a/pallets/wormhole/src/lib.rs
+++ b/pallets/wormhole/src/lib.rs
@@ -368,6 +368,10 @@ pub mod pallet {
 		type WeightInfo: WeightInfo;
 
 		/// Override system AccountId for wormhole operations
+		///
+		/// The `AsRef<[u8]>`/`From<[u8; 32]>` bounds allow `record_transfer` to
+		/// canonicalize recipients (reduce each 8-byte limb mod the Goldilocks prime)
+		/// before keying `TransferCount` and inserting ZK-tree leaves.
 		type WormholeAccountId: Parameter
 			+ Member
 			+ MaybeSerializeDeserialize
@@ -375,6 +379,8 @@ pub mod pallet {
 			+ MaybeDisplay
 			+ Ord
 			+ MaxEncodedLen
+			+ AsRef<[u8]>
+			+ From<[u8; 32]>
 			+ Into<<Self as frame_system::Config>::AccountId>
 			+ From<<Self as frame_system::Config>::AccountId>;
 
@@ -393,6 +399,11 @@ pub mod pallet {
 		StorageMap<_, Blake2_128Concat, [u8; 32], bool, ValueQuery>;
 
 	/// Transfer count per recipient - used to generate unique leaf indices in the ZK trie.
+	///
+	/// Keyed on the *canonical* recipient (each 8-byte limb reduced mod the Goldilocks
+	/// prime, matching the ZK leaf encoding — see `canonical_leaf_recipient`), so that a
+	/// recipient and its non-canonical byte aliases share one count sequence and two
+	/// distinct deposits can never commit to identical leaves.
 	#[pallet::storage]
 	#[pallet::getter(fn transfer_count)]
 	pub type TransferCount<T: Config> =
@@ -1217,6 +1228,22 @@ pub mod pallet {
 			}
 		}
 
+		/// Canonical form of a recipient for ZK-leaf keying.
+		///
+		/// Reduces each 8-byte limb of the 32-byte account mod the Goldilocks prime —
+		/// the exact reduction the leaf hash's lossy encoding applies — so that
+		/// `TransferCount` and the stored leaf recipient are keyed per leaf-encoding
+		/// class rather than per raw byte string. The withdrawal circuit binds the
+		/// leaf's recipient felts to a canonical Poseidon output, so canonicalizing
+		/// never changes who can exit a leaf. Non-32-byte accounts are returned
+		/// unchanged (the leaf hash requires 32-byte accounts anyway).
+		fn canonical_leaf_recipient(to: &T::WormholeAccountId) -> T::WormholeAccountId {
+			match <[u8; 32]>::try_from(to.as_ref()) {
+				Ok(bytes) => pallet_zk_tree::tree::canonicalize_account_bytes(bytes).into(),
+				Err(_) => to.clone(),
+			}
+		}
+
 		/// Record a transfer in the ZK tree and emit events.
 		///
 		/// This inserts the transfer data into the 4-ary Poseidon Merkle tree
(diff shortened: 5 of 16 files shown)


n13 (changes requested):
Verdict: request changes.

Blocking findings:

1. [P1] `pallets/wormhole/src/lib.rs:1038-1076` derives both `provides` and `priority` from public inputs that have not been authenticated. The new test explicitly demonstrates that a tampered, unverifiable proof is admitted. An attacker can submit the victim nullifiers first and occupy the semantic tag, or vary nullifiers and set inflated output amounts to fill the pool and force block authors to spend full verification work in `pre_dispatch`. The valid exit then conflicts with attacker-controlled pool state. The pool admission path needs an authenticated or resource-priced gate; hashing unverified nullifiers cannot provide safe deduplication.

2. [P1] `pallets/wormhole/src/lib.rs:1348-1354` changes the key semantics of `TransferCount` without migrating existing state. `STORAGE_VERSION` remains 1 and the runtime migration tuple still only runs v0 -> v1. After upgrade, historical counts under non-canonical aliases are ignored, so the next canonical deposit can reuse a transfer count already committed by an alias and recreate the leaf/nullifier collision this PR is intended to prevent. Please add a versioned migration that consolidates counts per canonical recipient, with try-runtime coverage.

3. [P2] An included proof now runs the cheap phase twice: `pre_dispatch` calls `validate_*_batch_proof`, which deserializes/parses/scans state, and the dispatch body calls `pre_validate_*_batch_proof` again to recover the bundle. The unchanged `verify_*` weights still describe one block-hash/nullifier scan and one deserialization path. This leaves the extra prevalidation work and DB reads outside the declared model. Please add the second prevalidation cost to both private/public weights or restructure so the model covers the actual execution path.

4. [P2] `record_transfer_proof(Some(0), ...)` is now a no-op, but `runtime/src/transaction_extensions.rs:220-229` still reports every collected asset-0 event as recorded, and static `Assets` calls still reserve one full insert. This overcharges fees/block weight and makes reconciliation inaccurate. Filter asset 0 from both the static count and recorded count.

Validation performed on head `9e952a11c164884c30c64600f9bcd8614e8f91f9`:
- `cargo test -p pallet-zk-tree -p pallet-wormhole`: 85 passed, 2 ignored fixture generators.
- `cargo test -p quantus-runtime transaction_extensions`: 35 passed.
- GitHub format, config, and clippy/doc checks pass; the Linux/macOS test matrices are still running.

The passing tests do not cover the upgrade-state or adversarial transaction-pool cases above.

n13 (changes requested):
## Review

Strong PR overall: each of the six hardening changes addresses a real soundness or metering hole, and the claims in the description check out against the code. I verified the load-bearing assumptions rather than taking the comments at face value (details below), and ran the test suites locally — all green. There is **one high-severity issue introduced by this PR** (pool-priority usurpation → cheap exit censorship) that should be fixed before merge, and one pre-deploy state question (missing `TransferCount` migration/audit).

### Verified against the code

- **`pre_dispatch` gate is real.** sp-runtime 45's `CheckedExtrinsic::apply` calls `I::pre_dispatch(&self.function)` before dispatch for `ExtrinsicFormat::Bare` (checked_extrinsic.rs:117), on both authoring and import. So an unverifiable proof cannot enter a block, and a block containing one fails import. The dispatch-body comment claiming this is accurate.
- **Weight accounting nets out to exactly one ZK verify per included extrinsic.** The verify moved to `pre_dispatch`; dispatch re-runs only cheap pre-validation; the declared `verify_private_batch()` weight still covers one verify + storage tail. Failure path charging `pre_validate_proof()` is correct since only cheap work runs there now.
- **`insert_leaf_db_ops_at_depth` formula matches the implementation.** At effective depth `d = min(depth+1, 32)`: reads `3d+4` = `LeafCount` + `Depth`×2 + `grow_tree` `Root` + 3 siblings/level; writes `d+5` = `Leaves` + `LeafCount` + `Root` + `(d−1)` internal nodes + grow's `Nodes`/`Root`/`Depth`. Costing at `depth+1` covers growth. Conservative in the right direction.
- **Canonicalization is circuit-sound.** The withdrawal circuit binds the leaf's recipient felts to `WA(secret)` (a canonical Poseidon output), so re-keying `TransferCount` and the stored leaf on the canonical form never changes who can exit; honest wallets always derive canonical addresses. The `is_ambiguous_account` / event path deliberately still uses the raw recipient — correct, since ambiguity is a property of the real system account.
- **`None` vs `Some(0)` split is consistent with every `record_transfer_proof` call site**: the runtime event scanner passes `None` for `Balances` events and `Some(asset_id)` for `Assets` events; mining-rewards passes `None`; reversible-transfers passes its stored `Option<AssetId>`. Pre-PR, `unwrap_or_default()` really did let a `pallet_assets` asset-0 mint inflate `PotentialWormholeBalance` (the inflation gate in `record_transfer` keys on `asset_id == default()`), so this closes a genuine mint-unbacked-native hole.
- **Depth-scaled PoV can't brick inclusion**: `RuntimeBlockWeights` uses a `u64::MAX` proof-size limit, so the (large, at depth 32 ≈ 260 KB/exit) PoV figures are safe on this chain.
- The golden-vector test plus the circuit cross-check (`pallet_leaf_hash_matches_circuit_leaf_hash`) pinning `hash_leaf` to the real circuit crates is an excellent addition.

### Findings

#### 1. [High] Nullifier `provides` tag + unverified `priority` = free, targeted exit censorship

`validate_unsigned` no longer verifies the proof, but still sets `priority(total_amount)` from `summed_output_amount` values parsed out of the (unverified) public inputs (lib.rs, `validate_unsigned`). Nothing on the cheap path bounds these amounts — `validate_exit_bundle_common` checks asset id, fee rate, block hash, and nullifier freshness only; the `PotentialWormholeBalance` soundness check runs in `process_exit_bundle` at dispatch. So priority is now fully attacker-controlled up to ~2^64.

Combined with the new semantic dedup tag, this enables cheap censorship of a specific exit:

1. Attacker sees a victim's exit proof in gossip (public inputs are plaintext in `proof_bytes`).
2. They mutate the PI section: keep the nullifiers (→ same `provides` tag as the victim's tx), inflate the amounts (→ higher priority). `ProofWithPublicInputs::from_bytes` doesn't cross-check PIs against the proof, and all cheap checks still pass.
3. The pool replaces a same-tag transaction when the newcomer has **strictly higher** priority (client/transaction-pool `graph/ready.rs:449-451`), so the junk **evicts the victim's valid exit**.
4. The junk dies at `pre_dispatch` and is banned by hash — but the victim's tx is already gone and must be resubmitted, and the attacker repeats with a fresh byte-variant. Cost to attacker ≈ zero (unsigned, fee-free, no proving needed).

Pre-PR this was impossible: admission required a full verify, so junk never entered the pool and tags were byte hashes. The same attacker-controlled priority also lets distinct-nullifier junk queue-jump all real exits and burn ~16.7 ms of author verify time each at proposal.

**Suggested fix:** stop deriving priority from unverified amounts — use a constant. The pool rejects same-tag transactions at *equal* priority (`old_priority >= tx.priority` → `TooLowPriority`), and nullifiers are unguessable before the victim gossips, so first-seen wins and targeted usurpation is closed. A cap (`min(total, CAP)`) is not sufficient: it still usurps any victim whose real amount is below the cap. The cost is losing amount-based ordering of exits, which seems acceptable for `Pays::No` traffic. (If priority becomes constant, the incidental issue that the `u64` amount sum can wrap in release builds is moot.)

Residual (non-blocking, worth documenting): with verification off the admission path, an attacker can still cheaply fill the pool with distinct-nullifier junk; that's inherent to this trade-off and strictly better than the pre-PR gossip DoS (where every node paid a full verify per junk proof), but a per-block/per-pool bound on wormhole unsigned txs may be worth a follow-up.

#### 2. [Medium] No migration/audit for existing raw-keyed `TransferCount` entries

The canonical re-keying applies only prospectively. Any pre-upgrade deposit to a non-canonical recipient left a `TransferCount` entry under the raw key; post-upgrade, deposits to that recipient class read the canonical key and restart/continue from a different count — which can reproduce exactly the leaf/nullifier collision this PR fixes, against pre-upgrade leaves. Naturally occurring non-canonical accounts are ~2^-32-per-limb unlikely, but the original bug was exploitable by anyone who could pick a recipient address, so an attacker could have pre-positioned alias deposits on the live chain.

**Ask:** before deploying, audit live-chain `TransferCount` keys for non-canonical limbs (trivial off-chain script over the storage map), and either confirm it's empty in the PR notes or ship a migration that merges alias entries into the canonical key (e.g. `max` of the counts).

#### Minor / optional

- `record_transfer_proof`'s `Some(id) if id == default => {}` arm drops the credit silently. Deliberate and documented, but a `log::warn!` would make an actual asset-0 mint observable in node logs (cheap detection of someone probing this path).
- The `pre_validate_*` 4-tuple `(verifier, proof, bundle, validity)` with callers destructuring `(_, _, bundle, validity)` would read better as a small struct, but not worth blocking on.
- `verify_private_batch()` / `per_transfer_weight()` now perform an uncharged `Depth` storage read each time dispatch info is computed. One read, negligible — just noting it's intentional.

### Test results (run locally on this branch)

- `cargo test -p pallet-zk-tree`: 26 passed, 0 failed (incl. golden vector, saturation, canonicalization tests)
- `cargo test -p pallet-wormhole`: 59 passed, 0 failed, 2 ignored (pre-existing) — incl. all six new security regression tests
- `cargo test -p quantus-runtime transaction_extensions` (SKIP_WASM_BUILD): 35 passed, 0 failed (incl. `per_transfer_weight_scales_with_tree_depth`)

## Verdict

**Request changes.** The six hardening changes are correct and well-tested, and the verify-at-`pre_dispatch` design is sound — but finding 1 replaces the DoS hole with a free, targeted exit-censorship vector, and the fix is a one-liner (constant priority). With that fixed and the `TransferCount` state audit answered (finding 2), this is good to merge.

n13 (approved):
## Re-review (head 1b2b438e)

All four findings from the previous review are addressed, each with a targeted commit and a regression test that pins the exact failure mode. I re-verified each fix against the code rather than the commit messages, and re-ran the suites locally.

### Finding-by-finding

**1. [High] Amount-derived pool priority → free targeted exit censorship — fixed** (d5a38639)

Both unsigned paths now use the constant `UNSIGNED_EXIT_PRIORITY = 1`; nothing on the admission path reads amounts out of the unverified public inputs anymore. Re-confirmed the pool invariant this relies on in the vendored pool: `client/transaction-pool/src/graph/ready.rs:450` rejects a same-tag newcomer unless its priority is *strictly* higher (`old_priority >= tx.priority` → `TooLowPriority`), so with equal priorities first-seen wins and the tag-usurpation replay is closed. `validate_unsigned_priority_ignores_unverified_amounts` reproduces the exact attack shape (mutated PI amount, same nullifiers → same `provides` tag) and asserts identical constant priority.

Residual, documented, non-blocking: an attacker who wins the gossip propagation race can still pre-occupy the tag on nodes that see their junk first. That junk dies at `pre_dispatch` and gets hash-banned, so this is delay-only, costs a race per attempt, and is strictly weaker than the pre-fix vector (which needed no race at all). Consistent with the accepted trade-off; the per-block/pool bound on wormhole unsigned traffic remains a reasonable follow-up.

**2. [Medium] No migration for raw-keyed `TransferCount` — fixed** (96e50fa7)

v1→v2 `CanonicalizeTransferCountKeys` merges non-canonical keys into the canonical key and removes the raw entries; `STORAGE_VERSION` bumped to 2; wired into the runtime `Migrations` tuple after v0→v1; try-runtime `post_upgrade` asserts zero non-canonical keys remain. Correctness details checked:
- `max` is the right merge: every committed leaf in an encoding class has count < max of the class's counts, so the next deposit index is fresh past all pre-upgrade leaves.
- Multiple aliases of one class merge correctly — each pass re-reads the canonical value before taking max.
- `max_count`'s `a.into() >= b.into()` is well-defined (`TransferCount: Into<u64>` bound).
- Collect-then-mutate avoids mutating the map mid-iteration.
- `migration_merges_non_canonical_transfer_count_keys` covers merge, raw-key removal, and count continuation.

One deployment note (non-blocking): the migration iterates the whole `TransferCount` map in one block, and that map holds an entry for every recipient the tx-extension ever recorded — effectively every transfer recipient in chain history. Weight is honestly counted but unbounded. Sanity-check the live entry count before shipping the upgrade; if it has grown into the hundreds of thousands, this wants a multi-block migration. (No PoV-limit concern on this chain, so an overweight upgrade block still imports — the cost is node stall time, not consensus risk.)

**3. [P2] Second prevalidation outside the weight model — fixed** (1b2b438e)

`verify_private_batch` / `verify_public_batch` now add one full `pre_validate_*` weight (compute, reads, and PoV) on top of the ZK + storage-tail base, matching the real inclusion path: `pre_dispatch` runs full validation, then the dispatch body re-runs cheap prevalidation to recover the bundle. If anything this double-counts the first scan's DB reads (already inside the calibration tail) — over-charge, the safe direction. `verify_weights_cover_second_prevalidation` pins the inequality for both paths.

**4. [P2] Dropped asset-0 credits still counted as recorded — fixed** (510edf46)

`record_transfer_proof` now returns `bool`; `record_proofs_from_events_since` counts only `true` returns, and `count_transfers` statically charges 0 for asset-0 `Assets` calls via `asset_transfer_counts`. Both directions are tested (`asset_zero_credit_is_not_counted_as_recorded_transfer` covers the static matcher, the full lifecycle showing zero extra block weight, and the direct event-scan count). All trait impls/mocks updated. The `log::warn!` on the dropped credit from the minor notes is in as well.

### Test results (local, head 1b2b438e)

- `cargo test -p pallet-zk-tree`: 26 passed, 0 failed
- `cargo test -p pallet-wormhole`: 62 passed, 0 failed, 2 ignored (pre-existing fixture generators)
- `cargo test -p quantus-runtime transaction_extensions`: 36 passed, 0 failed

CI at review time: format, config, clippy/doc, and the macOS build/test matrix green; the Linux matrix still running (same suites the macOS leg already passed).

## Verdict

**Approve.** The two blocking findings are properly closed — constant priority provably makes first-seen win under the pool's strict-inequality replacement rule, and the v2 migration removes the pre-upgrade collision window with try-runtime coverage — and both P2 metering gaps are now inside the declared weight model with tests locking them in. Only ask before deploying the upgrade: eyeball the live `TransferCount` entry count so the single-block migration is known to fit.

PR #633, opened 2026-07-30
V12 e2e wormhole

illuzen:
# Security hardening: header hashing, exit-bundle settlement, and hook metering

This PR fixes a batch of security-review findings across consensus, the wormhole
pallet, mining rewards, and the node build. Each fix is a separate commit with a
regression test where the behavior is testable.

### Reject headers whose digest overflows the hash window (`22052b7`)

`Header::hash()` commits the digest through a fixed 110-byte window
(`DIGEST_LOGS_SIZE`, mirrored by the ZK circuit) and silently truncated anything
beyond it, so two distinct sealed headers could share the same canonical block
hash once their encoded digest exceeded the cap — breaking the assumption that
the block hash commits to the full header.

- `sc-consensus-qpow` now rejects any block at import whose post-seal encoded
  digest exceeds `DIGEST_LOGS_SIZE` (new `DigestTooLong` error), so no header
  whose hash would truncate can ever enter the chain.
- `DIGEST_LOGS_SIZE` is now a `pub const` in `qp-header`, shared by the client
  check, with a `debug_assert!` in `Header::hash()` guarding the invariant in
  tests while keeping hashing infallible in release.
- Tests pin that the canonical PoW digest (preimage + seal) fills the window
  exactly, and that an oversized digest trips the assertion.

### Don't revert an exit bundle when a single mint fails (`81ffa0b`)

Exit-account and miner-fee credits used `increase_balance(Precision::Exact)?`,
so one below-existential-deposit mint to a fresh account aborted the dispatch
and rolled back every co-bundled user's exit. Since `MinimumTransferAmount`
only bounds the summed total and the extrinsic is unsigned with `Pays::No`,
anyone could grief a public batch for free with one dust-to-fresh-account
segment. The aggregator rebate was already hardened against exactly this
failure; the same isolation now covers the other two mints:

- **Miner fee**: on mint failure the fee is burned, mirroring the existing
  no-author branch.
- **Exit slot**: a failed mint skips just that slot, emits a new
  `ExitMintFailed` event (appended last in the enum, existing indexer positions
  unchanged), records no transfer proof, and is left out of the committed
  `TotalWormholeExits`. The skipped slot's nullifier stays marked, so the
  deposit can never be re-exited and the accounting cannot enable a
  double-mint.

### Reject intra-segment duplicate nullifiers before minting exits (`8ac966e`)

The private-batch exit grouping summed a replayed leaf into one inflated exit
while its shared nullifier would only be marked once. `segment_validity` now
denies any segment containing duplicate real nullifiers, so settlement cannot
mint the inflated value.

### Remove the unbounded v1→v2 TransferCount migration (`cdcf14d`)

`CanonicalizeTransferCountKeys` iterated the entire `TransferCount` map — an
unbounded, permissionlessly-grown storage map — into memory in a single
`on_runtime_upgrade`, so its work was attacker-controlled and could stall the
network at the upgrade boundary. The migration was also pointless: canonical
keying is enforced at write time and a live audit of the only pre-v2 chain
(Planck testnet) found zero non-canonical keys. Removed, with the storage
version rolled back to 1.

### Bound miner-digest extraction to O(1) scan (`a4c67a7`)

`extract_author_from_digest` cloned and scanned arbitrarily many
author-controlled digest items before finding the QPoW preimage, doing
unmetered work inside `on_finalize` (mining rewards) and `process_exit_bundle`
(miner fee). It now takes borrowed digest items (no payload clones) and
evaluates only the first `PreRuntime(POW_ENGINE_ID, ..)` item, so extraction is
O(1) in digest length regardless of any upstream bound.

### Price the mining-reward finalize hook by live ZK-tree depth (`083b449`)

`on_finalize` mints up to three rewards, each recording a wormhole transfer
proof that inserts a ZK-tree leaf whose DB cost grows with tree depth, but the
hook reserved a flat shallow-tree benchmark weight and under-charged as the
shared tree deepened. The weight now prices the leaf inserts from live depth
(`insert_leaf_db_ops`), matching the runtime's per-transfer proof metering; the
depth-blind `()` fallback prices at `MAX_TREE_DEPTH` so it can never
under-charge.

### Reject non-hex `SUBSTRATE_CLI_GIT_COMMIT_HASH` in the build script (`cac2a89`)

The vendored `generate_cargo_keys()` interpolates this env override into
line-oriented `cargo:` build-script output after only trimming it, so a value
with an embedded newline could smuggle an extra cargo directive
(`rustc-env`/`cfg`/`link-arg`) and gain authority over the compiled artifact.
The node build script now accepts only a short/full hex git hash (empty stays
inert) and falls back to `unknown` otherwise.

## Testing

- New/updated unit tests accompany each fix: digest-window fit and overflow
  assertions (`qp-header`), below-ED exit isolation and duplicate-nullifier
  denial (`pallet-wormhole`), depth-aware weight formulas (`pallet-mining-rewards`
  and `pallet-wormhole` weights), and build-script hash sanitization.
- Full `pallet-wormhole` (62), `pallet-mining-rewards`, and `qp-header` suites
  pass; `quantus-runtime` compiles.

---

> [!NOTE]
> **High Risk**
> Touches consensus import rules, wormhole minting/accounting, runtime migrations removal, and mandatory block hooks—errors could fork nodes, strand deposits, or mis-meter blocks.
> 
> **Overview**
> Security hardening across **consensus**, **wormhole settlement**, **mining rewards**, and the **node build**.
> 
> **Consensus:** `sc-consensus-qpow` now rejects blocks whose post-seal SCALE-encoded digest exceeds the shared **`DIGEST_LOGS_SIZE` (110)** window (`DigestTooLong`), so headers that would be truncated in `Header::hash()` cannot be imported. The constant is published from `qp-header` with tests and a debug assertion backing the invariant.
> 
> **Wormhole:** Exit bundles no longer roll back when a single **exit** or **miner fee** mint fails (e.g. below existential deposit)—failed slots emit **`ExitMintFailed`**, skip minting, and only successfully minted amounts update **`TotalWormholeExits`** while nullifiers stay spent. **`segment_validity`** now rejects **duplicate real nullifiers within a segment** to block inflated exits from replayed leaves. The unbounded **v1→v2 `TransferCount` migration** is removed and storage version stays at **1** (canonical keys enforced at write time). **`extract_author_from_digest`** takes borrowed digest items and only inspects the first PoW pre-runtime entry (O(1), no clones).
> 
> **Mining rewards:** **`on_finalize`** weight prices up to three ZK-tree leaf inserts using **live tree depth** via `pallet-zk-tree`, with a max-depth fallback and unit tests.
> 
> **Node build:** **`SUBSTRATE_CLI_GIT_COMMIT_HASH`** is sanitized to hex-only (or `unknown`) before `generate_cargo_keys()` to block cargo-directive injection via newlines.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb5d5c32c0fcb452505a3b119ef8783de002e9ab. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8853,6 +8853,7 @@ dependencies = [
  "parity-scale-codec",
  "parking_lot 0.12.4",
  "primitive-types 0.13.1",
+ "qp-header",
  "sc-client-api",
  "sc-consensus",
  "sp-api",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5941,6 +5941,7 @@ dependencies = [
  "log",
  "pallet-balances",
  "pallet-treasury",
+ "pallet-zk-tree",
  "parity-scale-codec",
  "qp-wormhole",
  "scale-info",
--- a/client/consensus/qpow/Cargo.toml
+++ b/client/consensus/qpow/Cargo.toml
@@ -33,6 +33,7 @@ default = ["std"]
 std = [
 	"codec/std",
 	"primitive-types/std",
+	"qp-header/serde",
 	"qp-header/std",
 	"sp-api/std",
 	"sp-consensus-qpow/std",
--- a/client/consensus/qpow/Cargo.toml
+++ b/client/consensus/qpow/Cargo.toml
@@ -15,6 +15,7 @@ log = { workspace = true, default-features = false }
 parking_lot = { workspace = true, default-features = true }
 primitive-types = { workspace = true, default-features = false }
 prometheus-endpoint = { workspace = true, default-features = true }
+qp-header = { workspace = true }
 sc-client-api = { workspace = true, default-features = false }
 sc-consensus = { workspace = true }
 sp-api = { workspace = true, default-features = false }
@@ -32,6 +33,7 @@ default = ["std"]
 std = [
 	"codec/std",
 	"primitive-types/std",
+	"qp-header/std",
 	"sp-api/std",
 	"sp-consensus-qpow/std",
 	"sp-runtime/std",
--- a/client/consensus/qpow/src/lib.rs
+++ b/client/consensus/qpow/src/lib.rs
@@ -70,10 +70,8 @@ pub enum Error<B: BlockT> {
 	CheckInherentsUnknownError(sp_inherents::InherentIdentifier),
 	#[error("Multiple pre-runtime digests")]
 	MultiplePreRuntimeDigests,
-	#[error(
-		"Header {0:?} has an encoded digest of {1} bytes, exceeding the {2}-byte commitment window"
-	)]
-	DigestTooLong(B::Hash, usize, usize),
+	#[error("Header has an encoded digest of {0} bytes, exceeding the {1}-byte commitment window")]
+	DigestTooLong(usize, usize),
 	#[error(transparent)]
 	Client(sp_blockchain::Error),
 	#[error(transparent)]
@@ -226,6 +224,21 @@ where
 		&self,
 		mut block_import_params: BlockImportParams<B>,
 	) -> Result<ImportResult, Self::Error> {
+		// Reject any header whose canonical (post-seal) digest encodes to more
+		// than the fixed window committed by `Header::hash()`. Past that window
+		// the digest bytes are not covered by the block hash, so silently
+		// truncating (as `hash()` does defensively) would let two distinct
+		// sealed headers collide on the same block hash. Fail closed before
+		// *any* `hash()` call on this header — hashing an oversized digest
+		// trips a debug assertion in `qp-header` — and, for the same reason,
+		// without embedding a hash of the malformed header in the error. The
+		// pre-seal digest is a prefix of the post-seal one, so this bound also
+		// covers the pre-seal `header.hash()` calls below.
+		let encoded_digest_len = block_import_params.post_header().digest().encode().len();
+		if encoded_digest_len > DIGEST_LOGS_SIZE {
+			return Err(Error::<B>::DigestTooLong(encoded_digest_len, DIGEST_LOGS_SIZE).into());
+		}
+
 		let parent_hash = *block_import_params.header.parent_hash();
 
 		if let Some(inner_body) = block_import_params.body.take() {
@@ -251,23 +264,6 @@ where
 		)?;
 
 		let pre_hash = block_import_params.header.hash();
-
-		// Reject any header whose canonical (post-seal) digest encodes to more
-		// than the fixed window committed by `Header::hash()`. Past that window
-		// the digest bytes are not covered by the block hash, so silently
-		// truncating (as `hash()` does defensively) would let two distinct
-		// sealed headers collide on the same block hash. Fail closed here
-		// instead, on the same digest that `post_header().hash()` commits below.
-		let post_header = block_import_params.post_header();
-		let encoded_digest_len = post_header.digest().encode().len();
-		if encoded_digest_len > DIGEST_LOGS_SIZE {
-			return Err(Error::<B>::DigestTooLong(
-				post_header.hash(),
-				encoded_digest_len,
-				DIGEST_LOGS_SIZE,
-			)
-			.into());
-		}
 
 		// Convert seal to nonce
 		let nonce: [u8; 64] = inner_seal
@@ -401,6 +397,19 @@ async fn extract_pow_seal<B>(
 where
 	B: BlockT<Hash = H256>,
 {
+	// This is the first point in the import pipeline that hashes a
+	// network-supplied header, and hashing a digest longer than the committed
+	// window trips a debug assertion in `qp-header` (the hash would silently
+	// truncate it). Reject the oversized digest before any `hash()` call; the
+	// authoritative check lives in `import_block`, this one only keeps the
+	// hashing below panic-free in debug builds.
+	let encoded_digest_len = block.header.digest().encode().len();
+	if encoded_digest_len > DIGEST_LOGS_SIZE {
+		return Err(format!(
+			"Header has an encoded digest of {encoded_digest_len} bytes, exceeding the {DIGEST_LOGS_SIZE}-byte commitment window"
+		));
+	}
+
 	let hash = block.header.hash();
 	let header = &mut block.header;
 	let block_hash = hash;
--- a/client/consensus/qpow/src/lib.rs
+++ b/client/consensus/qpow/src/lib.rs
@@ -13,6 +13,9 @@ use sp_consensus_qpow::{QPoWApi, Seal as RawSeal};
 use sp_runtime::traits::Block as BlockT;
 use std::{marker::PhantomData, sync::Arc, time::Duration};
 
+use codec::Encode;
+use qp_header::DIGEST_LOGS_SIZE;
+
 use crate::worker::UntilImportedOrTransaction;
 pub use crate::worker::{MiningBuild, MiningHandle, MiningMetadata, RebuildTrigger};
 use futures::{Future, Stream, StreamExt};
@@ -67,6 +70,10 @@ pub enum Error<B: BlockT> {
 	CheckInherentsUnknownError(sp_inherents::InherentIdentifier),
 	#[error("Multiple pre-runtime digests")]
 	MultiplePreRuntimeDigests,
+	#[error(
+		"Header {0:?} has an encoded digest of {1} bytes, exceeding the {2}-byte commitment window"
+	)]
+	DigestTooLong(B::Hash, usize, usize),
 	#[error(transparent)]
 	Client(sp_blockchain::Error),
 	#[error(transparent)]
@@ -244,6 +251,23 @@ where
 		)?;
 
 		let pre_hash = block_import_params.header.hash();
+
+		// Reject any header whose canonical (post-seal) digest encodes to more
+		// than the fixed window committed by `Header::hash()`. Past that window
+		// the digest bytes are not covered by the block hash, so silently
+		// truncating (as `hash()` does defensively) would let two distinct
+		// sealed headers collide on the same block hash. Fail closed here
+		// instead, on the same digest that `post_header().hash()` commits below.
+		let post_header = block_import_params.post_header();
+		let encoded_digest_len = post_header.digest().encode().len();
+		if encoded_digest_len > DIGEST_LOGS_SIZE {
+			return Err(Error::<B>::DigestTooLong(
+				post_header.hash(),
+				encoded_digest_len,
+				DIGEST_LOGS_SIZE,
+			)
+			.into());
+		}
 
 		// Convert seal to nonce
 		let nonce: [u8; 64] = inner_seal
--- a/client/consensus/qpow/src/lib.rs
+++ b/client/consensus/qpow/src/lib.rs
@@ -699,3 +699,82 @@ where
 		.get_difficulty(parent)
 		.map_err(|_| Error::Runtime("Failed to fetch difficulty".into()))
 }
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use sp_consensus::BlockOrigin;
+	use sp_runtime::{traits::BlakeTwo256, OpaqueExtrinsic};
+
+	type TestHeader = qp_header::Header<u32, BlakeTwo256>;
+	type TestBlock = sp_runtime::generic::Block<TestHeader, OpaqueExtrinsic>;
+
+	fn sealed_header(digest: Digest) -> TestHeader {
+		<TestHeader as HeaderT>::new(
+			1,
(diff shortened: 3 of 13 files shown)


n13 (approved):
## Review

Went through each fix against the surrounding code at `cb5d5c3`. The security reasoning holds up in every case, the regression tests pin the right behaviors, and CI is green (Linux build+test, clippy, fmt). Verdict at the bottom.

### Verified correct

- **Digest overflow rejection** — the check sits in `PowBlockImport::import_block`, so it covers every imported block (sync included), and it runs before `inner.import_block`, so an oversized digest is rejected before execution. The canonical digest (38-byte pre-runtime item + 71-byte seal item + 1-byte vec length) fills the 110-byte window exactly, and the new test pins that, so no legitimate block can trip the check. A second PoW pre-runtime item (148 bytes post-seal) can no longer reach the runtime, which also makes the single-item early-exit in `extract_author_from_digest` sound.
- **Exit-slot isolation** — order of operations checks out: mint attempts (aggregator rebate, miner fee) happen before the single burn executes, so a failed mint's fallback into `burn_amount_u128` is actually burned. `UsedNullifiers` are inserted during segment processing (lib.rs:850), before the mint loop, so a skipped slot stays spent and `committed_exits ≤ exits_after ≤ potential_balance` holds — no double-mint path. The new event is appended last in the enum, so existing indexer positions are stable.
- **Intra-segment duplicate rejection** — zero nullifiers are filtered out before the `BTreeSet` check, so dummy-leaf padding is unaffected; only replayed real nullifiers deny a segment. Denying the whole segment (not just the duplicate slot) is right, since the grouped exit amount is already inflated at that point.
- **Migration removal** — confirmed the v1→v2 migration never shipped: the latest release (`v0.8.0-exodus`) has `STORAGE_VERSION = 1` and no `MigrateV1ToV2`; the migration only existed on `main` between #631 and this PR. Rolling the constant back to 1 is therefore safe for any network running a released binary. Only caveat: an internal devnet that upgraded from a `main` build in that window would hold on-chain version 2 and would need a reset or a one-off version rewrite.
- **Depth-aware finalize weight** — `on_initialize` reserves the new weight, and `insert_leaf_db_ops_at_depth` already prices at `depth + 1` (capped at `MAX_TREE_DEPTH`), so a tree-growth event within the block is covered. The `()` fallback pricing at `MAX_TREE_DEPTH` can only over-charge, never under-charge, and the tests pin both properties.
- **Build-script sanitization** — allowlist approach is right; empty value stays inert (matches the doc comment), odd-length short hashes are accepted, and the 64-byte cap covers SHA-256 object names.

### Findings

1. **`spec_version` stays at 136 while runtime metadata changes** (medium — same as the Bugbot comment). `v0.8.0-exodus` already ships spec 136, and this PR adds `ExitMintFailed` to the event enum. If this runtime ever reaches an existing chain as an upgrade, clients that cache metadata by spec version will mis-decode. If v12 is a fresh-genesis relaunch this is moot, but bumping to 137 here is cheap insurance either way — recommend doing it in this PR unless the release process bumps it separately.

2. **`DigestTooLong` path panics in debug builds** (low — same as the Bugbot comment, plus one extra spot). The error branch calls `post_header.hash()`, which now trips the `debug_assert!` in `Header::hash()` — so in a debug build, a malicious peer sending an oversized-digest block crashes the node instead of getting a clean rejection. `fetch_seal(..., header.hash())` and `pre_hash` also hash before the check runs, which has the same issue for an oversized *pre-seal* digest. Suggest moving the length check to the top of `import_block` and building the error from `parent_hash` + block number instead of hashing the offending header. Release behavior is correct as-is, so this is polish, not a blocker.

3. **Skipped exits permanently forfeit the deposit** (design note, not a defect). A below-ED exit to a fresh account burns that user's funds irreversibly: nullifier spent, no retry, and the value also stays in `PotentialWormholeBalance` as headroom forever (marginally loosening the soundness bound by the skipped amount). This is the right trade against the free-griefing vector, and it's self-inflicted (the exiter chooses amount and destination) — but it's worth surfacing in wallet/SDK docs so clients warn before submitting a below-ED exit to a nonexistent account.

### Verdict

**Approve.** All six fixes are correct, conservatively designed, and well-tested; the two open items (spec bump, debug-build panic polish) are small and can land here or as a fast follow.

n13 (changes requested):
## GPT 5.6 Verdict: Request changes

The settlement, duplicate-nullifier, migration-removal, build-script, and author-extraction changes look sound, and the focused/full pallet tests pass on the merge result. One metering issue still blocks approval.

### Blocking: `on_finalize` reserves against a depth that can be stale

`pallet_mining_rewards::on_initialize` evaluates `on_finalize_rewarded_miner()` before any extrinsics run, but the three reward leaves are inserted later in `on_finalize`. Between those points, normal transfers and wormhole exits can increase `ZkTree::Depth`.

`insert_leaf_db_ops_at_depth` only prices `depth + 1`. That covers one growth, but it is not a bound on the depth at finalize. With current `main` merged, a public batch can contain up to `2 * 7 * 53 = 742` exit slots; from a shallow tree, one such extrinsic can grow through several depths in the same block. The finalize hook then walks the deeper tree while the block only reserved the start-of-block depth plus one. Because mandatory finalization cannot correct its consumed weight afterward, this leaves the hook under-metered precisely in the path this PR is intended to harden.

Please reserve at `MAX_TREE_DEPTH`, or derive and test a provable upper bound on the end-of-block depth. A regression test should set a shallow depth, reserve the initialize weight, grow the tree by the maximum allowed intervening work, and verify the reserved weight still covers finalize.

### Secondary: reject the overlong digest before hashing it

The `DigestTooLong` branch calls `post_header.hash()`, which triggers the new `debug_assert!` for the exact malformed digest being rejected. The earlier `header.hash()` calls have the same problem when the pre-seal digest is already oversized. Move the length check ahead of every hash call and do not include a hash of the malformed header in the error. This is release-safe today but remains a remotely triggerable node panic in debug builds.

The earlier `spec_version` concern is resolved by current `main`: the merge result retains version 137.

Validation on the merge result:

- `cargo test -p pallet-wormhole --lib` — 62 passed, 2 ignored
- `cargo test -p pallet-mining-rewards --lib` — 24 passed, 1 ignored
- `cargo test -p qp-header --lib` — 6 passed
- `cargo test -p qp-wormhole --lib` — passed (0 tests)
- `git diff --check` — clean

n13 (changes requested):
## GPT-5.6 re-review verdict: Request changes

Reviewer: **GPT-5.6**

Commit `986cea3d` correctly fixes the stale-depth finalize reservation: both runtime and fallback weights now price the three inserts at `MAX_TREE_DEPTH`, and the new all-depth regression test pins that bound.

One blocker remains in the digest fix.

### Blocking: network sync hashes the header before either new pre-check

The checks in `extract_pow_seal` and `PowBlockImport::import_block` protect the consensus import path once a header reaches it, but they are not the first hashes of network-supplied headers. Generic sync code hashes untrusted data earlier, for example:

- `client/network-sync/src/engine.rs`: `push_block_announce_validation` immediately calls `announce.header.hash()`.
- `client/network-sync/src/strategy/chain_sync.rs`: downloaded headers are hashed while matching responses and validating `b.hash`.

Those paths run before the QPoW import-queue verifier, so an oversized digest still reaches the `debug_assert!` in `qp_header::Header::hash()` and can panic a debug node. The existing `oversized_digest_trips_hash_invariant` test confirms the panic is still present.

The robust fix is to keep `Header::hash()` infallible for arbitrary decoded/network input: remove the `debug_assert!` and replace the panic test with regression tests showing the verifier/importer rejects an oversized digest cleanly. Import rejection can remain the consensus gate; importer-local checks cannot establish a precondition for every generic network caller of `HeaderT::hash()`.

Everything else reviewed cleanly, and current `main` still merges without conflicts.

Validation on the merge result:

- `cargo test -p sc-consensus-qpow --lib` — compiled and passed (0 tests)
- `cargo test -p pallet-mining-rewards --lib` — 24 passed, 1 ignored
- `cargo test -p qp-header --lib` — 6 passed
- `cargo test -p pallet-wormhole --lib` — 62 passed, 2 ignored
- `git diff --check` — clean

n13 (approved):
## Fable 5 Verdict: Approve

Re-review at `986cea3d`, following the requested changes. Both items from the earlier round are resolved and re-verified against the code; the newly raised sync-layer blocker is assessed below with build-profile evidence — my conclusion is that it is real but not blocking for any shipped artifact, and has a one-line resolution.

### Previously blocking — stale-depth finalize weight: resolved

Both `SubstrateWeight` and the `()` fallback now price the three leaf inserts at `MAX_TREE_DEPTH`. The reservation is depth-independent by construction — it no longer reads live tree state at all (the `pallet_zk_tree::Config` bound on `SubstrateWeight` is gone; only the free function and constant are used), so there is no window in which intervening extrinsics can invalidate it. The new regression test `reserved_weight_covers_any_end_of_block_depth` asserts the reserved weight covers the finalize cost at every depth `0..=MAX_TREE_DEPTH`, which is strictly stronger than the single grow-scenario test that was asked for, and is sufficient precisely because the reservation provably never varies with state.

Verified locally at the PR head: `cargo test -p pallet-mining-rewards --lib` — 24 passed, 1 ignored.

### Previously secondary — digest hashed before rejection: resolved within the import pipeline

The length check now runs at the top of `import_block`, ahead of `parent_hash`, `check_inherents`, `fetch_seal`, and `pre_hash`, and the `DigestTooLong` error no longer embeds a hash of the malformed header. The same gate was added at the top of `extract_pow_seal`, the verifier's first hash of a network-supplied header. I mapped every `hash()` call site in the crate: all hashes of network-supplied headers now sit behind one of the two gates; the one remaining ungated call hashes a locally-built proposal with a canonical digest. The pre-seal bound reasoning holds (the post-seal digest contains every pre-seal item plus the 71-byte seal item, so pre-seal encoded length is strictly smaller).

Verified locally: `cargo test -p qp-header --lib` — 6 passed, including the should-panic invariant test.

### New blocker claim — sync hashes untrusted headers before the consensus checks: real, but not an actual blocker for shipped artifacts

The observation is correct: `HeaderT::hash()` is overridden to route to the Poseidon `Header::hash()` (primitives/header/src/lib.rs:190), and generic sync code (block-announce handling, header downloads) hashes peer-supplied headers before either of the new gates runs. Those paths live outside this crate, so an oversized-digest announcement can still reach the `debug_assert!` — the two gates added in `986cea3d` close the import pipeline, not the networking layer above it.

Whether that makes it a *blocker* comes down to which binaries can actually panic — only those compiled with `debug-assertions = true`. I checked every path that produces a network-facing binary in this repo:

- `[profile.release]` does not set `debug-assertions`, so it defaults to **off**; `[profile.production]` inherits release. There is no `testnet` profile (upstream polkadot-sdk ships one with `debug-assertions = true`, so this was worth ruling out explicitly).
- `quantus-release.yml`, `quantus-hotfix-release.yml`, and `Dockerfile.local` all build `--release`.

So every artifact this project distributes has the assertion compiled out, and in release the behavior is strictly better than before this PR: sync hashes with the saturating backstop, then import rejects with `DigestTooLong`. The exposed party is a developer running a plain `cargo build` node against a public network — a real but narrow surface, and one that existed in no shipped configuration.

That said, the underlying principle (never assert on network-controlled input) is right, and the resolution costs one line: gate the assertion `#[cfg(test)]` instead of `debug_assert!` — the `oversized_digest_trips_hash_invariant` test keeps its teeth inside `qp-header`, while debug-built nodes lose the remotely reachable panic entirely. (Demoting to `log::error!` works too.) I'd land that either in this PR or as an immediate follow-up; it fully moots the disagreement. I'm not treating it as blocking because no shipped or CI-released binary can hit it, the failure mode requires an unusual operator configuration, and release-build behavior is fail-closed and correct.

### spec_version: resolved by main

Confirmed `main` is now at `spec_version: 137` and this PR does not touch that line, so the merge result retains 137. Bugbot's re-anchored "spec_version stays at 136" comment on the new commit is stale and can be ignored.

### Non-blocking note

The `MAX_TREE_DEPTH` pricing reserves ~307 reads / ~116 writes (roughly 19 ms at `RocksDbWeight`) every block regardless of actual depth — on the order of 1% of a 2 s block budget, permanently. Correct trade for a mandatory hook that cannot refund; if block space ever gets tight, a provable end-of-block depth bound could reclaim most of it.

### CI

Format and clippy are green on `986cea3d`; the Linux/macOS build matrices were still running at review time — worth letting them finish before merge, though the targeted suites pass locally.

**Approve.** Both requested changes are correctly implemented and pinned by tests. The remaining sync-layer assertion concern is legitimate hygiene but demonstrably cannot affect any binary this project ships; recommend the one-line `#[cfg(test)]` gate to close it out.

PR #634, opened 2026-07-30
fix(litep2p): make system DNS resolver the default

### LINKED ISSUE #587: litep2p: default DNS resolver config silently has zero nameservers (hickory-resolver 0.26)
## Problem

`hickory-resolver` 0.26 changed `ResolverConfig::default()` from Google's nameservers (0.24 behavior) to a derived `Default` with an **empty** `name_servers` list.

The vendored litep2p only reads the system DNS config when `use_system_dns_config` is set (via `ConfigBuilder::with_system_resolver()`). Otherwise it builds the resolver from `Default::default()`:

https://github.com/Quantus-Network/chain/blob/main/client/litep2p/src/lib.rs#L157-L168

That default path constructs a resolver with **zero nameservers**, so every `/dns/...` lookup fails silently (`NoConnections`, 0 peers) — exactly the failure mode fixed at the call site in #586.

## Footgun

#586 fixes the only production call site, but the opt-in flag remains. Any future `Litep2p` consumer that forgets `.with_system_resolver()` silently reinherits a dead DNS resolver.

## Proposal

Since the vendored litep2p is only consumed in-repo, make system DNS config the unconditional behavior in `Litep2p::new` and drop the `use_system_dns_config` flag / `with_system_resolver()` builder method. Startup already fails loudly with `Error::CannotReadSystemDnsConfig` when the system config can't be read, which is the behavior we want everywhere.

exocognosis:
## Summary

Make system DNS resolution the default for every consumer of the vendored litep2p configuration.

This change:

- reads the system resolver configuration unconditionally when `Litep2p` is created
- preserves a valid system resolver configuration and its resolver options
- falls back to the populated Google UDP and TCP resolver configuration when reading the system configuration fails
- also falls back when the system configuration is read successfully but contains no nameservers
- removes the `use_system_dns_config` flag and `with_system_resolver()` builder method
- removes the now redundant opt-in from the current network backend

## Root cause

hickory-resolver 0.26 changed `ResolverConfig::default()` so that it contains no nameservers. The vendored litep2p builder still defaulted `use_system_dns_config` to `false`, which selected that empty resolver configuration unless each caller remembered to invoke `with_system_resolver()`.

PR #586 added the opt-in at the current production call site, but the unsafe builder default remained available to future consumers. PR #613 made system configuration read failures fall back to a populated public resolver, but a successful read that returned an empty nameserver list could still leave DNS resolution unusable.

The result was a configuration footgun where new callers could compile successfully while all `/dns/` peer resolution silently failed.

## Proposed solution

Resolver selection now belongs to `Litep2p::new()` instead of an optional builder flag. Construction always attempts to load the system DNS configuration.

A valid system configuration is used unchanged. If the system read fails, or if it succeeds with zero nameservers, litep2p logs the condition and uses `ResolverConfig::udp_and_tcp(&GOOGLE)` with default resolver options. This preserves the startup resilience introduced by PR #613 while ensuring that the resolver never receives the known empty default configuration.

Removing the flag and builder method makes the safe behavior automatic for current and future consumers. The existing network backend no longer needs to opt in explicitly.

## Impact

- Future litep2p consumers cannot accidentally construct a resolver with zero nameservers by omitting a builder call.
- Existing system DNS behavior remains unchanged when the operating system provides a valid configuration.
- Nodes still start with a populated fallback resolver when the operating system configuration cannot be read.
- No user-facing configuration is removed because the deleted switch was an internal vendored builder API.

## Verification

- `cargo test -p litep2p resolver_config_falls_back_when_system_config_is_unavailable_or_empty -- --nocapture`
- `cargo test -p litep2p`: 322 passed, 2 ignored
- `cargo check -p sc-network`
- `cargo clippy -p litep2p --lib`: completed successfully with existing warnings
- `git diff --check`

## Baseline notes

- `cargo fmt --all -- --check` reports formatting differences across untouched upstream files with the pinned stable toolchain. The changed hunk identified by the formatter was corrected.
- Strict all-target clippy with `-D warnings` remains blocked by existing `unwrap`, `expect`, `panic`, and other lint findings in the vendored package. The normal library clippy command completes successfully and does not identify a warning in the added production code.

Fixes #587

---

> [!NOTE]
> **Medium Risk**
> Changes default peer dialing DNS for all litep2p users (now OS-first with public fallback), which affects networking reliability but removes a footgun that could silently break `/dns/` resolution.
> 
> **Overview**
> **DNS setup is no longer optional** when constructing `Litep2p`. The `use_system_dns_config` flag and `with_system_resolver()` builder API are removed; `Litep2p::new()` always loads the OS resolver and keeps its options when nameservers are present.
> 
> A new `resolver_config_or_fallback` helper centralizes behavior: failed system reads or configs with **zero nameservers** log an error and use Google public DNS instead of hickory’s empty default (which broke `/dns/` dials after hickory 0.26). Unit tests cover unavailable and empty system configs.
> 
> The Substrate network backend drops its explicit `with_system_resolver()` call because the safe path is now automatic for all litep2p consumers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cefc25745a14ff89d177f63ab4b071e22f2a8203. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/litep2p/src/config.rs
+++ b/client/litep2p/src/config.rs
@@ -112,9 +112,6 @@ pub struct ConfigBuilder {
 
 	/// Close the connection if no substreams are open within this time frame.
 	keep_alive_timeout: Duration,
-
-	/// Use system's DNS config.
-	use_system_dns_config: bool,
 }
 
 impl Default for ConfigBuilder {
@@ -144,7 +141,6 @@ impl ConfigBuilder {
 			known_addresses: Vec::new(),
 			connection_limits: ConnectionLimitsConfig::default(),
 			keep_alive_timeout: KEEP_ALIVE_TIMEOUT,
-			use_system_dns_config: false,
 		}
 	}
 
@@ -254,12 +250,6 @@ impl ConfigBuilder {
 		self
 	}
 
-	/// Set DNS resolver according to system configuration instead of default (Google).
-	pub fn with_system_resolver(mut self) -> Self {
-		self.use_system_dns_config = true;
-		self
-	}
-
 	/// Build [`Litep2pConfig`].
 	pub fn build(mut self) -> Litep2pConfig {
 		let keypair = match self.keypair {
@@ -285,7 +275,6 @@ impl ConfigBuilder {
 			known_addresses: self.known_addresses,
 			connection_limits: self.connection_limits,
 			keep_alive_timeout: self.keep_alive_timeout,
-			use_system_dns_config: self.use_system_dns_config,
 		}
 	}
 }
@@ -340,7 +329,4 @@ pub struct Litep2pConfig {
 
 	/// Close the connection if no substreams are open within this time frame.
 	pub(crate) keep_alive_timeout: Duration,
-
-	/// Use system's DNS config.
-	pub(crate) use_system_dns_config: bool,
 }
--- a/client/litep2p/src/lib.rs
+++ b/client/litep2p/src/lib.rs
@@ -51,7 +51,7 @@ use crate::{
 use crate::transport::websocket::WebSocketTransport;
 
 use hickory_resolver::{
-	config::{ResolverConfig, GOOGLE},
+	config::{ResolverConfig, ResolverOpts, GOOGLE},
 	TokioResolver,
 };
 use multiaddr::{Multiaddr, Protocol};
@@ -157,24 +157,8 @@ impl Litep2p {
 		let bandwidth_sink = BandwidthSink::new();
 		let mut listen_addresses = vec![];
 
-		let (resolver_config, resolver_opts) = if litep2p_config.use_system_dns_config {
-			// `ResolverConfig::default()` has no nameservers since hickory 0.26, so on a
-			// failed system read fall back to public DNS rather than leave the node unable
-			// to resolve `/dns/` peers.
-			match hickory_resolver::system_conf::read_system_conf() {
-				Ok(conf) => conf,
-				Err(error) => {
-					tracing::error!(
-						target: LOG_TARGET,
-						?error,
-						"failed to read system DNS config; falling back to public DNS",
-					);
-					(ResolverConfig::udp_and_tcp(&GOOGLE), Default::default())
-				},
-			}
-		} else {
-			(Default::default(), Default::default())
-		};
+		let (resolver_config, resolver_opts) =
+			resolver_config_or_fallback(hickory_resolver::system_conf::read_system_conf());
 		let resolver = Arc::new(
 			TokioResolver::builder_with_config(resolver_config, Default::default())
 				.with_options(resolver_opts)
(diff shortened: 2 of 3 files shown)


n13:
Thank you. I think we already did this. A very similar bug anyway. Will investigate.

n13:
## Assessment — Fable 5 (automated review, posted at the maintainer's request)

### Reference base (repo history)

The DNS story in this repo has three prior steps, all still present on `main`:

1. `e35305ec` bumped the vendored litep2p's hickory-resolver 0.24 → 0.26. Upstream breaking change (verified against crate sources): in 0.24, `ResolverConfig::default()` was Google's public DNS; in 0.26.1 it is a derived `Default` with an **empty** nameserver list.
2. #586 (`f0275b0a`) opted the sc-network backend into system DNS via `.with_system_resolver()`.
3. #613 (`dc5bf07d`) made a failed `read_system_conf()` log loudly and fall back to `ResolverConfig::udp_and_tcp(&GOOGLE)` instead of aborting startup.

So nothing from the earlier fixes was lost — this PR is not re-fixing a regression. It implements the proposal from #587: the fixed behavior lived at one call site behind an opt-in flag, and any `Litep2p` consumer that forgot `.with_system_resolver()` silently got a resolver with zero nameservers. That footgun is real in-repo today, not hypothetical: `client/network/src/litep2p/discovery.rs` and `shim/request_response/tests.rs` construct `Litep2p` without the flag and currently inherit the dead default resolver.

### What the PR does, verified against hickory 0.26.1 sources

Resolver selection moves into `Litep2p::new()`, unconditionally: system config first; on read error **or** an Ok-but-empty nameserver list, log an error and fall back to Google. The flag, builder method, and the sc-network opt-in call are deleted.

This preserves the intended behavior end to end — "prefer system DNS" from #586, "never abort, fall back to public DNS" from #613 — and closes one real gap:

- On Linux the Ok-but-empty guard is defense-in-depth only: hickory's `system_conf/unix.rs` already errors when no nameservers are found.
- On **macOS and Windows** it is load-bearing: hickory 0.26.1 uses platform-native readers (`apple.rs` reads the SystemConfiguration dynamic store, `windows.rs` the registry) and both can return `Ok` with an empty server list, which previously produced a resolver that resolves nothing. (The PR description's framing is slightly broader than reality here — minor nit only.)
- The #613 macOS scenario (zoned link-local nameserver `fe80::1%en0`) remains covered: `apple.rs` fails `IpAddr::from_str` on zoned addresses, which lands in the `Err` → Google fallback branch.

For the production node, behavior is identical to `main` on any host with a valid system DNS config. The only changes are the empty-config edge case above and that test consumers now get a working resolver.

### Verification performed locally

- Single commit, based on current `main` HEAD (`ece185b8`); diff touches only the three listed files. No dependency, `Cargo.lock`, CI, or build-script changes; no `unsafe`; the only network endpoints involved are hickory's own `GOOGLE` constant (8.8.8.8 / 8.8.4.4 + IPv6), already the fallback on `main`.
- `cargo test -p litep2p --lib resolver_config_falls_back` — passes.
- `cargo check -p sc-network` — passes (consumer compiles cleanly without the removed builder call).

### Notes for the maintainer

- CI has not run on this branch (fork PR awaiting approval) — trigger it before merging.
- Removing `with_system_resolver()` grows the divergence from upstream `paritytech/litep2p`, which still has the opt-in API. Acceptable since the crate is vendored and only consumed in-repo (and #587 explicitly proposed this), but worth remembering on future upstream syncs.

### Verdict (guidance only — not an approval)

The change is correct, minimal, matches the intent of #586/#613/#587, and closes a real macOS/Windows edge case plus the in-repo footgun. I found no red flags in content or provenance-sensitive surfaces (deps, CI, build scripts, network endpoints). Recommendation: **approve and merge once CI passes**.

n13 (approved):
Nice - GTG

n13:
thank you for the contribution @exocognosis

exocognosis:
@n13 🙏

PR #640, opened 2026-08-05
V12 audit: fix referenda, scheduler, and frame-system findings

n13:
# V12 Security Audit — Remediation

Full accounting of **all 246 findings** in `export-findings.json` (auditor: V12): 23 triaged valid, 223 triaged invalid. Every valid finding is either fixed in this PR, fixed via #639 (merged into this branch), or documented not-fix below with reasoning — see the tables at the bottom of this description for the complete per-finding accounting.

**`spec_version` bumped 139 → 140** (runtime WASM changed).

> **Merge note:** `main` (#639) independently fixed three scheduler findings also addressed here. Resolved toward main's implementations: **#162523** (failed reschedule) keeps main's copy-first / no-op-on-failure design — this branch contributes its two extra regression tests; **#162526** (unaligned timestamp retries) is covered by main's fail-fast `InvalidRetryPeriod` validation at `set_retry` time — this branch's normalization approach was dropped; **#162524** landed identically in both (same `Retries::remove` line).

### pallets/referenda
- **`set_alarm` retry loop** — fixes **#161704** + **#162455** (HIGH). On a full scheduler agenda, alarms slide forward up to 16 blocks (mirroring `schedule_enactment`) instead of being silently dropped. A dropped alarm could permanently stall a referendum — and with `max_deciding = 1` and no sudo backstop, deadlock the entire governance lane, with no runtime-upgrade path out.
- **Ghost-queue fix** — **#161743**: when a referendum was evicted from a full `TrackQueue` by `force_insert_keep_right`, `in_queue` stayed `true` with no alarm, stranding it (timeout check is gated on `!in_queue`). The `NotQueued` service branch now clears the flag and marks the status dirty.
- **`nudge_referendum_approved` weight** — **#162501**: accounts for the locally-added 17-attempt enactment scheduling retry loop (both weight impls).

### pallets/scheduler
- **`service_task_base` weight charges `Retries::take`** (1r/1w, unconditional on every successful dispatch) — **#162534**. Also fixes the service-task benchmarks, which used a zero-limit `WeightMeter` and therefore never measured the success path.
- **Preimage request leak** — **#162453**: `bound` on a fresh oversized call already notes the hash (`Requested { count: 1 }`); the unconditional follow-up `request` bumped the count to 2 while every terminal path drops once, leaking the bytes permanently. The request is now skipped unless the hash was already requested before `bound` (keeps the user-noted `Unrequested` deposit flow balanced, and the multi-task-same-hash case).
- Regression tests for **#162523** (retry-config survival, named task not bricked after failed reschedule — the fix itself is main's, from #639).

### pallets/frame-system
- **`validate_unsigned` honors `check_version`** — **#162411**: pool validation for `apply_authorized_upgrade` hardcoded `false`; now matches the dispatch path (diverges from upstream, which has the same bug).
- **`set_heap_pages` range validation** — **#162546**: rejects values outside `64..=65536` (4 MiB executor-practical minimum … 4 GiB wasm32 linear-memory hard max) with new `Error::InvalidHeapPages`. Root-only, but a bad value bricks execution and this chain has no sudo backstop.

### Docs / comments
- `docs/TECH_REFERENDA_FEATURES.md` claimed `nudge_referendum` is "permissionless" — it is Root-only (`pallets/referenda/src/lib.rs:651`). Corrected; this misstated the exact recovery lever the HIGH findings depend on.
- `runtime/src/configs/mod.rs`: the `UndecidingTimeout` comment claimed queued referenda time out — they don't (upstream-intended). Comment corrected.

## Testing

Regression tests added across all three pallets; all suites green on the merged branch:
- `cargo test -p frame-system` — **76 passed**
- `cargo test -p pallet-referenda` — **42 passed**
- `cargo test -p pallet-scheduler` — **88 passed** (+ `--features runtime-benchmarks`: 105 passed, incl. benchmark tests for the reworked service-task benchmarks)
- `cargo check -p quantus-runtime` — clean (pre-existing warnings only)

## Deployment note

Runtime WASM changed → deploy via governance runtime upgrade per `docs/RUNTIME_UPDATE.md` (spec_version 140).

## Follow-up (out of scope, pre-existing)

Failure paths of `do_schedule_inner`/`do_schedule_named_inner` drop the preimage even when `bound` added no reference (hash already `Requested` by another live task), decrementing someone else's count. Pre-existing behavior, not touched here — worth a separate issue.

---

# Complete findings accounting

### Valid findings — full accounting (23)

| Finding | Severity | Title | Status / reasoning |
|---|---|---|---|
| #161704 | HIGH | Silent alarm-drop on scheduler failure stalls referendum lifecycle | Fixed in this PR — `set_alarm` slides forward up to 16 blocks on full agendas |
| #162455 | HIGH | Scheduler exhaustion can permanently retain a deciding slot | Fixed in this PR — same `set_alarm` retry fix as #161704 (shared root cause) |
| #162469 | HIGH | Queued referenda can bypass the undeciding deadline | Not fixing (code) — verbatim upstream-intended semantics (don't punish depositors for congestion); the misleading `UndecidingTimeout` comment was corrected in this PR |
| #162488 | HIGH | Account reaping resets nonce and enables transaction replay | Not fixing — known upstream design property; defense is mortal txs via `CheckEra` (present). All first-party clients are mortal (cli `.mortal(256)`, SDK `eraPeriod: 64`). Rejecting immortal txs breaks third-party tooling for marginal gain |
| #162534 | HIGH | Task weights omit retry and preimage-cleanup costs | Fixed in this PR — `service_task_base` charges `Retries::take` (1r/1w); benchmarks use real WeightMeter |
| #162540 | HIGH | Storage writes are not weighted by key and value size | Not fixing — `ensure_root` only; sudo removed; only TechCollective governance (members-only, 2h+3d delays) can call. No public spam path |
| #162546 | HIGH | Unvalidated heap-page setting can disable runtime execution | Fixed in this PR — `set_heap_pages` validates range 64..=65536 (`InvalidHeapPages`) |
| #162572 | HIGH | Runtime code installation weight ignores blob length | Not fixing — `set_code` is `ensure_root` only; `apply_authorized_upgrade` is gated to a governance-authorized blob hash. No public path |
| #162411 | MEDIUM | Unsigned validation for apply_authorized_upgrade skips required version check | Fixed in this PR — `validate_unsigned` honors `res.check_version` |
| #162459 | MEDIUM | Non-atomic legacy preimage migration destroys status | Not fixing — moot: chain launched with pallet-preimage 39.0.0 (`STORAGE_VERSION = 1` since 2022); v0 storage can never exist, so the legacy conversion path is unreachable |
| #162461 | MEDIUM | Schedule the preimage v0-to-v1 storage migration | Not fixing — moot for the same reason as #162459: on-chain version was 1 since genesis; nothing to migrate |
| #162541 | MEDIUM | Storage deletion weight omits key length | Not fixing — `kill_storage` is `ensure_root` only, same rationale as #162540 |
| #162542 | MEDIUM | Runtime code installation undercharges variable-length blobs | Not fixing — duplicate of #162572, same rationale |
| #162570 | MEDIUM | Storage-write weight omits key and value sizes | Not fixing — duplicate of #162540 (weight impl detail), same rationale |
| #161743 | LOW | Ghost-queued referenda can be stranded when no longer in the track queue | Fixed in this PR — `NotQueued` branch clears `in_queue` + marks dirty |
| #162232 | LOW | Runtime-benchmarks feature disables runtime-version checks | Not fixing — production artifacts never build with `runtime-benchmarks` (release workflow builds plain `--release`); fork is already stricter than upstream |
| #162453 | LOW | Lookup preimages retain an orphaned request after task cleanup | Fixed in this PR — skip redundant preimage request when `bound` just noted the hash |
| #162501 | LOW | Approved nudge weight omits enactment scheduling retries | Fixed in this PR — `nudge_referendum_approved` weight covers 17-attempt retry loop |
| #162503 | LOW | Make block-number-provider migration one-time and complete | Not fixing — dead code: `MigrateBlockNumberProvider` is not wired into the runtime `Migrations` tuple; upstream-identical; left untouched |
| #162523 | LOW | Failed rescheduling destroys tasks and leaves coupled state stale | Fixed via #639 (merged into this branch) — copy-first reschedule, failure is a complete no-op; this PR adds regression tests |
| #162524 | LOW | Terminally overweight tasks retain retry policies | Fixed via #639 (merged; identical `Retries::remove` in both PRs) |
| #162526 | LOW | Unaligned timestamp retries are never serviced | Fixed via #639 (merged) — fail-fast `InvalidRetryPeriod` rejects non-bucket-aligned periods at `set_retry` |
| #162571 | LOW | Storage deletion weight omits key length | Not fixing — duplicate of #162541 (weight impl detail), same rationale |

The 6 invalid-triaged findings flagged `needs-review` by the tool were manually reviewed as well:

| Finding | Title | Review conclusion |
|---|---|---|
| #162239 | set_storage and kill_storage accept unbounded item counts | Same family as #162540: `set_storage`/`kill_storage` are `ensure_root` only — no public path; governance self-limiting |
| #162472 | Bound multi-block migration steps by the remaining block budget | Not applicable to our config: the runtime `Migrations` tuple uses one-shot `on_runtime_upgrade` migrations, not multi-block stepped migrations |
| #162477 | Raw storage administration can bypass runtime upgrade controls | Same family as #162540: raw storage admin is `ensure_root` only — governance itself is the control |
| #162529 | Worst-case agenda weight gate prematurely defers scheduled tasks | Upstream-intended: the agenda weight gate (`MaximumWeight` = 80% of block) defers tasks to later blocks under load; standard scheduler behavior |
| #162538 | Unbounded event data enables unpriced proof growth and silent event loss | Events are not state and do not affect consensus weight; bounded by block-size limits on the extrinsic side. Accepted upstream property |
| #162545 | Report incomplete bounded prefix deletion | Informational/monitoring request; bounded prefix deletion was already hardened in #616/#617. No correctness issue |

### Invalid findings (223) — triaged invalid by the audit tool, verified in aggregate

| Finding | Severity | Title | Triage category |
|---|---|---|---|
| #161555 | LOW | Timestamp retry scheduling bypasses future-time validation | theoretical-attack-vector |
| #161561 | LOW | Agenda housekeeping weights undercount storage accesses | invalid-finding |
| #161568 | LOW | service_agendas_base undercharges IncompleteBlockSince write | invalid-finding |
| #161569 | LOW | service_timestamp_agendas_base undercounts timestamp housekeeping reads/writes | negligible-economic-impact |
| #161579 | LOW | Nonce consumption on dispatch failure is not documented as an invariant | code-quality-issue |
| #161584 | LOW | Nonce provides/requires tags and stale/future handling match intent | invalid-finding |
| #161585 | LOW | MAX_EXTRINSIC_DEPTH enforced before unchecked-to-checked conversion | invalid-finding |
| #161600 | LOW | Zero-sender rejection behaves as documented | invalid-finding |
| #161603 | LOW | Genesis hash binding enforced in signed implication | invalid-finding |
| #161604 | LOW | Era birth hash binding and AncientBirthBlock match intent | invalid-finding |
| #161607 | LOW | Account provider/sufficient gating of nonce admission matches intent | invalid-finding |
| #161629 | LOW | Order of event reset, upgrade hooks, System::initialize, and Mandatory weight registration is correct | invalid-finding |
| #161630 | LOW | runtime_upgraded() predicate matches documented spec-version/spec-name semantics | invalid-finding |
| #161631 | LOW | Unchecked downgrade/equal-version code install correctly suppresses upgrade hooks as documented | intended-behavior |
| #161641 | LOW | VERSION identity and upgrade ordering verified | invalid-finding |
| #161648 | LOW | CheckWeight extension weight omits storage I/O costs | invalid-finding |
| #161649 | LOW | CheckWeight extension weight omits storage I/O costs | known-issue |
| #161655 | LOW | Block weight and extrinsic index are not reset at finalization | user-operational-error |
| #161657 | LOW | Missing runtime header number re-verification on import | invalid-finding |
| #161665 | LOW | Pallet-level on_idle/on_poll nondeterminism is a latent chain-split risk | theoretical-attack-vector |
| #161666 | LOW | No panic-based Ok-with-mismatch path found in import execution | invalid-finding |
| #161679 | LOW | Signed payload can be decoupled from unsigned call in offchain helper | theoretical-attack-vector |
| #161680 | LOW | ForAll iterates over live keystore without snapshot or deduplication | user-operational-error |
| #161681 | LOW | Offchain submission pipeline returns opaque errors without rollback | code-quality-issue |
| #161693 | LOW | PreimageDeposit allows zero-footprint tickets and lacks stale-ticket guard | invalid-finding |
| #161696 | LOW | Fee policy wiring consistent with documentation | invalid-finding |
| #161697 | LOW | HighSecurityConfig whitelist exact and complete | invalid-finding |
| #161698 | LOW | NonWormholeAccounts exclusions match policy | invalid-finding |
| #161699 | LOW | Scheduler configuration matches documented limits | invalid-finding |
| #161702 | LOW | Executive type alias in docs omits configured migration tuple | code-quality-issue |
| #161706 | LOW | Vote-weight overflow for maximum-rank members in Linear/Geometric | deployment-configuration-error |
| #161707 | LOW | Unchecked vote-sum overflow in Tally::approval | deployment-configuration-error |
| #161708 | LOW | Unconditional status and alarm rewrite on every poll access | theoretical-attack-vector |
| #161709 | LOW | try_state_index panics on broken forward/inverse mapping | code-quality-issue |
| #161710 | LOW | try_state_members has weak per-rank check and overflow-prone accumulator | code-quality-issue |
| #161711 | LOW | remove_from_rank saturates a zero MemberCount instead of flagging corruption | code-quality-issue |
| #161712 | LOW | remove_member doc comment contradicts rank-0 removal | code-quality-issue |
| #161713 | LOW | Misleading comment on Curve::passing threshold comparison | code-quality-issue |
| #161734 | LOW | Approved referendum left permanently unscheduled after 17 consecutive full agendas | theoretical-attack-vector |
| #161740 | LOW | Exported MigrateBlockNumberProvider has an impossible generic bound | code-quality-issue |
| #161747 | LOW | Post-vote alarm calculation uses non-saturating addition | theoretical-attack-vector |
| #161748 | LOW | Manual nudge can leave stale scheduler alarm | theoretical-attack-vector |
| #161750 | LOW | cancel leaves stale metadata and preimage reference behind | false-invariant |
| #161752 | LOW | Technical referendum wiring matches intended authority model | invalid-finding |
| #161753 | LOW | Migration tuple does not include referendum migrations | invalid-finding |
| #161769 | LOW | Try-runtime migration tuple does not short-circuit on failure | informational-monitoring |
| #161773 | LOW | Wormhole v0→v1 migration non-idempotent counter re-seed | theoretical-attack-vector |
| #161780 | LOW | Pallet init/finalize hooks run ungated during MBM | theoretical-attack-vector |
| #161781 | LOW | Pre/post upgrade state bytes lack integrity check | invalid-finding |
| #161782 | LOW | StorageNoopGuard is a post-hoc assertion, not strong isolation | invalid-finding |
| #161783 | LOW | LastRuntimeUpgrade written before try-state checks | invalid-finding |
| #161790 | LOW | No overflow in RemovePallet/RemoveStorage weight arithmetic | invalid-finding |
| #161791 | LOW | Post-upgrade checks correctly tolerate bounded residual keys | intended-behavior |
| #161800 | LOW | Final release branch relies on a debug-only count assertion | theoretical-attack-vector |
| #161801 | LOW | Trait unrequest/unnote methods swallow release errors in release builds | code-quality-issue |
| #161821 | LOW | ensure_updated weight proof mismatches reserve/unreserve path | invalid-finding |
| #161824 | LOW | Preimage migration try-runtime check silently accepts data loss | theoretical-attack-vector |
| #161826 | LOW | Inclusion mode correctly restricts extrinsics during migrations | invalid-finding |
| #161827 | LOW | Inherent handoff and phase transitions are correct | invalid-finding |
| #161828 | LOW | Depth-limited decode is applied before signature check | invalid-finding |
| #161829 | LOW | Weight accounting and extrinsic recording are correct | invalid-finding |
| #161847 | LOW | ZkTreeRoot and ExtrinsicData cleanup rely on implicit cross-block overwriting and accurate counters | invalid-finding |
| #161852 | LOW | Unchecked addition in `inc_sufficients` creation test risks wraparound false creation | theoretical-attack-vector |
| #162112 | LOW | Saturating tally arithmetic masks vote-tally inconsistency | theoretical-attack-vector |
| #162113 | LOW | Fee waiver and replacement-fee logic align with Voting record | invalid-finding |
| #162121 | LOW | Verification: technical submitter must be a current member | invalid-finding |
| #162127 | LOW | Verification: TechReferenda instance and TechCollective::Polls wiring match | invalid-finding |
| #162128 | LOW | cleanup_poll fee waiver excludes successful zero-removal calls | code-quality-issue |
| #162175 | LOW | Stale block length and extrinsic count survive initialization | invalid-finding |
| #162178 | LOW | TxExtension includes three extra extensions beyond the eight-member pipeline | code-quality-issue |
| #162179 | LOW | on_idle gate collapses to ref-time-only when proof_size is u64::MAX | design-decision |
| #162181 | LOW | Proof-size weight dimension is intentionally uncapped by runtime configuration | design-decision |
| #162182 | LOW | RUNTIME_SURFACE.md lists Executive with empty migrations tuple | code-quality-issue |
| #162184 | LOW | final_checks verifies all required header commitments exactly | invalid-finding |
| #162185 | LOW | execute_block and finalize_block share identical hook and weight paths | invalid-finding |
| #162187 | LOW | Core::execute_block discards header; correctness is enforced by internal panics | external-dependency-behavior |
| #162188 | LOW | Core::execute_block discards header; correctness is enforced by internal panics | intended-behavior |
| #162189 | LOW | Malicious blocks can trigger deterministic import panic after full execution | theoretical-attack-vector |
| #162196 | LOW | Production and try-runtime upgrade paths differ in invariant coverage | false-invariant |
| #162197 | LOW | Try-runtime upgrade path skips multi-block migration stepping | theoretical-attack-vector |
| #162198 | LOW | Multi-block migration gate is only enforced in execute_block | invalid-finding |
| #162199 | LOW | Unvalidated prefix strings allow removal of wrong storage | theoretical-attack-vector |
| #162201 | LOW | Try-runtime guard only covers post-migration checks | false-invariant |
| #162203 | LOW | VersionedMigration relies on caller for StorageNoopGuard | invalid-finding |
| #162204 | LOW | Production upgrade path skips pre/post checks | intended-behavior |
| #162208 | LOW | SteppedMigration documentation typo: 'Nonce' instead of 'None' | code-quality-issue |
| #162228 | LOW | Signature-sender binding verified before dispatch | invalid-finding |
| #162239 | LOW | set_storage and kill_storage accept unbounded item counts | needs-review |
| #162240 | LOW | Reauthorization silently overwrites pending upgrade authorization | invalid-finding |
| #162256 | LOW | Root administration calls correctly enforce `ensure_root` | invalid-finding |
| #162259 | LOW | No replay risk after authorization is consumed | invalid-finding |
| #162260 | LOW | Hash computation timing is not a security vulnerability | theoretical-attack-vector |
| #162261 | LOW | Fee waiver does not bypass block weight limit | code-quality-issue |
| #162299 | LOW | Block-weight arithmetic uses non-saturating operations | theoretical-attack-vector |
| #162300 | LOW | check_inherents does not replay the validate_transaction state-write defect | invalid-finding |
| #162304 | LOW | Operational reserved weight uses non-saturating subtraction | invalid-finding |
| #162305 | LOW | Base weight validation uses non-saturating addition | invalid-finding |
| #162312 | LOW | Scheduler call disabling and migration set verified correct | invalid-finding |
| #162313 | LOW | fast-governance feature does not affect system/executive component | invalid-finding |
| #162374 | LOW | Reschedule can transfer mismatched-domain retry config | theoretical-attack-vector |
| #162390 | LOW | try_execute_block digest check truncates on mismatched lengths | theoretical-attack-vector |
| #162391 | LOW | Vote benchmark underestimates weight for vote changes | invalid-finding |
| #162392 | LOW | Benchmarks fall back to error path when no poll classes exist | theoretical-attack-vector |
| #162393 | LOW | cleanup_poll weight extrapolates beyond benchmarked range | theoretical-attack-vector |
| #162403 | LOW | High-security whitelist excludes wrapped and alternate reversible calls | intended-behavior |
| #162410 | LOW | StoredMap::try_mutate_exists can remove account with active consumers | invalid-finding |
| #162414 | LOW | Runtime Version Comment Contradicts Configuration | code-quality-issue |
| #162421 | LOW | Benchmark Origin Fails Membership Check | code-quality-issue |
| #162454 | LOW | Retain only live scheduler weight coverage | code-quality-issue |
| #162456 | LOW | Scheduled dispatches can exceed the reported weight budget | false-invariant |
| #162457 | LOW | Retry placement is charged at maximum agenda occupancy | theoretical-attack-vector |
| #162458 | LOW | V3 scheduler cancellation traits bypass task-origin privilege checks | theoretical-attack-vector |
| #162460 | LOW | Metadata references do not retain their preimages | informational-monitoring |
| #162462 | LOW | Saturating preimage request counter loses reference accounting | theoretical-attack-vector |
| #162463 | LOW | Preimage submission weight omits runtime reserve costs | invalid-finding |
| #162464 | LOW | Preimage weights are benchmarked against the wrong deposit path | known-issue |
| #162465 | LOW | Preimage deposits charge blob count as bytes | negligible-economic-impact |
| #162466 | LOW | Failed ticket updates can release the existing reservation | theoretical-attack-vector |
| #162467 | LOW | Stale public block-hash retention constant conflicts with active config | code-quality-issue |
| #162468 | LOW | System calls use uncalibrated fallback weights | theoretical-attack-vector |
| #162470 | LOW | Runtime upgrade weight is not bounded by the block budget | theoretical-attack-vector |
| #162471 | LOW | Lifecycle hooks can bypass the block-weight limit | false-invariant |
| #162472 | LOW | Bound multi-block migration steps by the remaining block budget | needs-review |
| #162473 | LOW | Reset aggregate extrinsic length before pool validation | invalid-finding |
| #162474 | LOW | Weight schedule undercharges reclamation storage accesses | theoretical-attack-vector |
| #162475 | LOW | Reject extrinsic lengths that exceed `u32` accounting bounds | theoretical-attack-vector |
| #162476 | LOW | Authorized origin drops the original signer identity | intended-behavior |
| #162477 | LOW | Raw storage administration can bypass runtime upgrade controls | needs-review |
| #162478 | LOW | Make inactive migration configuration explicit | code-quality-issue |
| #162479 | LOW | Event counter overflow silently omits later events | theoretical-attack-vector |
| #162480 | LOW | Unbounded event topics can create underpriced proof growth | theoretical-attack-vector |
| #162481 | LOW | Make block-hash pruning boundary explicit and tested | code-quality-issue |
| #162482 | LOW | Public consumer increment bypasses the configured reference limit | false-invariant |
| #162483 | LOW | Omitted transaction extensions for weight reclaim and authorization | invalid-finding |
| #162484 | LOW | Under-declared execution weight escapes block-weight accounting | intended-behavior |
| #162485 | LOW | Configure multi-block migration handlers | code-quality-issue |
| #162486 | LOW | Prevent signing transactions for nonexistent accounts | user-operational-error |
| #162487 | LOW | Authorized dispatch bypasses original-origin restrictions | invalid-finding |
| #162489 | LOW | Pool-valid future nonces can invalidate block import | theoretical-attack-vector |
| #162490 | LOW | Pool validation correctly defers extension state changes | invalid-finding |
| #162491 | LOW | Offchain nonce counter wraps after exhaustion | invalid-finding |
| #162492 | LOW | Authorized transaction helper lacks active authorization pipeline | theoretical-attack-vector |
| #162493 | LOW | Block builder bypasses inherent ordering and inclusion checks | user-operational-error |
| #162494 | LOW | Validation relies on host rollback for block-state writes | user-operational-error |
| #162495 | LOW | Upgrade rehearsal failures panic instead of returning diagnostics | intended-behavior |
| #162496 | LOW | Make migration ordering dependencies explicit and tested | code-quality-issue |
| #162497 | LOW | Failed migrations can repeatedly halt runtime upgrades | invalid-finding |
| #162498 | LOW | Transaction validation accepts an unverified parent hash | external-dependency-behavior |
| #162499 | LOW | Non-None origins bypass the transaction authorization gate | false-invariant |
| #162500 | LOW | Failed alarm replacement can permanently stall referenda | theoretical-attack-vector |
| #162502 | LOW | Live membership changes can bias active referendum support | known-issue |
| #162504 | LOW | Referenda weights are unvalidated for the configured queue bound | invalid-finding |
| #162505 | LOW | Test feature can remove production governance delays | deployment-configuration-error |
| #162506 | LOW | Executive and migration configuration is correct | invalid-finding |
| #162507 | LOW | Runtime upgrade migration order is correctly configured | invalid-finding |
| #162508 | LOW | Transaction signature verification is consistently enforced | invalid-finding |
| #162509 | LOW | Runtime version and migration invariants are correctly implemented | invalid-finding |
| #162510 | LOW | Transaction extension ordering and payload binding are correct | invalid-finding |
| #162511 | LOW | Block resource limits match the documented policy | design-decision |
| #162512 | LOW | Extrinsics-root hashing and import validation are correctly aligned | invalid-finding |
| #162513 | LOW | Shared finalization preserves builder/import lifecycle consistency | invalid-finding |
| #162514 | LOW | Mandatory extrinsics are correctly excluded from pool validation | invalid-finding |
| #162515 | LOW | Expected transaction-extension re-export surface | invalid-finding |
| #162516 | LOW | Governance adapter wiring enforces the intended technical lane | invalid-finding |
| #162517 | LOW | Flat collective rank and membership policy is correctly enforced | invalid-finding |
| #162518 | LOW | Linear vote weighting correctly handles the configured rank | code-quality-issue |
| #162519 | LOW | Technical referendum path enforces Root proposal origins | invalid-finding |
| #162520 | LOW | Caller-supplied parent hash has no scoped validation effect | invalid-finding |
| #162521 | LOW | Block length lacks an aggregate cross-class cap | invalid-finding |
| #162522 | LOW | Nonce and transaction-authentication checks are correctly enforced | invalid-finding |
| #162525 | LOW | Reject zero-count retry configurations | user-operational-error |
| #162527 | LOW | Same-bucket retries are erased by agenda writeback | theoretical-attack-vector |
| #162528 | LOW | Enforce per-task scheduler overhead against the weight meter | invalid-finding |
| #162529 | LOW | Worst-case agenda weight gate prematurely defers scheduled tasks | needs-review |
| #162530 | LOW | Ignored ticket-drop failures can orphan provider deposits | theoretical-attack-vector |
| #162531 | LOW | Provider ticket lifecycle contradicts documented request semantics | code-quality-issue |
| #162532 | LOW | Emit an event when the final preimage request is released | informational-monitoring |
| #162533 | LOW | Do not report a preimage before byte storage succeeds | theoretical-attack-vector |
| #162535 | LOW | Remove stale preimage deposit configuration | code-quality-issue |
| #162536 | LOW | Correct stale multisig weight-limit documentation | code-quality-issue |
| #162537 | LOW | Align multisig inner-call cap with active block-weight policy | invalid-finding |
| #162538 | LOW | Unbounded event data enables unpriced proof growth and silent event loss | needs-review |
| #162539 | LOW | Runtime scheduler, pallet indices, and migrations are correctly configured | invalid-finding |
| #162543 | LOW | Authorization deletion relies on an external rollback boundary | invalid-finding |
| #162544 | LOW | Installer errors do not consume authorized upgrades | invalid-finding |
| #162545 | LOW | Report incomplete bounded prefix deletion | needs-review |
| #162547 | LOW | Unchecked code replacement can bypass upgrade reconciliation | intended-behavior |
| #162548 | LOW | Missing extrinsic entries are silently committed as empty data | theoretical-attack-vector |
| #162549 | LOW | Unchecked weight registration can exceed the block budget | trust-assumption |
| #162550 | LOW | Authorized-origin guard requires explicit use by protected calls | intended-behavior |
| #162551 | LOW | Stale ZK tree root can be committed to subsequent headers | invalid-finding |
| #162552 | LOW | Clear extrinsic data entries outside the finalized range | invalid-finding |
| #162553 | LOW | Provider underflow can trigger a spurious account-death lifecycle | false-invariant |
| #162554 | LOW | Sufficient-reference underflow can trigger invalid account reaping | theoretical-attack-vector |
| #162555 | LOW | Unmatched consumer decrements are only logged | code-quality-issue |
| #162556 | LOW | Sufficient decrement can reap accounts with active consumers | theoretical-attack-vector |
| #162557 | LOW | Rejected checked upgrades consume their authorization | intended-behavior |
| #162558 | LOW | Detect gaps and clear stale extrinsic data | code-quality-issue |
| #162559 | LOW | Nonce replay safeguards are correctly enforced | invalid-finding |
| #162560 | LOW | Inherent ordering and block-boundary checks are enforced | invalid-finding |
| #162561 | LOW | Unchecked hook weight can exceed the block budget | false-invariant |
| #162562 | LOW | Meter post-extrinsic hooks against the block-weight budget | false-invariant |
| #162563 | LOW | Validation admits transactions forbidden during migrations | theoretical-attack-vector |
| #162564 | LOW | Mandatory and ordinary dispatch failures are handled as intended | invalid-finding |
| #162565 | LOW | Validate parent hash before overwriting prior hash state | external-dependency-behavior |
| #162566 | LOW | Block flowchart misorders migration and post-transaction hooks | code-quality-issue |
| #162567 | LOW | Initialize current block state before runtime-upgrade hooks | informational-monitoring |
| #162568 | LOW | Mandatory initialization weight can exceed the block limit | design-decision |
| #162569 | LOW | Single-extrinsic cap omits base execution weight | false-invariant |
| #162573 | LOW | Make base genesis and technical-collective seeding atomic | user-operational-error |
| #162574 | LOW | Reconcile offchain signing nonces with chain inclusion | invalid-finding |
| #162576 | LOW | Signer enumeration permits duplicate keys and unstable indexes | code-quality-issue |
| #162577 | LOW | Report incompatible offchain key types instead of skipping them | invalid-finding |
| #162579 | LOW | Unsynchronized offchain nonce tracking stalls signed dispatches | known-issue |
| #162580 | LOW | Unsigned signed-payload submissions lack enforced replay protection | theoretical-attack-vector |
| #162581 | LOW | Authorization validation trusts an unenforced weight declaration | theoretical-attack-vector |
| #162582 | LOW | Reconcile authorization refunds with declared weight | invalid-finding |
| #162583 | LOW | Runtime surface documentation is stale | code-quality-issue |
| #162584 | LOW | Report the actual amount deducted when slashing deposits | informational-monitoring |
| #162585 | LOW | Zero-limit cleanup discards the continuation cursor | theoretical-attack-vector |
| #162586 | LOW | Prefix-removal migrations undercharge traversal weight | invalid-finding |
| #162587 | LOW | Migration permanently deletes invalid legacy preimages and locks deposits | theoretical-attack-vector |
| #162979 | LOW | Offchain signing helpers lack domain separation, enabling cross-context signature replay | theoretical-attack-vector |

### DIFF
--- a/docs/TECH_REFERENDA_FEATURES.md
+++ b/docs/TECH_REFERENDA_FEATURES.md
@@ -54,7 +54,7 @@ Referenda's `Tally = pallet_ranked_collective::TallyOf<Runtime>` and
 | `submit` | Root or member | Create a referendum |
 | `place_decision_deposit` / `refund_decision_deposit` | signed | 1000 UNIT decision bond |
 | `refund_submission_deposit` | signed | 100 UNIT submission bond |
-| `nudge_referendum` | **permissionless** | Force the state machine to re-evaluate now (the "fast resolution" lever, §6) |
+| `nudge_referendum` | **Root** (dispatched via governance) | Force the state machine to re-evaluate now (§6) |
 | `cancel` | Root | Stop; refunds deposits |
 | `kill` | Root | Stop; slashes deposits (`Slash = ()` → burned) |
 | `one_fewer_deciding` | signed | Free a deciding slot |
@@ -157,7 +157,8 @@ Phases: `prepare_period` → **deciding** (≤ `decision_period`) → **confirmi
 (`confirm_period`) → enactment (`min_enactment_period`).
 
 There is no dedicated fast-resolve call. Early resolution comes from the
-**confirming** mechanism plus the permissionless **`nudge_referendum`** poke.
+**confirming** mechanism; the Root-only **`nudge_referendum`** (dispatched via
+governance) can additionally force a re-evaluation.
 Once both curves are satisfied the referendum enters confirming; if it stays
 passing for `confirm_period` it is approved **without waiting out the full
 decision period**:
@@ -181,8 +182,9 @@ support_needed.passing(x, tally.support(id)) && approval_needed.passing(x, tally
 
 If support drops below threshold during confirming, confirmation aborts
 (`ConfirmAborted`) and must restart — so it is not a one-way ratchet, which is
-what makes `confirm_period` a defense window. `nudge_referendum` lets anyone
-force a re-evaluation immediately instead of waiting for the scheduled alarm.
+what makes `confirm_period` a defense window. `nudge_referendum` (Root only, so
+in practice dispatched via governance) forces a re-evaluation immediately
+instead of waiting for the scheduled alarm.
 
 ---
 
--- a/frame/system-benchmarking/src/inner.rs
+++ b/frame/system-benchmarking/src/inner.rs
@@ -83,8 +83,10 @@ mod benchmarks {
 
 	#[benchmark]
 	fn set_heap_pages() -> Result<(), BenchmarkError> {
+		// V12 audit #162546: `set_heap_pages` now enforces a `64..=65536` range, so the
+		// benchmark must use a value within it.
 		#[extrinsic_call]
-		set_heap_pages(RawOrigin::Root, Default::default());
+		set_heap_pages(RawOrigin::Root, 64);
 
 		Ok(())
 	}
--- a/pallets/frame-system/src/lib.rs
+++ b/pallets/frame-system/src/lib.rs
@@ -730,6 +730,10 @@ pub mod pallet {
 		#[pallet::weight((T::SystemWeightInfo::set_heap_pages(), DispatchClass::Operational))]
 		pub fn set_heap_pages(origin: OriginFor<T>, pages: u64) -> DispatchResultWithPostInfo {
 			ensure_root(origin)?;
+			// V12 audit #162546: reject values that would prevent the executor from
+			// instantiating the runtime. 64 pages (4 MiB) is the practical executor minimum,
+			// 65536 pages (4 GiB) is the wasm32 linear-memory hard maximum.
+			ensure!((64..=65536).contains(&pages), Error::<T>::InvalidHeapPages);
 			storage::unhashed::put_raw(well_known_keys::HEAP_PAGES, &pages.encode());
 			Self::deposit_log(generic::DigestItem::RuntimeEnvironmentUpdated);
 			Ok(().into())
@@ -988,6 +992,8 @@ pub mod pallet {
 		NothingAuthorized,
 		/// The submitted code is not authorized.
 		Unauthorized,
+		/// The provided number of heap pages is outside the supported range of `64..=65536`.
+		InvalidHeapPages,
 	}
 
 	/// Exposed trait-generic origin type.
@@ -1165,7 +1171,10 @@ pub mod pallet {
 		fn validate_unsigned(source: TransactionSource, call: &Self::Call) -> TransactionValidity {
 			if let Call::apply_authorized_upgrade { ref code } = call {
 				if let Ok(res) = Self::validate_code_is_authorized(&code[..]) {
-					if Self::can_set_code(&code, false).is_ok() {
+					// V12 audit #162411: honor the authorization's `check_version` flag so pool
+					// validation rejects code that dispatch would reject (upstream hardcodes
+					// `false` here).
+					if Self::can_set_code(&code, res.check_version).is_ok() {
 						return Ok(ValidTransaction {
 							priority: u64::max_value(),
 							requires: Vec::new(),
--- a/pallets/frame-system/src/mock.rs
+++ b/pallets/frame-system/src/mock.rs
@@ -117,7 +117,7 @@ pub type SysEvent = frame_system::Event<Test>;
 
 /// A simple call, which one doesn't matter.
 pub const CALL: &<Test as Config>::RuntimeCall =
-	&RuntimeCall::System(frame_system::Call::set_heap_pages { pages: 0u64 });
+	&RuntimeCall::System(frame_system::Call::set_heap_pages { pages: 64u64 });
 
 /// Create new externalities for `System` module tests.
 pub fn new_test_ext() -> sp_io::TestExternalities {
--- a/pallets/frame-system/src/tests.rs
+++ b/pallets/frame-system/src/tests.rs
@@ -691,6 +691,70 @@ fn set_code_drains_remaining_block_weight() {
 	});
 }
 
+#[test]
+fn validate_unsigned_apply_authorized_upgrade_honors_check_version() {
+	struct ReadRuntimeVersion(Vec<u8>);
+
+	impl sp_core::traits::ReadRuntimeVersion for ReadRuntimeVersion {
+		fn read_runtime_version(
+			&self,
+			_wasm_code: &[u8],
+			_ext: &mut dyn sp_externalities::Externalities,
+		) -> Result<Vec<u8>, String> {
+			Ok(self.0.clone())
+		}
+	}
+
+	let code = vec![1, 2, 3, 4];
+	let code_hash = BlakeTwo256::hash(&code);
+	let call = Call::<Test>::apply_authorized_upgrade { code };
+
+	// The mock runtime runs `spec_name: "test"` at `spec_version: 1`, so a candidate that
+	// does not increase the spec version fails the version check.
+	let bad_version =
+		RuntimeVersion { spec_name: "test".into(), spec_version: 1, ..Default::default() };
+	let good_version =
+		RuntimeVersion { spec_name: "test".into(), spec_version: 2, ..Default::default() };
+
+	let test_data = vec![
+		// The authorization requires the version check and it fails -> invalid.
+		(bad_version.clone(), true, false),
+		// Same failing version, but the authorization skips the check -> valid.
+		(bad_version, false, true),
+		// The authorization requires the version check and it passes -> valid.
+		(good_version, true, true),
+	];
+
+	for (version, check_version, expect_valid) in test_data.into_iter() {
+		let read_runtime_version = ReadRuntimeVersion(version.encode());
+
+		let mut ext = new_test_ext();
+		ext.register_extension(sp_core::traits::ReadRuntimeVersionExt::new(read_runtime_version));
+		ext.execute_with(|| {
+			if check_version {
+				assert_ok!(System::authorize_upgrade(RawOrigin::Root.into(), code_hash));
+			} else {
+				assert_ok!(System::authorize_upgrade_without_checks(
+					RawOrigin::Root.into(),
+					code_hash
+				));
+			}
+
+			let res = <System as sp_runtime::traits::ValidateUnsigned>::validate_unsigned(
+				TransactionSource::External,
+				&call,
+			);
+
+			if expect_valid {
+				let valid = res.expect("transaction should be valid");
+				assert_eq!(valid.provides, vec![code_hash.encode()]);
+			} else {
+				assert_eq!(res, Err(InvalidTransaction::Call.into()));
+			}
+		});
+	}
+}
+
 fn assert_runtime_updated_digest(num: usize) {
 	assert_eq!(
 		System::digest()
@@ -750,8 +814,30 @@ fn runtime_updated_digest_emitted_when_heap_pages_changed() {
 	new_test_ext().execute_with(|| {
 		System::reset_events();
 		System::initialize(&1, &[0u8; 32].into(), &Default::default());
-		System::set_heap_pages(RawOrigin::Root.into(), 5).unwrap();
+		System::set_heap_pages(RawOrigin::Root.into(), 64).unwrap();
 		assert_runtime_updated_digest(1);
+	});
+}
+
+#[test]
+fn set_heap_pages_validates_range() {
+	new_test_ext().execute_with(|| {
+		System::reset_events();
+		System::initialize(&1, &[0u8; 32].into(), &Default::default());
+
+		// V12 audit #162546: values below the 4 MiB executor minimum and above the 4 GiB
+		// wasm32 linear-memory maximum are rejected.
+		for pages in [0u64, 63, 65537] {
+			assert_noop!(
+				System::set_heap_pages(RawOrigin::Root.into(), pages),
+				Error::<Test>::InvalidHeapPages
+			);
+		}
+
+		// Both bounds of the allowed range are accepted and still emit the digest item.
+		assert_ok!(System::set_heap_pages(RawOrigin::Root.into(), 64));
+		assert_ok!(System::set_heap_pages(RawOrigin::Root.into(), 65536));
+		assert_runtime_updated_digest(2);
 	});
 }
 
--- a/pallets/referenda/src/branch.rs
+++ b/pallets/referenda/src/branch.rs
@@ -18,8 +18,24 @@
 //! Helpers for managing the different weights in various algorithmic branches.
 
 use super::Config;
-use crate::weights::WeightInfo;
-use frame_support::weights::Weight;
+use crate::{weights::WeightInfo, Pallet};
+use frame_support::{traits::Get, weights::Weight};
+
+/// Worst-case overhead of the `set_alarm` full-agenda retry loop (V12 audit #161704 /
+/// #162455): every failed `T::Scheduler::schedule` attempt reads the candidate block's
+/// `Scheduler::Agenda` before sliding forward one block. Nothing is written until an
+/// attempt succeeds (and `Preimages::drop` of the inline nudge call is storage-free), and
+/// the succeeding attempt's cost is already part of the benchmarked base weight, so only
+/// the extra attempts are charged here.
+///
+/// This is added to every branch that can (re)schedule a referendum alarm or defer
+/// `one_fewer_deciding` through `set_alarm`. Kept in this hand-written module so it
+/// survives regeneration of the benchmarked `weights.rs`.
+pub fn alarm_retry_weight<T: Config<I>, I: 'static>() -> Weight {
+	T::DbWeight::get()
+		.reads(1)
+		.saturating_mul(Pallet::<T, I>::MAX_ALARM_SCHEDULE_RETRIES as u64)
+}
 
 /// Branches within the `begin_deciding` function.
 pub enum BeginDecidingBranch {
@@ -62,7 +78,7 @@ impl ServiceBranch {
 	/// Return the weight of the `nudge` function when it takes the branch denoted by `self`.
 	pub fn weight_of_nudge<T: Config<I>, I: 'static>(self) -> frame_support::weights::Weight {
 		use ServiceBranch::*;
-		match self {
+		let base = match self {
 			NoDeposit => T::WeightInfo::nudge_referendum_no_deposit(),
 			Preparing => T::WeightInfo::nudge_referendum_preparing(),
 			Queued => T::WeightInfo::nudge_referendum_queued(),
@@ -78,6 +94,24 @@ impl ServiceBranch {
 			Approved => T::WeightInfo::nudge_referendum_approved(),
 			Rejected => T::WeightInfo::nudge_referendum_rejected(),
 			TimedOut | Fail => T::WeightInfo::nudge_referendum_timed_out(),
+		};
+		match self {
+			// Branches that (re)schedule the referendum's alarm, or defer
+			// `one_fewer_deciding` via `set_alarm` (`Approved`/`Rejected`), bear the
+			// worst-case retry overhead.
+			NoDeposit |
+			Preparing |
+			NotQueued |
+			BeginDecidingPassing |
+			BeginDecidingFailing |
+			BeginConfirming |
+			ContinueConfirming |
+			EndConfirming |
+			ContinueNotConfirming |
+			Approved |
+			Rejected => base.saturating_add(alarm_retry_weight::<T, I>()),
+			// These branches leave the referendum without an alarm (or only cancel one).
+			Queued | RequeuedInsertion | RequeuedSlide | TimedOut | Fail => base,
 		}
 	}
 
@@ -99,6 +133,7 @@ impl ServiceBranch {
 			.max(T::WeightInfo::nudge_referendum_approved())
 			.max(T::WeightInfo::nudge_referendum_rejected())
 			.max(T::WeightInfo::nudge_referendum_timed_out())
+			.saturating_add(alarm_retry_weight::<T, I>())
 	}
 
 	/// Return the weight of the `place_decision_deposit` function when it takes the branch denoted
@@ -108,11 +143,18 @@ impl ServiceBranch {
 	) -> Option<frame_support::weights::Weight> {
 		use ServiceBranch::*;
 		let ref_time_weight = match self {
-			Preparing => T::WeightInfo::place_decision_deposit_preparing(),
+			// `Preparing`, `NotQueued` and `BeginDeciding*` (re)schedule the referendum's
+			// alarm and so bear the worst-case `set_alarm` retry overhead; `Queued` leaves
+			// the referendum alarm-less.
+			Preparing => T::WeightInfo::place_decision_deposit_preparing()
+				.saturating_add(alarm_retry_weight::<T, I>()),
 			Queued => T::WeightInfo::place_decision_deposit_queued(),
-			NotQueued => T::WeightInfo::place_decision_deposit_not_queued(),
-			BeginDecidingPassing => T::WeightInfo::place_decision_deposit_passing(),
-			BeginDecidingFailing => T::WeightInfo::place_decision_deposit_failing(),
+			NotQueued => T::WeightInfo::place_decision_deposit_not_queued()
+				.saturating_add(alarm_retry_weight::<T, I>()),
+			BeginDecidingPassing => T::WeightInfo::place_decision_deposit_passing()
+				.saturating_add(alarm_retry_weight::<T, I>()),
+			BeginDecidingFailing => T::WeightInfo::place_decision_deposit_failing()
+				.saturating_add(alarm_retry_weight::<T, I>()),
 			BeginConfirming |
 			ContinueConfirming |
 			EndConfirming |
@@ -137,6 +179,7 @@ impl ServiceBranch {
 			.max(T::WeightInfo::place_decision_deposit_not_queued())
 			.max(T::WeightInfo::place_decision_deposit_passing())
 			.max(T::WeightInfo::place_decision_deposit_failing())
+			.saturating_add(alarm_retry_weight::<T, I>())
 	}
 }
 
@@ -165,8 +208,12 @@ impl OneFewerDecidingBranch {
 		use OneFewerDecidingBranch::*;
 		match self {
 			QueueEmpty => T::WeightInfo::one_fewer_deciding_queue_empty(),
-			BeginDecidingPassing => T::WeightInfo::one_fewer_deciding_passing(),
-			BeginDecidingFailing => T::WeightInfo::one_fewer_deciding_failing(),
+			// `begin_deciding` schedules the promoted referendum's alarm, bearing the
+			// worst-case `set_alarm` retry overhead.
+			BeginDecidingPassing => T::WeightInfo::one_fewer_deciding_passing()
+				.saturating_add(alarm_retry_weight::<T, I>()),
+			BeginDecidingFailing => T::WeightInfo::one_fewer_deciding_failing()
+				.saturating_add(alarm_retry_weight::<T, I>()),
 		}
 	}
 
@@ -176,5 +223,6 @@ impl OneFewerDecidingBranch {
 			.max(T::WeightInfo::one_fewer_deciding_queue_empty())
 			.max(T::WeightInfo::one_fewer_deciding_passing())
 			.max(T::WeightInfo::one_fewer_deciding_failing())
+			.saturating_add(alarm_retry_weight::<T, I>())
 	}
 }
--- a/pallets/referenda/src/lib.rs
+++ b/pallets/referenda/src/lib.rs
@@ -916,6 +916,12 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 	/// old permissionless pre-fill.
 	const MAX_ENACTMENT_SCHEDULE_RETRIES: u32 = 16;
 
+	/// How many subsequent blocks to try if the preferred alarm block's agenda is full.
+	///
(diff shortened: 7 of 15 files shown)


n13 (commented):
Reviewed every fix against the code (not just the diff) and re-ran the suites locally: `frame-system` 76 ✓, `pallet-referenda` 42 ✓, `pallet-scheduler` 84 ✓ and 101 ✓ with `runtime-benchmarks` — all matching the PR's claims.

**Verified correct** (the load-bearing claims I checked in the source):

- `set_alarm` retry loop: the actually-scheduled block is returned and stored in `status.alarm`; alarm calls (`nudge_referendum`, `one_fewer_deciding`) are always inline-bounded, so the failed attempts hitting `do_schedule_inner`'s drop-on-failure are no-ops (`Preimages::drop` on `Inline` does nothing) — no preimage corruption from retries. Runtime `AlarmInterval = 1`, so the +1-block slide keeps alignment trivially.
- Ghost-queue fix: an evicted referendum whose ayes sort to position 0 hits the `NotQueued` arm; the flag clear + `dirty` is right, and if `now >= timeout` it times out in the same service pass (the check at `pallets/referenda/src/lib.rs:1244` sees the cleared flag).
- Preimage request leak (#162453): checked against pallet-preimage semantics — `note(None)` on a fresh hash creates `Requested{count:1}` but does **not** bump an already-requested count, so `lookup_already_requested` (computed pre-`bound`, same hashing as preimage) picks exactly the cases where the task needs its own `request`. Single-task, multi-task-same-hash, and user-noted-`Unrequested` flows all balance. Keeping the v3 shims unconditional is correct since those callers bound the call themselves.
- Failed-reschedule restore (#162523): `place_task` has no partial side effects before failing (`push_to_agenda` fails first; `Lookup` write and `Scheduled` event only on success), and `restore_task` exactly reconstructs the slot even after `cleanup_agenda` truncation. Named `Lookup` stays valid as claimed.
- `Retries` on `PermanentlyOverweight` (#162524) uses the same key as the `CallUnavailable` arm; `service_task_base` weight bump matches the unconditional `Retries::take` on the success path (`pallets/scheduler/src/lib.rs:1346`), and the zero-limit-meter benchmark fix is real (the old benchmark never measured the success path).
- `schedule_retry` normalization (#162526) matches `resolve_time`'s strict round-up convention (exact boundaries advance a bucket); the test's 35000→40000 expectation is consistent.
- `validate_unsigned` now mirrors the dispatch path at `pallets/frame-system/src/lib.rs:905`; the `ReadRuntimeVersionExt` test covers all three cases. `set_heap_pages` bounds are sane, and the mock `CALL` change (0→64) doesn't drift test weights (u64 is fixed-width).
- Both weight edits land in `nudge_referendum_approved` in both impls; docs fixes match the code (`nudge_referendum` is `ensure_root`, `pallets/referenda/src/lib.rs:651`).

**Non-blocking findings** (most significant first):

1. **Spurious `Canceled` event on the restored-reschedule path** — `pallets/scheduler/src/lib.rs:901`/`1024`: `Event::Canceled` is emitted *before* `place_task`; when placement fails and `restore_task` (lines 910/1035) puts the task back, the event stands with no compensating `Scheduled` event, so indexers tracking agenda state via events will believe the task is gone while it still executes later. Upstream has the same ordering, but there the task really was dropped, so the event was accurate. Suggest emitting `Canceled` only after `place_task` succeeds. Only reachable via the v3 shims today, hence non-blocking.

2. **Ghost referendum still needs two Root nudges to fully recover** — `pallets/referenda/src/lib.rs:1217`: after the `NotQueued` arm clears `in_queue`, `alarm` stays `max_value`, so the referendum is stored flag-clean but alarm-less; a second nudge (routing through `ready_for_deciding`) or the timeout check on a later nudge is needed. Setting `alarm = timeout` in that arm (mirroring the `NoDeposit` branch at line 1239) would make the ghost state self-heal at the undeciding timeout with no governance intervention — strictly better given nudge is Root-only and this fork has no sudo.

3. **The new alarm retry loop is unweighed** — the #162501 reasoning applies to it too: every `ensure_alarm_at`/`set_alarm` path (submit, place_decision_deposit, all nudge branches) can now perform up to 17 `Scheduler::Agenda` reads under full-agenda congestion, uncharged. Magnitude is small on a standalone chain (~17 reads ≈ 0.4 ms ref-time worst case), but for consistency the next weight pass should add a conservative bump to the alarm-setting paths like the enactment loop got.

4. **Follow-up issue deserves priority** — the drop-on-failure in `do_schedule_named_inner` interacts badly with the *pre-existing* enactment retry loop when the proposal is a `Lookup`: each failed attempt in the loop drops the same hash once, so up to 17 decrements against a count the loop doesn't own — the first can zero the count and delete the preimage bytes, after which even a *successful* later placement enacts nothing (`CallUnavailable`). Your out-of-scope note already covers the single-drop shape; the retry loop amplifies it, so the separate issue is worth filing promptly.

5. *Informational*: for generic configs with `AlarmInterval > 1`, slid alarms break interval alignment and `ensure_alarm_at`'s `when != alarm` comparison will cancel/re-schedule a slid alarm on every service while congestion persists (bounded churn, correct outcome). Irrelevant for this runtime (`AlarmInterval = 1`), maybe worth a doc line in the pallet.

**Verdict: Approve.** All 11 fixes are correct as implemented, the won't-fix rationales check out where I verified them (v3-shim reachability, `AlarmInterval`, root-only lanes), test coverage is genuinely good (each fix has a regression test that fails without it), and the findings above are improvements, not defects.

n13:
**Merge with `main` (#639) resolved** — `ff026f43`. #639 had independently fixed three of the same scheduler findings, so conflicts were resolved toward main's cleaner designs:

- **#162523** (failed reschedule destroys task) — kept main's copy-first / no-op-on-failure approach; dropped this branch's `restore_task` helper. This branch's two regression tests stay: they pass under both implementations and add coverage main lacks (retry-config survival, named task not bricked + subsequent successful reschedule).
- **#162524** (PermanentlyOverweight retains `Retries`) — both sides landed the identical `Retries::remove` line; only the comment conflicted.
- **#162526** (unaligned timestamp retries) — superseded by main's fail-fast `InvalidRetryPeriod` validation at `set_retry` time; dropped this branch's `schedule_retry` normalization and its test.
- This branch's remaining fixes (referenda `set_alarm` retry, ghost-queue flag, `nudge_referendum_approved` weight, `service_task_base` weight + benchmark meter, conditional preimage request, `validate_unsigned` `check_version`, heap-pages range) merged cleanly, untouched.

Verified on the merged tree: `pallet-scheduler` 88 + `pallet-referenda` 42 + `frame-system` 76 tests pass, scheduler benchmark-feature tests 105 pass, `cargo check -p quantus-runtime` clean.

n13:
All three additional commits are good - checked out ✅

illuzen (approved):

PR #653, opened 2026-08-09
fix: clear dilithium clippy denies and mirror publish lint gates in CI

illuzen:
Fix empty-line doc and large_enum_variant lints that failed dilithium publish under -D warnings, and run the same strict clippy on published crates in main CI.


n13 (approved):
Small, well-targeted PR: it clears the three clippy denies that are currently breaking the dilithium publish workflow, and adds the same gate to main CI so the next break is caught on the PR that causes it rather than at release time. I reviewed it in a worktree at `ead3a983` and verified the claims by running the commands rather than reading them. No blocking issues.

## Verified

**The premise is real — the publish workflow is red on `main` today.** Reverting just `primitives/dilithium-crypto/src/types.rs` and re-running the publish gate reproduces exactly three failures: one `empty_line_after_doc_comments` and two `large_enum_variant`. With the PR's changes the gate passes clean, and `cargo test --package qp-dilithium-crypto` is 18/18.

**The CI step is a byte-exact mirror, not an approximation.** Both new commands match the publish workflows character for character (`dilithium-publish.yml:96`, `miner-api-publish.yml:96`) — same `--locked`, same `--no-deps`, same `-- -D warnings`, and deliberately no `SKIP_WASM_BUILD=1`, matching how the publish job actually builds. I ran both locally: dilithium clean, miner-api exits 0. The step is also correctly placed after `rustup component add clippy`.

**The gated set is complete.** `cargo publish` appears exactly twice in `.github/workflows/` (`dilithium-publish.yml:121`, `miner-api-publish.yml:121`), so these two packages are the entire crates.io surface this repo has. Nothing is left ungated.

**The doc-comment change fixes a real docs bug, not just a lint.** This is the part worth calling out, because the diff reads like pure lint appeasement and it isn't. That floating `///` block was separated from `WrappedPublicBytes` by a blank line but still bound to it as an outer attribute, so rustdoc attached crate-level prose to the struct. Built the docs both ways to confirm: pre-fix, `types/struct.WrappedPublicBytes.html` and the module index both carry "Resonance Crypto Types", and the struct's one-line summary in `types/index.html` is that text instead of its own. Post-fix the index correctly reads "Wrapper around Substrate's PublicBytes to provide Dilithium-specific implementations". Converting to `//` was the right call.

**The `large_enum_variant` rationale checks out.** `Box<T>` does delegate its `Encode`/`Decode` impls and `TypeInfo` is transparent, so boxing genuinely would preserve both the SCALE encoding and the metadata — which means the comment is right that the only reason to decline it is runtime cost, not wire compatibility. And `~1.6k existing lints` is fair: I measured 1478 warnings on a workspace clippy run.

## Minor notes

**1. The rationale comments describe the wrong property** — `types.rs:56`, `types.rs:78`. Both say "Variants are intentionally large". `large_enum_variant` fires on the size *difference between* variants (Dilithium87 at 4627 bytes vs Dilithium65 at 3309), not on absolute size; an enum whose variants were all equally huge wouldn't trip it. The conclusion is unaffected, it's the framing that's off by one step.

**2. There's a stronger argument for the allow than the one given** — same two sites. `qp-dilithium-crypto` is published, and `DilithiumSignatureScheme` is the runtime's `Signature` type, so these enums are public API of a released crate (v0.5.0). Changing `Dilithium87(Dilithium87Public)` to `Dilithium87(Box<Dilithium87Public>)` is a source-breaking change for every downstream consumer *even though* the encoding is stable. That's a harder constraint than heap traffic and worth a few words in the comment, since it's the reason this can't simply be revisited later.

**3. The mirror covers clippy only, so publish can still surprise you** — `ci.yml:102`. The publish workflows also run `cargo build --release` and `cargo test --release` per package (`dilithium-publish.yml:90,93`), while CI tests the workspace in dev profile with `--features quantus-runtime/fast-governance`. A release-profile or standalone-feature-resolution breakage in these two crates would still show up first at release. Separately, neither the publish gate nor this mirror passes `--all-targets`, so lints in those crates' own test code stay ungated on both sides — worth adding on both if you want the property to be airtight. Follow-up, not this PR.

**4. Cost note, not a defect** — `ci.yml:102`. Using `--package` instead of `--workspace`, and dropping `SKIP_WASM_BUILD=1`, means this step resolves features for a standalone graph and can't reuse the preceding workspace clippy's artifacts. That's precisely what makes it faithful to how publish builds, so it's the right trade; just be aware the step pays for its own dependency compile. The Analysis job came in at 4m on this run, so in practice it's cheap.

Verdict: **APPROVE**. The fix is complete and minimal, the CI gate is an exact mirror of the release gate over the full set of published crates, and the doc change quietly repairs misattributed rustdoc on a published crate. The notes above are all wording or follow-up scope.

n13:
Kept digging after posting the review above. Verdict is unchanged — **still APPROVE**, nothing here is blocking — but four things turned up, including one that corrects something I got wrong.

**Correction to my own review: the doc block now renders nowhere.** I praised the `///` → `//` conversion for fixing the misattribution, and that part is right — but I checked only that the text stopped landing on the wrong item, not where it landed instead. `primitives/dilithium-crypto/src/lib.rs:1` has no crate-level docs at all, so after this change the published crate's docs.rs page has no overview of the ML-DSA-87/65 layout. `clippy::empty_line_after_doc_comments` suggests the `//!` form precisely for this case: moving those three lines to the top of `lib.rs` as `//!` clears the lint *and* keeps the documentation, instead of trading one for the other.

**The repo already has this gate, and it's stronger — it just doesn't run on PRs.** `pm-quality-check.yml:113` runs `SKIP_WASM_BUILD=1 cargo clippy --locked --package <name> --all-targets --all-features -- -D warnings` across a crate matrix that already lists `miner-api` (`:39`) and `primitives/dilithium-crypto` (`:48`), alongside fmt/doc/build/test checks. It's `workflow_dispatch`-only, so it never fires on a PR. Wiring that workflow into PR CI for the published crates — or extracting it as a `workflow_call` — would be strictly stronger than the hand-rolled subset here and would leave one source of truth instead of a third copy of the command to keep in sync. As it stands, the strict gate now lives verbatim in three files with nothing enforcing the match, which is the same class of drift the PR set out to prevent.

**`-D warnings` does not restore the workspace's own denies.** `Cargo.toml:66-72` declares `panic`, `unwrap_used` and `expect_used` as `"deny"`, but only `runtime/`, `client/network/`, `client/service/` and `client/transaction-pool/` carry `[lints] workspace = true`. `primitives/dilithium-crypto/Cargo.toml` has no `[lints]` section, so those denies never apply to it — `scheme_macro.rs:102`, `:108` and `:365` keep `.expect(...)` on the signature/verify path with no lint pressure. Given the stated aim of keeping published crates strict, adding the opt-in to both published crates is the change that actually delivers it; `-D warnings` alone doesn't.

**Two smaller ones.** `--package` without `--all-features` builds `default = ["std"]`, so the crate's `#[cfg(not(feature = "std"))]` paths are never linted even though every runtime consumer takes it with `default-features = false` — `--all-features`, as pm-quality-check already passes, covers that. And the new step sits ahead of `Run doc` in the same job under the default `if: success()`, so a `-D warnings` hit now hides all rustdoc breakage for that run; moving it after the doc step, or marking the doc step `if: always()`, keeps both signals in one round-trip.

None of this argues against merging as-is — the immediate breakage is fixed and the gate works. It's about where the durable version of this gate should live.

PR #654, opened 2026-08-09
V12 Vesting

illuzen:
# Security review fixes: governance, high-security accounts, and wormhole proof-weight accounting

This PR works through the security review findings item by item. Every fix follows the same pattern: the finding was first reproduced with a failing (red) test, the root cause was fixed, and the test now pins the behavior. `spec_version` is bumped to 143. No storage migrations are required.

## Governance

- **Enforce the tech-collective member floor on Root removals** (`9a9f35de`): Root could remove ranked-collective members below the floor that keeps the collective functional; the floor is now enforced on every removal path.
- **Pin referendum proposal preimages for their whole lifecycle** (`1f678a3e`): a referendum's proposal preimage could be unnoted while the referendum was live, breaking enactment. Preimages are now requested/pinned for the referendum's full lifetime.
- **Reconcile ongoing referendum tallies when a member leaves** (`3c5de631`): removing, exchanging, or demoting a collective member left their votes in ongoing poll tallies, inflating support and potentially flipping outcomes. Membership transitions now remove the member's votes from every ongoing poll.
- **Bound active referenda globally at submission** (`94daae7e`): nothing limited the total number of ongoing referenda, allowing state/scheduler exhaustion. New `MaxActive` config + `ActiveReferendaCount` storage enforce an admission bound in `submit`, with the counter decremented on every terminal transition.

## High-security accounts and multisig

- **Re-check high-security policy on every nested utility dispatch** (`fa93e635`): the whitelist check applied only to the outer call, so blocked calls could be smuggled through `Utility` wrappers. The policy now re-applies recursively.
- **Remove the unbounded single-block multisig v0→v1 migration** (`ee39b8c5`): the migration iterated all multisigs in one block and could brick a chain with a large set; replaced per the review's recommendation.

## Wormhole proof-recording weight model

The extension that records transfer proofs (ZK-tree leaf inserts) charged weight that could diverge from the work actually performed, in both directions. This PR moves the whole model to a simple, boundary-proof shape:

- **Flat leaf-insert pricing at the circuit depth ceiling** (`75260d0f`): insert costs were derived from the live tree depth, which repeatedly under-reserved when a batch crossed capacity boundaries. Inserts are now priced flat at `CIRCUIT_MAX_TREE_DEPTH = 16` (`INSERT_LEAF_*` constants), trading a small overcharge on a young tree for the removal of an entire class of metering bugs. All callers (vesting, mining-rewards, reversible-transfers, wormhole, and the extension) share the constants.
- **Fee-charge proof inserts for `ReversibleTransfers::{cancel, recover_funds}` and `Recovery::close_recovery`** (`6a37bbcc`): these calls emit `TransferOnHold`/`ReserveRepatriated` events that the scanner turns into proof inserts, but the static classifier charged them zero — the insert cost fell on block capacity instead of the transaction fee. They are now statically counted (`recover_funds` at the `MaxPendingPerAccount + 1` worst case).
- **Charge the event deposits proof recording performs** (`92577212`): each recorded proof deposits two events (`LeafInserted` + `NativeTransferred`/`AssetTransferred`) whose `EventCount`/`Events` work no mechanism charged. The per-transfer reservation now includes it, and a test pins the events-per-proof multiplier.
- **Document why pot-touching transfers stay statically counted** (`14b204db`): the review flagged the overcharge on transfers into the vesting pot; it is an accepted, documented trade-off (endpoint resolution would put a `Lookup` on the hottest call path), and pot-as-source is unreachable through the extension.

## Vesting

- **Roll back payouts when the proof recorder drops the credit** (`ecee6893`): `pay_out` ignored the recorder's boolean, so a dropped credit would still finalize the payout — advancing `claimed` or deleting the schedule with no wormhole leaf and no way to retry. A dropped credit now errors (`PayoutProofNotRecorded`) and rolls the transfer back, keeping the schedule retryable. Not reachable with the current Wormhole recorder; this hardens the generic trait boundary.
- **Apply the 90-day cliff to Bob's second testnet schedule** (`aa5be5c4`): the preset tuple used the start timestamp as its cliff, so the grant accrued from day 0 in the `dev` and `heisenberg` presets. Fixed, with a preset-level regression test asserting every shipped schedule locks for exactly `start + 90 days`.
- **Non-blocking review polish** (`0b4ce12d`, ported from `fix/vesting-review-polish` and adapted to flat pricing): PoV charge in the extension's per-transfer weight, per-test reset of the vesting mock's thread-local config, `settle` extraction, lazy pot derivation in the event scan, const-asserts tying the payout quantum to the ZK leaf scale factor and max supply to the leaf's u32 ceiling, and `MAX_SUPPLY`/`MILLIS_PER_DAY` dedup.

## Test plan

- [x] Red test written for every code fix, confirmed failing before and passing after
- [x] `pallet-vesting`: 57 tests green
- [x] `pallet-referenda`, `pallet-ranked-collective`, `pallet-zk-tree`: suites green
- [x] `quantus-runtime --lib`: 43 tests green (extension lifecycle, presets, genesis)
- [x] Governance integration tests (`runtime/tests`): green


n13 (changes requested):
## Review — V12 vesting / security-review fixes (spec 143)

Reviewed the full diff (29 files, +1502/−585) at head `b1cb0cd0` against a fan-out of finder angles plus per-candidate verification. The individual fixes are, in isolation, well-targeted and thoroughly tested — the governance, high-security-recursion, and flat-pricing changes each do what they claim. What blocks the PR is a **missing runtime-upgrade migration** for the two new pieces of referenda state, which turns two of the fixes into regressions on any chain that has a live referendum at the upgrade, and one **unmetered unbounded storage scan** added on the member-removal path.

Findings are ordered most-severe first. `file:line` references are against this PR's head.

---

### Does #654 subsume #649?

**Yes, substantively — #649 can be closed once two small pieces are carried over.** Every item in #649's description is present in #654, either verbatim or correctly adapted to flat pricing: the `settle` extraction, the full vesting-mock thread-local reset (here a strict superset — it also resets `DROP_CREDITS`), lazy pot derivation, both runtime `const _: () = assert!` guards, the `MAX_SUPPLY`/`MILLIS_PER_DAY` dedup, the `per_transfer_weight` PoV-omission fix and its assertion, and the docs/`TestCommons` changes. #649's depth-aware shared helper (`insert_leaf_weight_at_depth`) is legitimately superseded by the flat `INSERT_LEAF_*` constants, so the depth-monotonicity tests it added are correctly dropped.

Two things from #649 were dropped rather than adapted and are worth porting into #654 (or a tiny follow-up) before closing #649:
- the `payout_weight_never_undercharges_the_benchmarked_base` regression test — `payout_weight` (`pallets/vesting/src/weights.rs:28`) still subtracts hand-maintained `BENCHMARK_TREE_*` counts, and nothing now pins that the flat insert added back exceeds them;
- the derivation doc-comments on `BENCHMARK_TREE_READS/WRITES` / `CLAIM_BENCHMARK_TREE_WRITES`, which are the only record of how those magic numbers were obtained.

---

### Blocking

**1. No upgrade migration for the new referenda state — breaks the PR's own `try_state` invariant and can silently lose an approved governance proposal.** `pallet_referenda` gains `ActiveReferendaCount` (`pallets/referenda/src/lib.rs:282`) and starts pinning proposal preimages, but `runtime/src/lib.rs:176` adds no referenda migration and `pallets/referenda/src/migration.rs` only got test edits. Two consequences on a chain that has any `Ongoing` referendum at the 142→143 upgrade:

- *(a)* `ActiveReferendaCount` is `ValueQuery`/defaults to 0 while `K` referenda are already ongoing. The `do_try_state` check the PR itself adds (`lib.rs:1701`, "`ActiveReferendaCount` must match the number of `Ongoing` referenda") fails immediately under try-runtime; and because `note_one_fewer_active` uses `saturating_sub` (`lib.rs:1622`), the counter floors at 0 as those `K` conclude and permanently under-reports — silently loosening the new `MaxActive` DoS bound by up to `K`.
- *(b)* Every terminal transition now calls `T::Preimages::drop(&status.proposal)` (`lib.rs:684/723/1461/1505/1542`) assuming `submit` took a matching `request`. A referendum submitted at 142 never took one. On the **approve** path the drop at `lib.rs:1505` runs *after* `schedule_named` has taken the scheduler's own enactment request, so it consumes *that* pin (count 1→0, reverting to `Unrequested`). The noter can then `unnote_preimage`, the bytes are deleted, enactment fails with `CallUnavailable`, and the approved proposal is dropped with no retry path — exactly the failure the preimage-pinning fix was written to prevent. A shared-hash variant lets a straddling referendum's terminal drop delete bytes out from under a live post-upgrade referendum.

One `OnRuntimeUpgrade` iterating `ReferendumInfoFor` fixes both: seed `ActiveReferendaCount` from the `Ongoing` count and `T::Preimages::request(&hash)` each `Ongoing` lookup proposal.

**2. `remove_votes_from_ongoing_polls` does an unbounded, unmetered `Voting::iter()` in a mandatory scheduler hook.** `pallets/ranked-collective/src/lib.rs:801` scans the entire poll-major `Voting` double map on every removal (an account-scoped `iter_prefix` isn't possible given the key order), filters for the one member, and calls `T::Polls::access_poll` per hit — which for the referenda impl rewrites `ReferendumInfoFor` and reschedules the poll's alarm. Declared weight is unchanged (`remove_member(rank)`, and the post-dispatch weight is pinned back to that benchmark figure at `lib.rs:618`), and the benchmark runs against an empty `Voting` map, so the loop contributes nothing to the measured cost even if `weights.rs` were regenerated. `Voting` records for completed polls are only pruned by the permissionless, optional `cleanup_poll`, so the map grows monotonically. On this runtime `remove_member` is enacted by the Scheduler in `on_initialize` (Root via a passed referendum), so the overweight block cannot be avoided by dropping a transaction — and the member being removed is exactly the party who can inflate `Voting` cheaply (submit + vote on referenda) to make their own removal expensive. Needs a `max`-witness bound (as `cleanup_poll` has) and/or an account→polls reverse index, plus a weight component that reflects the real work. (The alarm-reschedule fan-out is bounded at `MaxActive` = 128 per removal, ~70× the declared weight; the map scan is the unbounded part.)

---

### High (strongly recommend before ship)

**3. `MaxActive` can freeze the only governance lane for 45+ days.** `submit` gates on the global bound (`lib.rs:553`) with no exemption and no per-account cap. `SubmitOrigin` is members-only, `CancelOrigin`/`KillOrigin` are `EnsureRoot`, and Root is reachable *only* through a passed referendum — so once one member fills all 128 slots with cheap depositless `Bounded::Inline` referenda, every further `submit` (from anyone, including the Root-only recovery path's prerequisite) fails with `TooManyActive`. It self-heals at the 45-day `UndecidingTimeout` (so not a permanent brick), but the freeze is renewable, and it can be *extended* toward ~100 days: placing a refundable decision deposit (`place_decision_deposit`, plain `ensure_signed`) parks a referendum in the queue, which is exempt from the undeciding timeout. Cost is only refundable submission/decision deposits. Consider reserving headroom in `MaxActive` (bound only depositless referenda) or a per-account submission cap.

**4. Terminal-transition weights not updated for the new work.** `cancel` (`lib.rs:662`), `kill` (`lib.rs:705`) and the `nudge_referendum` service branches (timeout/approve/reject) each gained a `Preimages::drop` — which can delete a `PreimageFor` blob up to `MAX_SIZE` (4 MiB) — plus an `ActiveReferendaCount` read/write, but only `submit`'s weight annotation was bumped; `nudge_referendum` still uses the stale benchmarked `max_weight_of_nudge`. This is under-accounting inside the mandatory scheduler hook.

**5. `submit` pins preimages deposit-free, and the new `request` write is unweighted.** A member notes a ≤4 MiB preimage, submits (pinning it via `lib.rs:544`), then `unnote_preimage`s to reclaim the storage deposit while the bytes are retained until the referendum ends (up to 45 days); ×128 that is ~512 MiB held against only refundable deposits. Separately, the new `Preimages::request` read+write is not reflected in `submit`'s declared weight (only the `ActiveReferendaCount` `reads_writes(1,1)` was added).

---

### Medium / low (non-blocking)

**6. `Utility::dispatch_as` / `dispatch_as_fallible` do the high-security policy read but don't charge it.** These two Root-origin wrappers (`pallets/utility/src/lib.rs:405,644`) perform `is_call_allowed` (one `HighSecurityAccounts` read) but their `#[pallet::weight]` wasn't updated, unlike every other call-carrying dispatchable in this PR; `qp_high_security`'s contract says callers must charge that read. Minor undercharge on a Root/scheduler path.

**7. `CIRCUIT_MAX_TREE_DEPTH = 16` flat pricing has no enforced ceiling and no compile-time link to the circuits' `MAX_DEPTH`.** `pallets/zk-tree/src/lib.rs:68` hand-copies `qp_zk_circuits_common::zk_merkle::MAX_DEPTH` (a crate zk-tree doesn't depend on), and `grow_tree` still permits depth up to `MAX_TREE_DEPTH = 32` (only a `debug_assert`, compiled out in release). Past 4^16 leaves every leaf-insert weight across five pallets under-charges and wormhole proofs become unverifiable. Reaching that is impractical (~4.3e9 leaves), but the constant is unpinned — add `const _: () = assert!(CIRCUIT_MAX_TREE_DEPTH as usize == qp_zk_circuits_common::zk_merkle::MAX_DEPTH)` in a crate that depends on both (wormhole or the runtime, which already carries the sibling `SCALE_DOWN_FACTOR` guard), and cap `grow_tree` at the circuit depth.

**8. No `post_dispatch` refund for statically over-charged transfers.** The extension only ever registers *extra* weight (`register_extra_weight_unchecked`); it never refunds. `recover_funds` is charged the static worst case (`MaxPendingPerAccount + 1` = 17 inserts) and failed/short `batch`/`if_else` reserve per-transfer weight that is never returned, even though `Pre` already carries `charged_transfers`. Implementing `post_dispatch_details` to refund `per_transfer_weight() * (charged_transfers - recorded)` would recover it.

**9. (Pre-existing, not introduced here) the "eventless `increase_balance`" comment for mining rewards is wrong.** `runtime/src/transaction_extensions.rs:109` claims mining rewards/treasury are credited event-free; `pallet_mining_rewards::mint_reward` actually uses `mint_into`, which emits `Balances::Minted` (the event the scanner records at `:321`, with no `MintingAccount` guard). They avoid double-recording only because distribution runs in `on_finalize`, outside every scan window — the load-bearing invariant, stated wrongly. The comment reads like the wormhole pallet's rationale (which genuinely uses event-free `increase_balance` and pins it with a test) copied onto the wrong pallet. Doc-only today, but a future move of distribution to `on_initialize`/an extrinsic would double-record and inflate wormhole exit capacity. Worth correcting while this file is open.

**10. (Latent, unreachable in this runtime) `do_demote_member`'s `Some(rank)` arm doesn't reconcile votes.** A demotion that lowers rank but keeps membership (`lib.rs:893`) leaves the member's votes in ongoing tallies while `remove_from_rank` has already shrunk the `support` denominator — the same inflation the PR fixes for full removal. `DemoteOrigin`/`PromoteOrigin` are `NeverEnsureOrigin` and the collective is flat rank-0, so it can't fire here, but this is a vendored pallet fork; hoist the reconcile call above the match (and add a rank-1→0 test) so a future multi-rank config doesn't reopen it.

**11. Test-infra: referenda mock `MaxActive` static leaks between tests.** `submission_bounded_by_max_active`/`timed_out_referendum_frees_active_slot` call `MaxActive::set(..)` (`pallets/referenda/src/mock.rs:110`, `tests.rs:80/115`) and never restore it, and `ExtBuilder::build()` resets nothing — so under a serialized harness the leaked value fails later submit-heavy tests. This is the exact leak class this PR just fixed on the vesting side with `reset_thread_local_state`; apply the same pattern here.

**12. The preset-cliff regression test is weaker than its message.** `preset_vesting_schedules_enforce_the_published_cliff` (`runtime/src/genesis_config_presets.rs:710`) ends with `assert!(checked >= 4)`, but `dev` alone contributes 4 schedules, so a `heisenberg` regression wouldn't be caught — assert the exact total. Also, `assert_eq!(*cliff, start + days_ms(90))` is applied to *every* preset, so a future allocation table that legitimately ships a non-90-day cliff will fail this test.

**13. Stale docs contradict the PR's own change.** `docs/TECH_COLLECTIVE_GOVERNANCE_TUNING.md:122/128` and `docs/RUNTIME_SURFACE.md:150` still describe `RemoveOrigin` as `EnsureRootWithSuccess<AccountId, ConstU16<0>>` / "any-member" — but this PR changed it to `EnsureRootRemoveKeepsMemberFloor`, and that floor is the load-bearing security property here. `RUNTIME_SURFACE.md` (which this PR does edit) also omits the new `MaxActive`/`ActiveReferendaCount`/`TooManyActive` surface and `Vesting::PayoutProofNotRecorded`.

**14. DRY / cleanup** (the repo's DRY rule): the high-security gate now exists in three copies — the new `ensure_nested_call_allowed` plus the pre-existing inline blocks in `dispatch_as`/`dispatch_as_fallible` (`utility/src/lib.rs:424/662`), which weren't refactored onto it; the re-check block is pasted at five call sites. The tally-withdrawal arithmetic is duplicated between `vote` and `remove_votes_from_ongoing_polls` (`ranked-collective/src/lib.rs`). The `Preimages::drop` + `note_one_fewer_active` pair is copy-pasted at five terminal sites — and the benchmark-only `end_ongoing` path (`referenda/src/lib.rs:942`) already diverged (it got the counter decrement but not the drop), which is the exact drift a `conclude_ongoing(status)` helper would prevent. `#[frame_support::transactional]` on `vesting::pay_out` is redundant (FRAME already wraps call bodies in a storage layer). The four `weights.rs` still carry now-dead `pallet_zk_tree::Config` bounds and stale "reads live depth" doc comments after the switch to flat pricing.

---

### Candidates investigated and cleared (for the record)

- **Deleting the multisig v0→v1 migration does *not* strand proposals.** `call_weight` was added to and removed from the stored `ProposalData` entirely within an unreleased window (between spec-129 `v0.6.5` and spec-130 `v0.7.0`); no released runtime ever persisted the v0 layout, every release since carries the current layout, and the deleted migration was never in a release. Keeping it would have been the hazard (it would run against already-current records and bump the on-chain version). One residual nit: `STORAGE_VERSION` stays `1` while a live pallet's on-chain version is likely `0` — harmless (same encoding), but if you want the try-runtime version check clean, set it back to `0`.
- **`force_batch` billing a rejected child its full `call_weight`** (via `Err(e.into())` → `actual_weight: None`) is real but matches FRAME's own `CallFiltered` semantics exactly; it's a fee-only missing refund, self-inflicted and bounded by the caller's own batch, and over-accounts block weight in the safe direction. Not a defect.

---

**Verdict: Request changes** — finding 1 (missing referenda upgrade migration) is a correctness regression that breaks the PR's own `try_state` invariant and can lose an approved governance proposal, and finding 2 (unbounded unmetered scan in a mandatory hook) needs a bound before ship. The rest of the PR is solid and, once these land, it fully replaces #649.

n13 (changes requested):
## Re-review — V12 vesting / security fixes at `9336142a`

The 12-commit follow-up fixes most of the previous review well: the unbounded vote scan is gone; per-submitter admission, proposal-size collateralization, terminal/preimage weights, utility policy-read weights, mock reset, vesting weight pin, docs, and DRY cleanup are all materially improved. Two correctness blockers remain, and the new wormhole refund only restores block capacity rather than the payer's fee.

### Blocking

**1. There is still no runtime migration for pre-143 ongoing referenda.** `runtime/src/lib.rs:181-187` wires only the wormhole and treasury migrations. The new `ActiveReferendaCount` / `ActiveSubmissionCount` storage therefore starts at zero for every referendum already `Ongoing` when spec 142 upgrades, even though `pallets/referenda/src/lib.rs:1785-1803` now requires both values to match live state. Conclusions then `saturating_sub` those zero/misaligned counters (`:1691-1698`), potentially consuming counts belonging to post-upgrade submissions and permanently weakening both admission limits.

The preimage side is more serious: pre-143 lookup proposals never took the submit-time request that `conclude_ongoing` now unconditionally drops (`:1701-1707`). On approval, `schedule_named` takes the scheduler's enactment request and `conclude_ongoing` immediately consumes it. The noter can then unnote/delete the bytes before enactment, so an approved proposal can still terminate as `CallUnavailable`. The edits in `pallets/referenda/src/migration.rs` only pre-seed the new counters inside old migration tests; they do not migrate live state. Add and wire a state-aware migration (bounded/stepped if the live cardinality requires it) that seeds both counters and requests every pre-existing ongoing lookup proposal.

**2. Membership changes still leave an invalid vote counted until an optional later extrinsic.** `pallets/ranked-collective/src/lib.rs:768-781` explicitly makes reconciliation a separate permissionless `remove_ineligible_vote` call, and `do_remove_member_from_rank` at `:966-979` deletes membership without touching the vote. The new regression test demonstrates the bad intermediate state directly: `pallets/ranked-collective/src/tests.rs:670-674` removes the sole aye voter and support jumps from `1/3` to `1/2` until cleanup is submitted.

“Removal proposals should batch cleanup” is not a protocol invariant: a Root proposal can call `remove_member` directly, cleanup may never be included, and a scheduled referendum service can observe the inflated tally first. This leaves the original governance-tally vulnerability open indefinitely. Keep the bounded permissionless call for backlog cleanup if useful, but membership loss/demotion needs bounded **atomic** reconciliation (for example, a reverse account→active-polls index maintained by `vote`, or equivalent state that the removal path can consume without a global scan).

### High

**3. The wormhole overcharge is not refunded to the sender.** `runtime/src/lib.rs:154-171` orders `ChargeTransactionPayment` before `WormholeProofRecorderExtension`. FRAME runs transaction-extension post-dispatch hooks left-to-right, so transaction payment computes and finalizes the actual fee before the wormhole extension returns its unspent weight at `runtime/src/transaction_extensions.rs:426-477`. The trailing `WeightReclaim` correctly returns that weight to block capacity, but cannot undo the already deposited transaction fee. Consequently failed transfers, short batches/`if_else`, and `recover_funds` with fewer than the static maximum still pay the full proof-insert reservation despite this commit's refund claim. Put the weight adjustment before fee correction (or otherwise correct the fee explicitly) and add a full-pipeline assertion on the payer balance / `TransactionFeePaid`, not only `PostDispatchInfo` and block weight.

### Residual follow-up

`pallets/zk-tree/src/lib.rs:63-68,266-277` still allows release builds to grow beyond depth 16 while both circuit validity and flat insert pricing stop at 16, and the hand-copied circuit constant has no compile-time link to `qp_zk_circuits_common::zk_merkle::MAX_DEPTH`. The current dependency is 16, so this is not today's blocker, but enforce the ceiling and pin the cross-crate equality before dependency drift or tree exhaustion turns into unverifiable, under-metered leaves.

### Validation

- `git diff --check 0eb8052a...9336142a` — passed
- `cargo test --locked -p pallet-referenda` — 60 passed
- `cargo test --locked -p pallet-ranked-collective` — 24 passed, 2 ignored
- `cargo test --locked -p pallet-utility` — 44 passed
- `cargo test --locked -p pallet-vesting` — 58 passed
- `cargo test --locked -p quantus-runtime --lib` — 46 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime --test mod --features fast-governance` — 40 passed, 1 ignored (two unused-import warnings)
- Current GitHub format, Linux/macOS test matrix, Clippy/doc, and compose-config checks — all green

**Verdict: Request changes.** The bounded redesign resolves the mandatory-hook DoS, but findings 1 and 2 still permit lost approved proposals and stale-member votes to affect governance. Finding 3 means the advertised fee refund is incomplete.

n13 (changes requested):
## Re-review — V12 vesting / security fixes at `a6bfaf35`

The transaction-extension reorder fixes the fee-refund finding, and the merge of #656 preserves the single-copy event reader. Two governance correctness blockers remain.

### Blocking

**1. Spec 142 → 143 still has no migration for existing ongoing referenda.** `ActiveReferendaCount` and `ActiveSubmissionCount` are new storage (`pallets/referenda/src/lib.rs:301-317`), but `runtime/src/lib.rs:185-191` still wires only the wormhole and treasury migrations. A valid spec-142 `Ongoing` entry therefore starts spec 143 absent from both counters, despite `do_try_state` requiring those counters to equal live state (`pallets/referenda/src/lib.rs:1785-1802`).

I reproduced the resulting state corruption against this head: after simulating one pre-143 ongoing referendum, a post-upgrade submission set the global count to 1; concluding the old referendum then saturated that count back to 0 while the new referendum remained ongoing, and `do_try_state` rejected the state. Per-submitter counts have the same initialization problem.

The missing preimage migration can also lose an approved proposal. A pre-143 lookup proposal did not take the new submit-time request. On approval, the scheduler takes its enactment request, then `conclude_ongoing` unconditionally drops one (`pallets/referenda/src/lib.rs:1701-1707`). My diagnostic reproduced the request count returning to zero and the noter successfully unnoting/deleting the proposal bytes before enactment. Add and wire a state-aware migration that seeds both counters and requests every pre-existing ongoing lookup proposal. Because the old runtime had no active-count bound, this needs an explicitly verified cardinality bound or a stepped migration rather than assuming the new limit applies retroactively.

**2. The vote-retention revert preserves the governance threshold flip from the previous review.** `do_remove_member_from_rank` now intentionally leaves the removed account's vote in every ongoing tally while shrinking the current-member denominator (`pallets/ranked-collective/src/lib.rs:894-910`); the regression test pins the resulting support increase from `1/3` to `1/2` (`pallets/ranked-collective/src/tests.rs:626-653`). The 100% clamp only limits the extreme case; it does not prevent a decision threshold from being crossed.

For the runtime's 60% support threshold, six members and three ayes initially have 50% support and fail. If one of those aye voters is removed while the poll remains ongoing, the retained three ayes become `3/5 = 60%`, so the referendum can pass even though only two current members support it. A queued/preparing ongoing referendum can receive votes while a separate removal referendum decides and enacts, so this is reachable despite `MaxDeciding = 1`. Matching the upstream pallet's behavior does not make that transition safe for this runtime. Membership loss/demotion needs bounded atomic reconciliation, such as an account-to-active-polls reverse index; optional cleanup or a global unbounded scan is insufficient.

The PR description also still says ongoing tallies are reconciled when a member leaves, which no longer matches the implementation.

### Cleared and validation

- `ab1ad356` moves wormhole proof recording before transaction payment. The full signed-extrinsic test confirms the payer's balance and `TransactionFeePaid`, and all 14 proof-recorder tests pass.
- I inspected the #656 merge conflict resolution; it correctly combines flat proof-insert accounting with the single-copy, per-byte-metered event scan.
- `git diff --check 083116cf...a6bfaf35` — passed
- `cargo test --locked -p pallet-referenda` — 62 passed, including two temporary upgrade-state diagnostics described above
- `cargo test --locked -p pallet-ranked-collective` — 19 passed, 2 ignored
- Focused full-pipeline runtime fee-refund test — passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime --lib wormhole_proof_recorder_` — 14 passed
- Current GitHub format, compose-config, and Clippy/doc checks are green; Linux and macOS test matrices are still running on this exact head.

**Verdict: Request changes.** The fee-refund issue is resolved, but shipping without the referenda migration can corrupt admission accounting and unpin an approved proposal, while stale votes can still change governance outcomes after membership removal.

n13 (changes requested):
## Updated review — V12 vesting / security fixes at `a6bfaf35`

Head is unchanged since my last review; this revises the verdict after re-checking the two blocking findings against the runtime's actual governance parameters. **One of the two blockers is withdrawn.** All CI is now green on this head (format, config, Clippy/doc, Linux and macOS matrices).

### Withdrawn — vote retention on member removal (previously blocking, twice)

I accept the author's position and am dropping this to a documented trade-off.

The decisive argument is the remedy: `min_approval` is a constant 61% (`runtime/src/governance/definitions.rs:119-123`) and `Tally::approval` is `ayes / (ayes + nays)` (`pallets/ranked-collective/src/lib.rs:141-143`). In the 6-member / 3-aye scenario I raised, removing one aye voter gives `3/5 = 60%` support — but two nays from the remaining members drop approval to `3/5 = 60% < 61%` and the referendum cannot pass. With `confirm_period = DAYS` plus the decision period, there is at least a full day for those nays to arrive. That is exactly the design the track comment already states ("2 nays always block", "the 24h confirm period guarantees nays can arrive for a full day before any approval"), and the removal itself requires a passed Root referendum, so the precondition is already a governing majority. An in-protocol remedy available to any single pair of remaining members is not a blocker.

One correction on the other stated reason, for the record rather than to reopen anything: `max_deciding = 1` does not by itself prevent removal during voting. Votes are accepted on **any** `Ongoing` referendum, including one still `Preparing` or `Queued` (`pallets/referenda/src/lib.rs:946-964` matches `ReferendumInfo::Ongoing` with no deciding check), so a referendum can bank votes, sit undecided while a separate removal referendum takes the single deciding slot and enacts, then start deciding against the shrunken electorate. `max_deciding = 1` only protects the deciding window. This doesn't change the conclusion — the downvote remedy covers the reachable case — so the finding is closed as accepted and documented (`pallets/ranked-collective/src/lib.rs:896-904`).

**Housekeeping:** the PR description still claims "Reconcile ongoing referendum tallies when a member leaves" and describes commit `3c5de631` as a fix; `a6bfaf35` reverted that. Please update the description so the merge commit doesn't record the opposite of what shipped.

### Blocking (1 remaining)

**No 142 → 143 migration for the new referenda state.** `ActiveReferendaCount` and `ActiveSubmissionCount` are new storage (`pallets/referenda/src/lib.rs:307,316`), and `submit` now takes a preimage request for `Lookup` proposals — but `runtime/src/lib.rs:186-191` still wires only the wormhole and treasury migrations, and the pallet has no `on_runtime_upgrade` of its own (`pallets/referenda/src/lib.rs:514-522`).

Any referendum that is `Ongoing` across the upgrade block is then inconsistent with code that assumes it isn't:

- **Preimage unpinning can lose an approved proposal.** A pre-143 `Lookup` proposal never took the submit-time request, but `conclude_ongoing` unconditionally calls `T::Preimages::drop(proposal)` (`pallets/referenda/src/lib.rs:1705-1708`). On the approve path `schedule_named` takes the scheduler's own enactment request first, so the drop consumes *that* pin, returning the hash to `Unrequested`. The noter can then `unnote_preimage`, the bytes are deleted, and enactment fails with `CallUnavailable` — the exact failure `1f678a3e` was written to prevent. This is the serious half, and it lands on `set_code`-sized proposals specifically, since those are the ones that must be `Lookup`.
- **Counter drift weakens both admission bounds.** Both counters default to 0 while `K` referenda are already ongoing; `note_one_fewer_active` then `saturating_sub`s slots belonging to post-upgrade submissions (`:1691-1698`), permanently under-reporting. `do_try_state` (`:1786-1802`) rejects the state immediately under try-runtime. Lower severity than the preimage case, but it silently loosens the DoS bound this PR just added.

That this release already ships `pallet_wormhole::MigrateV1ToV2` and `pallet_treasury::MigrateV0ToV1` is the point: live state is being migrated, so live referenda state needs the same treatment. One `OnRuntimeUpgrade` iterating `ReferendumInfoFor` closes both halves — seed the two counters from the `Ongoing` set and `Preimages::request` each ongoing `Lookup` proposal. The old runtime had no active-count bound, so size it with an explicitly verified cardinality bound or make it stepped rather than assuming `MaxActive` applies retroactively.

If you would rather not carry a migration, the alternative is an explicit operational gate: verify at the upgrade block that no referendum is `Ongoing` (cheap to check, and likely true) and record that as a release precondition. Shipping with neither is what I'm blocking on.

### Non-blocking follow-ups

- `pallets/zk-tree/src/lib.rs:63-68,266-277` — release builds can still grow past depth 16 while circuit validity and flat insert pricing both stop there, and `CIRCUIT_MAX_TREE_DEPTH` has no compile-time link to `qp_zk_circuits_common::zk_merkle::MAX_DEPTH`. Unreachable today (~4.3e9 leaves); worth pinning before dependency drift.

### Previously blocking, confirmed resolved

- `ab1ad356` reorders the proof-recorder extension ahead of `ChargeTransactionPayment`, so the wormhole overcharge now refunds to the payer's fee, pinned by a full-pipeline `TransactionFeePaid` / balance assertion.
- `03b466b0` → `a6bfaf35` removed the unbounded `Voting::iter()` scan from the scheduler-enacted removal path.
- The #656 merge resolution correctly combines flat proof-insert accounting with the single-copy, per-byte-metered event scan.

**Verdict: Request changes** — one blocker left. Everything else in this PR is ready; ship the referenda migration (or gate the release on there being no ongoing referendum at the upgrade) and this is good to merge.

n13 (approved):

PR #661, opened 2026-08-11
cache Dilithium pubkeys on chain to enable signature-only transactions

illuzen:
# Cache Dilithium pubkeys on chain, enable signature-only transactions (`pallet-pubkey`)

## Problem

`AccountId` is the Poseidon hash of the account's ML-DSA public key, and lattice signatures have no pubkey recovery, so every signed extrinsic must carry the full public key next to the signature (`SignatureWithPublic = [sig ‖ pubkey]`):

| Scheme | Pubkey | Signature | Total auth blob |
| --- | --- | --- | --- |
| ML-DSA-65 | 1952 B | 3309 B | ~5.3 KB |
| ML-DSA-87 | 2592 B | 4627 B | ~7.2 KB |

That pubkey is identical in every transaction an account ever sends.

## Solution

Remember the pubkey on chain after the first transaction and let subsequent transactions omit it:

- **`pallet-pubkey` (new, index 23):** a single storage map `Pubkeys: AccountId32 → DilithiumSigner`. No extrinsics, events, or genesis config.
- **Automatic registration:** the first successful full-signature verification writes the carried pubkey to the map. Verification already proves the signature *and* `Poseidon(pubkey) == account`, so the binding is trusted and permanent (an account cannot rotate to a key with a different hash). Existing entries are never rewritten.
- **New wire variants:** `Dilithium87SigOnly` / `Dilithium65SigOnly` appended to `DilithiumSignatureScheme` (existing variant indices untouched). They carry only the signature; verification resolves the pubkey from the map and fails with `BadProof` if no key of the matching parameter set is cached.
- **`CachedSignature<Runtime>`:** the runtime `Signature` type is now a SCALE-transparent wrapper around the enum whose `Verify` impl can read and fill pallet storage (the phantom runtime parameter encodes to nothing but brings `pallet_pubkey::Config` into scope).

**Savings:** 1952 B (ML-DSA-65) or 2592 B (ML-DSA-87) per transaction after the first — ~36% of the auth blob. Asserted byte-for-byte in the e2e tests.

## Compatibility

- **Existing clients need no changes.** Full `SignatureWithPublic` transactions encode and verify exactly as before, forever. Sig-only is an opt-in size optimization.
- Wallets should switch to sig-only only after the registering transaction is **in a block** (pool-validation writes land in a discarded overlay), e.g. after first confirmation or a state query of `Pubkeys`.
- `transaction_version` 3 → 4 (new signature encodings), `spec_version` 144 → 145.

## Changes

- `primitives/dilithium-crypto`: appended `SigOnly` variants, `carried_signer()` helper, `MaxEncodedLen` on `DilithiumSigner`; the pure `Verify` impl rejects `SigOnly` (no storage context).
- `pallets/pubkey`: storage map, `CachedSignature` wrapper + `Verify` impl, `pubkey_of` getter.
- `runtime`: `Signature = pallet_pubkey::CachedSignature<Runtime>`, `AccountId = AccountId32` directly, pallet wired at index 23, version bumps.
- `node/src/benchmarking.rs`, `runtime/tests`: signature construction adapted to the wrapper.
- `docs/RUNTIME_SURFACE.md`: type table, pallet index table, new index-23 section.

## Notes / trade-offs

- Sig-only verification adds one storage read; full-signature verification adds one `contains_key` read (plus one insert the first time).
- Registration stores ~2–2.6 KB per account with no deposit. Only a key that hashes to the signing account can ever be registered, so there is no griefing vector for *other* accounts; if state growth becomes a concern, a deposit can be added later.
- Pubkey recovery from the signature (ECDSA-style) is not possible with ML-DSA; an on-chain cache is the only way to drop the pubkey from the wire.

## Testing

- `pallet-pubkey` unit tests (10): register on full verify, no cache on failed verify, sig-only success/failure, parameter-set mismatch, wrong account, no overwrite, wire-format equivalence with the bare enum.
- Runtime e2e (`runtime/tests/transactions/sig_only.rs`, 4): first full tx caches the key; sig-only tx applies through the full `Executive` pipeline and saves exactly `Dilithium65Public::LEN` bytes; sig-only rejected with `BadProof` when unregistered; sig-only from an impostor key rejected.
- All pre-existing transaction/crypto tests pass (18 runtime transaction tests, 18 `qp-dilithium-crypto` tests); `quantus-node` compiles.

## Follow-ups

- Teach wallet/SDK signing to emit `SigOnly` automatically once the account's key is on chain.
- Optionally price the sig-only storage read into weights and/or add a registration deposit.


n13 (changes requested):
**Reviewer model:** GPT Sol

Verdict: REQUEST_CHANGES — the cache lifecycle permits permanent orphaned state, its database work is outside transaction weight accounting, and the required format gate is failing.

Blocking findings:

1. `pallets/pubkey/src/lib.rs:62-80`: `Pubkeys` entries are permanent even after the corresponding system account is reaped. The runtime inherits `OnKilledAccount = ()`, and this pallet neither installs a cleanup hook nor holds a deposit/provider. A newly funded account can therefore submit one full-signed `transfer_all(..., keep_alive = false)`: verification inserts its 1.9–2.6 KB key, dispatch reaps the account and returns the remaining balance, but the key stays forever. Repeating fund/register/reap creates unbounded orphan state. Remove the cache entry from the account-death lifecycle (and test that full-signed reap leaves no entry), or otherwise collateralize the permanent storage.

2. `pallets/pubkey/src/lib.rs:136-158`: every signed extrinsic now performs a cache read, and the first full signature also performs a multi-kilobyte write, from `Verify`. `UncheckedExtrinsic::check` runs this before `TxExtension` weight/payment handling, so neither the block weight limit nor the fee includes these database operations. Account for the worst-case read/write in the signed-extrinsic base weight, or move the cache work into a weighted path; add coverage showing the charged weight includes cached and first-use cases.

3. The required `cargo +nightly fmt --all -- --check` fails on `pallets/pubkey/src/lib.rs`, `pallets/pubkey/src/tests.rs`, and `runtime/src/lib.rs`; GitHub's Fast Checks job is red and consequently skipped the matrix and analysis jobs.

Also fix the changed runtime-surface version line: `docs/RUNTIME_SURFACE.md:10` still says `spec_version = 143`, while `runtime/src/lib.rs:79` sets 145.

Validation at head `82241427cdb81c83cd163fa2245832ff4c27bd4c`:

- `cargo test --locked -p pallet-pubkey` — 10 passed
- `cargo test --locked -p quantus-runtime --test mod sig_only::` — 4 passed (focused filter)
- `cargo test --locked -p qp-dilithium-crypto` — 18 passed
- strict targeted Clippy for `pallet-pubkey` and `qp-dilithium-crypto` — passed
- `taplo format --check --config taplo.toml` — passed
- `cargo +nightly fmt --all -- --check` — failed as described above

n13 (changes requested):
**Reviewer model:** GPT Sol

Verdict: REQUEST_CHANGES — the earlier lifecycle, signature-verification metering, and formatting blockers are addressed, but the new account-death cleanup introduces an additional unmetered database write.

Blocking finding:

1. `runtime/src/configs/mod.rs:86-123,166-168` and `pallets/pubkey/src/lib.rs:90-93`: `PubkeyCacheVerifyWeight` budgets one read plus one write for signature verification, but installing `Pubkey` as `OnKilledAccount` makes every account reap also execute `Pubkeys::remove`. The new `full_signed_reap_clears_cached_pubkey` path demonstrates the worst case: a first full signature does `contains_key` + `insert`, then `transfer_all(..., keep_alive = false)` reaps the same account and performs a second write to remove that entry. The base surcharge covers only the read and first write. The existing generated `pallets/balances/src/weights.rs:124-133` predates this hook and accounts only for `System::Account`, so it does not absorb the `Pubkeys` deletion; reaps dispatched internally (for example by a scheduled/root call) do not gain signature-verification coverage either. This leaves block weight and fees below the work performed. Account for the cleanup on every reachable reap path—typically by regenerating/adjusting the affected runtime weights or moving cleanup into an explicitly weighted lifecycle—and add coverage for the combined first-registration-plus-reap cost rather than testing registration and cleanup independently.

Non-blocking follow-up: `docs/RUNTIME_SURFACE.md:10` still reports `spec_version = 143`, while `runtime/src/lib.rs:75` sets 145. This line is modified by the PR for `transaction_version`, so please make both values match the runtime.

Validation at head `835b2f54b2c7fc05fa732c2e61c65bf3fc582613`:

- `cargo test --locked -p pallet-pubkey` — 11 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime --test mod sig_only::` — 6 passed
- `cargo test --locked -p qp-dilithium-crypto` — 18 passed
- `cargo clippy --locked -p pallet-pubkey --no-deps -- -D warnings` — passed
- `taplo format --check --config taplo.toml` — passed
- Current GitHub CI — all five reported checks passed

n13 (changes requested):
**Reviewer model:** Opus 5

Reviewed at `769f4ed4` against `origin/main`. I built the branch (`cargo check -p pallet-pubkey -p quantus-runtime --tests`, clean) and ran `cargo test -p pallet-pubkey` (11/11) and `cargo test -p quantus-runtime --test mod transactions::` (20/20) — all green. The core crypto reasoning is sound: `DilithiumSignatureScheme::verify` already enforces `Poseidon(pubkey) == signer` (`primitives/dilithium-crypto/src/traits.rs:156`), so writing the carried key on a successful full verify is a safe, unforgeable binding, and the `SigOnly` arms resolve the key strictly by the claimed signer with a parameter-set match. I found no way to impersonate an account. The `RuntimeBlockWeights` hand-rolled builder is a faithful reproduction of `with_sensible_defaults` (same 10% `avg_block_initialization`, same class limits) plus the surcharge.

One blocking issue.

---

### 1. Blocking — the signed payload does not commit to the signature form, so anyone can re-encode `SigOnly` ⇄ full

`pallets/pubkey/src/lib.rs:167`, `primitives/dilithium-crypto/src/types.rs:77`

The ML-DSA signature covers `(call, tx_ext, implicit)` only. The *variant* of `DilithiumSignatureScheme` is not part of the signed message, and once a key is in `Pubkeys` it is world-readable. So the same signature is valid in both encodings, and a third party can rewrite either into the other.

I confirmed this end-to-end with a scratch test on this branch (reverted afterwards, nothing committed): register the key with a full tx, take the victim's sig-only tx for nonce 1, pull the raw signature out of its preamble, re-wrap it as `Dilithium65SignatureWithPublic` with the pubkey read from `Pubkeys`, and re-submit:

```
REENCODE OK: sig_only_len=3392 inflated_len=5344 extra_bytes=1952 sender_paid=11120787000
```

`Executive::apply_extrinsic` accepted the rewritten extrinsic and the **sender** paid the inflated length fee. Three consequences:

* **The optimization is not incentive-compatible.** `LengthToFeeMultiplier` charges `1e6` per byte, so the 1952 extra bytes cost the sender ~1.95 mUNIT — ~21% of this transaction's total fee (2592 B / ML-DSA-87). Transaction fees are not burned: `FungibleAdapter<Balances, TransactionFeesCollector<Runtime>>` (`runtime/src/configs/mod.rs:479`) routes them to `CollectedFees`, which `pallet_mining_rewards::on_finalize` re-mints to the block author (`pallets/mining-rewards/src/lib.rs:132`). A rational miner therefore inflates every sig-only extrinsic it includes: it collects a strictly larger fee, the signature still verifies, and nothing on chain reveals the substitution. The 36% wire saving is exactly the miner's incremental revenue for undoing it.
* **Extrinsic-hash malleability.** The included extrinsic hashes differently from the one the wallet submitted. `author_submitAndWatchExtrinsic` and any hash-based tracking will report the user's transaction as dropped/invalid while its effects actually landed.
* **Third-party griefing** of the fee, independent of miner incentive.

Cheapest fix that closes it without touching the legacy wire format: domain-separate the two forms at signing time — have the `SigOnly` arms verify over a prefixed message (e.g. `b"qsigonly" ‖ msg`) rather than `msg`. The signer then picks the form before signing, the two signatures are no longer interchangeable, and full `SignatureWithPublic` transactions keep verifying byte-for-byte as today. Whatever the fix, it also needs a test asserting that a re-wrapped extrinsic is rejected.

---

### 2. `Dilithium87SigOnly` has no positive test

`pallets/pubkey/src/tests.rs:77`

Every success-path test (pallet unit tests and the four e2e tests in `runtime/tests/transactions/sig_only.rs`) uses `Dilithium65`. `Dilithium87SigOnly` appears only in `sig_only_fails_on_parameter_set_mismatch`, where the expected result is `false`. A copy-paste slip in that arm — `verify_ml_dsa_65` instead of `verify_ml_dsa_87`, or matching the wrong `DilithiumSigner` variant — would leave the whole suite green while ML-DSA-87 sig-only transactions were silently unusable (or worse). This is the chain's higher-security parameter set; please mirror at least `sig_only_verifies_against_cached_pubkey` for it.

### 3. ~2 KB of undeposited permanent state per transacting account

`pallets/pubkey/src/lib.rs:79`

The PR notes this and correctly argues there is no vector against *other* accounts, but the self-spam arithmetic is worth stating: `EXISTENTIAL_DEPOSIT = MILLI_UNIT` (1e9), so the standing cost of parking a 1952–2592 B `Pubkeys` entry is 0.001 UNIT plus one transaction fee — and per finding 1's fee analysis, a miner gets that fee back in the same block's reward. That's roughly a 20× state-bloat amplification versus the ~100 B `System::Account` entry that the same ED bought before this PR. `OnKilledAccount` bounds the map to live accounts, which is the right hook, but it does not bound cost-per-byte. A deposit (or a `Pubkeys` entry priced into the first registration) would; at minimum this deserves a decision rather than a follow-up bullet, because it is not retrofittable to already-registered accounts without a migration.

### 4. Cleanup weight is a hand-maintained shim rather than registered where the work happens

`runtime/src/configs/mod.rs:251`

`BalancesWeightsWithPubkeyCleanup` enumerates five kill-capable balances calls. I checked the alternatives and could not find a currently-reachable reap path it misses — `hold`/`transfer_on_hold` (the reversible-transfers execution path, `pallets/reversible-transfers/src/lib.rs:943`) uses `Preservation::Protect` internally and provably leaves ED behind, so it cannot reap; `withdraw_fee`, vesting, treasury and recovery all use `Preserve`/increase balances; `Utility`/`Scheduler`/`Multisig`/`Recovery` sum the inner call weight. So this is correct today, not a bug.

It is fragile by construction, though: the invariant "every path that can call `OnKilledAccount` charges one write" is enforced nowhere, and the next pallet or upstream `pallet_balances` bump that adds a kill path will silently underweight it. `frame_system::Pallet::<T>::register_extra_weight_unchecked(PubkeyCleanupWeight, DispatchClass::Normal)` inside `Pallet::on_killed_account` (available at `pallets/frame-system/src/lib.rs:1956`) charges it exactly where the write happens and is path-complete by construction. Also note `docs/RUNTIME_SURFACE.md` and the doc comment at `runtime/src/configs/mod.rs:245` say "every call that can reap an account", which is broader than what the shim actually covers — please narrow the wording to "every kill-capable **balances** call" if you keep this approach.

### 5. The public key is copied on every full-signature verification, cached or not

`pallets/pubkey/src/lib.rs:174`

`carried_signer()` calls `sig_public.public()`, which is `ByteArray::from_slice(...)` — a fresh 1952/2592-byte copy (`primitives/dilithium-crypto/src/scheme_macro.rs:106`) — and `note_pubkey` then discards it whenever `contains_key` is true, i.e. on every transaction after the account's first. Checking `Pubkeys::<T>::contains_key(signer)` before extracting avoids the copy on the overwhelmingly common path. The `if let Some(public)` dance also disappears if you bind the signer in the match arm instead of re-matching inside `carried_signer()`.

### 6. Nothing observable marks a registration

The pallet emits no event, and there is no runtime API for `pubkey_of`. Since the documented client rule is "switch to sig-only only after the registering transaction is in a block", wallets have to raw-poll the `Blake2_128Concat` storage key rather than react to a block subscription. A `PubkeyRegistered { who }` event (or exposing `pubkey_of` through a runtime API) would make the opt-in path usable without special-casing storage layout in every SDK.

### 7. Minor

* `runtime/src/lib.rs:116` — hardcoding `AccountId = sp_runtime::AccountId32` drops the compile-time link to `<Signature as Verify>::Signer`. Same type today; a future signer change would now diverge silently instead of failing to build.
* `runtime/src/configs/mod.rs:253` — `pallet_balances::weights::SubstrateWeight::<Runtime>::` is spelled out 11 times; a `type Stock = pallet_balances::weights::SubstrateWeight<Runtime>;` alias would make the four surcharged methods stand out.
* `pallets/pubkey/Cargo.toml:41` — the `try-runtime` feature lists only `frame-support/try-runtime`; `frame-system/try-runtime` and `sp-runtime/try-runtime` are the usual companions.
* `runtime/src/configs/mod.rs:100` — `PubkeyCacheVerifyWeight` reaches for `RocksDbWeight` directly instead of `<Runtime as frame_system::Config>::DbWeight`; identical today, but it silently stops tracking if `DbWeight` is ever retargeted.
* PR description says "existing clients need no changes". The wire format is indeed unchanged, but `transaction_version` 3 → 4 and the new `CachedSignature` metadata type mean any client that pins `transaction_version` or caches metadata must refresh. Worth stating explicitly in the release notes.

---

Requesting changes on finding 1 — as it stands, block authors are paid to undo the size saving this PR exists to deliver, and users get charged for it. The rest are non-blocking, though 2 is cheap and I'd take it in the same pass.

n13 (changes requested):
**Reviewer model:** Opus 5

Reviewed the full diff (18 files, ~1.2k lines) against the codebase at `8b071b1` and ran the affected suites locally.

The core design holds up well: full-signature verification proves both the ML-DSA signature and `Poseidon(pubkey) == account` before the entry is written, so the cache binding is sound; `SigOnly` is domain-separated so a signature is only valid in the encoding form its author chose; `OnKilledAccount` closes the orphan-state loop; and the verify-path DB work is charged in `base_extrinsic` because it happens in `UncheckedExtrinsic::check`, outside any weighted dispatch. I could not find a way to register a key for an account whose key I don't hold, to re-encode an extrinsic across forms, or to get a persistent `Pubkeys` write out of an extrinsic that fails a `TxExtension` (`sc-block-builder::push` rolls the whole `apply_extrinsic` back).

One blocking issue: a test this PR introduces fails.

---

### 1. `runtime/tests/transactions/sig_only.rs:380` — new weight test fails deterministically (blocking)

`first_registration_plus_reap_weight_covers_combined_db_ops` fails on this branch. Reproduced with the exact CI invocation (`SKIP_WASM_BUILD=1 cargo test --locked --features quantus-runtime/fast-governance`), three times, identical numbers:

```
assertion `left == right` failed: a reaping transfer_all must register exactly
the Pubkeys::remove write on top of an otherwise identical non-reaping one
  left:  Weight { ref_time: 115050000, proof_size: 0 }
  right: Weight { ref_time: 100000000, proof_size: 0 }
```

The `Check & Test Matrix` jobs were still pending when I reviewed; they will go red.

Root cause: the two `transfer_all` extrinsics are *not* "otherwise identical" in registered block weight, independently of the reap. `WormholeProofRecorderExtension::post_dispatch` (`runtime/src/transaction_extensions.rs:471`, `:516`) registers `event_scan_weight(events_at_scan, …)` via `register_extra_weight_unchecked`, and `events_at_scan` is the **cumulative** block event count — so the second extrinsic in a block always pays for re-scanning the first one's events.

I confirmed this by measuring two identical **non-reaping** `transfer_all` extrinsics from two different funded accounts, back to back in one block:

```
first  = 5_824_357_000
second = 5_838_007_000     delta = 13_650_000  (no reap involved at all)
```

The 15_050_000 residual in the failing assertion is that same positional effect, plus the extra `KilledAccount`/reap events the second extrinsic emits (`15_050_000 = 10 events × 1µs + 505 bytes × 10ns`, matching `EVENT_SCAN_DECODE_REF_TIME_PS`/`EVENT_SCAN_DECODE_BYTE_REF_TIME_PS` exactly).

So the mechanism under test is fine — `on_killed_account` really does register exactly one DB write, and the pallet unit test `killed_account_registers_cleanup_weight` proves it — but the e2e differential is confounded. Options: apply each extrinsic as the *first* extrinsic of its own `TestExternalities`, or relax to `assert!(reap.saturating_sub(keep).all_gte(cleanup))`, or subtract a baseline measured from a third identical non-reaping extrinsic.

---

### 2. `pallets/pubkey/src/lib.rs:135` — cleanup weight is always billed to `DispatchClass::Normal`

`register_extra_weight_unchecked(T::DbWeight::get().writes(1), DispatchClass::Normal)` is unconditional, but `on_killed_account` fires on every reap path — including reaps inside `Operational` dispatches and inside `on_initialize`/`on_finalize` (Mandatory), e.g. a scheduler-enacted `transfer_all`. Those reaps then consume the `Normal` class budget (75% of the block) that they did not use, shrinking the room left for user transactions in that block.

The block *total* stays correct, so this is attribution rather than a soundness hole, and the class isn't available inside `OnKilledAccount` — but the doc comment above the hook (which argues the choice is "path-complete by construction") should say that the class attribution is deliberately approximate.

Related, minor: the write is registered on *every* reap, including accounts that never registered a key (dusted receive-only accounts). That is an over-charge, not an under-charge, so it is safe — just worth knowing it makes every reap on the chain 100ms of block weight more expensive.

---

### 3. `primitives/dilithium-crypto/src/traits.rs:162` — the perf commit's copy is still made three times on the same path

Commit `8e1a5ee` ("extract the carried pubkey only when the cache entry is written") moves one 1.9–2.6 KB copy behind a closure, but the `self.0.verify(msg, signer)` call that immediately precedes it still makes three:

```rust
let account = sig_public.public().clone().into_account();   // :162 — copy 1 (public()) + copy 2 (.clone())
...
    sig_public.public().as_ref(),                            // :167 — copy 3
```

`public()` returns an **owned** `Dilithium65Public`/`Dilithium87Public` (`scheme_macro.rs:106` rebuilds it from a slice), so the `.clone()` at `:162`/`:173` is dead weight, and `.public().as_ref()` at `:167`/`:178` allocates a fresh copy just to borrow it. Binding `let public = sig_public.public();` once per arm and reusing it takes the full-signature verify path from four copies to one. Not blocking, but it makes the stated perf win mostly notional.

---

### Non-blocking notes

- **Metadata shape.** `ExtrinsicMetadata::signature_ty` changes from the `DilithiumSignatureScheme` enum to a single-unnamed-field composite (`CachedSignature`; scale-info drops the `PhantomData` field). SCALE bytes are byte-identical and `transaction_version` is bumped, and both subxt and polkadot-js unwrap newtype composites — but it's worth decoding one extrinsic with the actual wallet/SDK against the new metadata before release, since nothing in the test suite covers the tooling side.
- **Pool-validation cost.** A bogus `SigOnly` extrinsic aimed at an already-registered account is ~37% smaller on the wire yet still forces a full ML-DSA verify plus a ~2.6 KB `Pubkeys` read in `validate_transaction`, which is never fee-charged. It's a bounded (≤1.6×) amplification of a surface that already exists for full signatures, not a new class of problem — but sig-only makes the cheapest form of that spam cheaper.
- **Unsigned extrinsics pay the surcharge.** `PubkeyCacheVerifyWeight` is added to `base_extrinsic` for all of `DispatchClass::non_mandatory()`, so unsigned `Normal`/`Operational` transactions would pay for a `Verify` they never run. Currently moot — I found no `ValidateUnsigned` in the runtime — so this is only a note for whoever adds the first one.
- **`BlockWeights` rewrite.** I diffed the hand-rolled builder against `with_sensible_defaults` in `pallets/frame-system/src/limits.rs:323`: same `max_total`/`reserved`, and `avg_block_initialization(Perbill::from_percent(10))` matches upstream's hard-coded 10%. The only delta is the intended `base_extrinsic` surcharge. No behavior drift.
- **No-deposit decision.** Documented and explicitly accepted, so just recording the number: at `ED = 1 mUNIT` and `LENGTH_FEE_MULTIPLIER = 1e6`, a registration costs ~6.3 mUNIT (≈5.3 mUNIT length fee + ED) for ~2.6 KB of permanent state — roughly 2.4× the price of a transient byte. Cheap, but the same order of magnitude as what the registering transaction already pays for bandwidth, and bounded by `OnKilledAccount`.

### Test results

- `pallet-pubkey`: 15/15 pass.
- `quantus-runtime --test mod`: 48 pass, **1 fail** (finding 1), 1 ignored.

---

**Verdict: Request changes** — the blocker is finding 1; findings 2 and 3 are follow-ups that don't need to gate the merge.

n13 (changes requested):
**Reviewer model:** GPT Sol

Verdict: REQUEST_CHANGES

The latest revisions resolve the earlier signature-binding, public re-export, API, and unsigned-base-weight issues. One runtime-accounting blocker remains.

- **Blocker — pubkey cleanup weight is neither charged to the submitter nor admitted before dispatch.** `pallets/pubkey/src/lib.rs:140-146` calls `register_extra_weight_unchecked` from `OnKilledAccount`. That updates `System::BlockWeight`, but it does not update the outer call's `PostDispatchInfo`. Transaction payment computes the corrected fee from `post_info.calc_actual_weight(info)`, so a reaping transaction still pays only the generated call weight; the `Pubkeys::remove` write is absent from its fee. The same weight is registered only after `CheckWeight` admitted the extrinsic, so it cannot prevent a boundary overshoot either.

  The comment's one-write-per-extrinsic assumption also does not hold under call composition. A signed `utility.batch` can reap several funded derivative accounts in one outer extrinsic. In a focused runtime probe with three such `transfer_all` calls, dispatch returned an actual weight of `1,254,068,680` ref-time while the hook separately registered `300,000,000` ref-time (three DB writes). The returned post-dispatch weight never included those writes. The existing tests establish block-weight registration, but do not assert the payer's `TransactionFeePaid` amount or a pre-dispatch bound for multiple reaps.

Please make the cleanup cost part of compositional declared/actual transaction weight (or otherwise precharge a sound bound that `CheckWeight` and transaction payment both see), then refund only genuinely unused weight. Keep the lifecycle hook if needed for state correctness, but do not rely on post-hoc unchecked registration as fee/admission accounting. Add regressions for the reaping fee and a wrapper that performs multiple reaps.

Validation at head `6e67ac75096776260be7134d4d08011d8d97ac8e`: `git diff --check`; `cargo +nightly fmt --all -- --check`; `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-pubkey` (15 passed); `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime --test mod --features fast-governance` (49 passed, 1 ignored). All four GitHub checks are green.

n13 (changes requested):
**Reviewer model:** GPT Sol

Verdict: REQUEST_CHANGES

The latest revision fixes the direct and `Utility`-wrapped cleanup accounting: visible reap-capable calls reserve the cleanup before `CheckWeight`, realized reaps stay in the payer's corrected weight, and unused reservations are refunded. The signature binding, domain separation, cache lifecycle, and direct-path tests also hold up. One compositional accounting blocker remains.

### Blocker — stored-call execution still bypasses cleanup admission and fees

`runtime/src/transaction_extensions.rs:58-68,116,157-174`; `pallets/multisig/src/lib.rs:1123-1181`

`count_reaps` explicitly treats stored inner calls such as `Multisig::execute` as opaque and therefore returns zero for the top-level call. But `execute` decodes the stored arbitrary `RuntimeCall` and dispatches it as the multisig account, so a proposal can contain the same `Utility::batch_all` of derivative `transfer_all(keep_alive = false)` calls that the new direct-path regression covers.

For that reachable path, every reap increments `ExtrinsicReaps`, but `post_dispatch_details` only calls `register_extra_weight_unchecked` after dispatch. That cannot inform `CheckWeight`'s pre-dispatch admission, and it does not add cleanup weight to `PostDispatchInfo`, so transaction payment bills none of those database operations. The new `statically_invisible_reap_is_registered_against_the_block` test actually pins the latter behavior at `runtime/src/transaction_extensions.rs:2494-2529`: block weight increases, while `post_info.actual_weight` remains unchanged. With a stored call up to `MaxCallSize = 10 KiB`, one `execute` can compose many such reaps, so this is not a one-write-only edge case.

Please reserve a sound upper bound for opaque dispatch wrappers such as `Multisig::execute` before `CheckWeight`, then use the existing counter to refund unrealized reservations, or make the wrapper's declared/actual weight include the cleanup work. Add a full-pipeline regression with a real multisig proposal containing multiple reaping calls that asserts both declared admission weight and the payer's `TransactionFeePaid` delta.

Validation at head `4fe6b3cfd32413bcf74903b2d67104f8e243f0cc`: `git diff --check`; `cargo +nightly fmt --all -- --check`; `SKIP_WASM_BUILD=1 cargo test --locked -p qp-dilithium-crypto` (18 passed); `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-pubkey` (16 passed); `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime --test mod --features fast-governance` (51 passed, 1 ignored). GitHub Fast Checks passed; the remaining three CI jobs were still pending at review time.

n13 (changes requested):
**Reviewer model:** GPT Sol

Verdict: REQUEST_CHANGES

The latest revision closes the previously reported stored-call accounting gap: `Multisig::execute` now carries and byte-binds the inner call, so the reap and transfer matchers can pre-charge its side effects. The signature binding, sig-only domain separation, pubkey lifecycle, and direct/wrapped cleanup fee paths also hold up. Two correctness issues remain in the new execute contract.

### 1. Accepted proposal bytes can be impossible to execute

`pallets/multisig/src/lib.rs:594-600,1135-1142`

`propose` calls `RuntimeCall::decode(&mut &call[..])`, but never checks that decoding consumed the complete bounded byte vector. A payload formed as `valid_call.encode() || trailing_bytes` therefore passes the validity, weight, and high-security checks and is stored unchanged; `approve` can approve those exact stored bytes. The new `execute` API accepts a typed `RuntimeCall` and requires `call.encode()` to equal the complete stored slice, so no submitted call can ever match that accepted payload. The proposal can reach `Approved`, burn its proposal fee, and reserve its deposit, yet every execution returns `CallMismatch`.

Please enforce a canonical full decode at proposal creation (for example, require no remaining input and `decoded_call.encode() == call`, or store the canonical encoding), and add a regression that appends a byte to a valid encoded call and verifies `propose` rejects it without creating proposal state or charging the deposit/fee.

### 2. A mismatched execute can report more weight than it declared

`pallets/multisig/src/lib.rs:1070-1075,1111-1116,1140-1142`

The pre-dispatch weight sizes bookkeeping from the submitted call only. After loading the proposal, the mismatch path sizes its returned `actual_weight` from `max(stored_len, submitted_len)`. For a large stored proposal and a small mismatching call, that post-dispatch weight exceeds the declared weight. FRAME explicitly forbids post-dispatch weight increases: `PostDispatchInfo::calc_actual_weight` logs the invariant violation and clamps the larger components to the pre-dispatch value. Because `CallMismatch` leaves the proposal in storage, a signer can repeat this path.

Reserve a sound bookkeeping bound before dispatch (the prior `MaxCallSize` reservation is sufficient) and refund to the real size afterward, or otherwise ensure every error path's reported weight is no greater than its declaration. The mismatch regression should assert that invariant for a max-size stored call and a small submitted call.

Validation at head `e5934565545709a8cdd06c833417e1d9a4f930dd`:

- `git diff --check 803094f7...e593456` — passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p qp-dilithium-crypto` — 18 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-pubkey` — 16 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-multisig` — 57 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime --test mod --features fast-governance` — 52 passed, 1 ignored
- All four current GitHub checks are green.

n13 (approved):
**Reviewer model:** GPT Sol

Verdict: APPROVE

Reviewed the complete 26-file diff against base `803094f7` at head `a86becfa`. No blocking findings remain.

The latest revision closes both previously reported multisig issues: `propose` now rejects any stored payload that is not the decoded call's exact canonical encoding before proposal state, deposit, or proposal-fee changes; and `execute` reserves bookkeeping at `MaxCallSize`, so a small mismatching submitted call cannot make a large stored proposal report more post-dispatch weight than it declared. The new regressions cover both cases.

The broader design also holds up under adversarial review: full signatures prove the ML-DSA signature and `Poseidon(pubkey) == account` before caching; sig-only signatures are domain-separated from full signatures; account reaping removes the cache; signed-only extension weight covers verification storage work; and direct, batched, derivative, recovery, utility, and resubmitted multisig call paths pre-charge and reconcile every realized cleanup for admission and fees.

Non-blocking release-note caveat: the PR body's broad "existing clients need no changes" statement should be narrowed. Full-signature clients retain their wire encoding, but clients invoking `Multisig::execute` must refresh metadata and supply the newly required inner `call` argument.

Validation at `a86becfa1af4ec79560d72681a844921010e8469`:

- `git diff --check 803094f7...a86becfa` — passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p qp-dilithium-crypto` — 18 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-pubkey` — 16 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-multisig` — 59 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime --test mod --features fast-governance` — 52 passed, 1 ignored
- `SKIP_WASM_BUILD=1 cargo check --locked -p quantus-runtime --features runtime-benchmarks,try-runtime` — passed

GitHub Fast Checks and Analysis are green; both platform matrix jobs are still pending at review time.

PR #662, opened 2026-08-11
QUIC auth

illuzen:
## Summary

- Add shared-secret authentication to the external miner QUIC channel: miners must send `Ready { token }` before they are registered.
- Persist a self-signed miner TLS cert under `<base-path>/chains/<chain>/` and publish its SHA-256 fingerprint (`miner-tls-cert-sha256`) for miners to pin.
- Auto-create the auth token file (`miner-auth-token`, override with `--miner-auth-token-file`); log the file path only (never the token). Log the TLS fingerprint for pinning.
- If `--miner-listen-port` is set and the miner server fails to start, abort instead of falling back to local mining.
- Bump `quantus-miner-api` to `0.3.0` (breaking protocol change) and update mining docs + miner-stack Compose to share token + fingerprint files with the miner.

## Companion PR

Requires a matching update in [quantus-miner](https://github.com/Quantus-Network/quantus-miner) so clients send the token and pin the TLS cert fingerprint (`--auth-token` / `--tls-cert-sha256`, or the `*-file` variants).

## Test plan

- [ ] Start a validator with `--miner-listen-port 9833` and confirm logs show the auth **token file path** (not the token value) + TLS fingerprint; files exist under the chain config dir (`miner-auth-token`, `miner-tls-cert.der`, `miner-tls-key.der`, `miner-tls-cert-sha256`).
- [ ] Restart the node and confirm the same token/cert/fingerprint are reused (not regenerated).
- [ ] Connect a miner with the correct token + fingerprint and confirm it registers and receives jobs.
- [ ] Connect with a wrong/missing token and confirm rejection before miner registration.
- [ ] Connect with a wrong TLS fingerprint and confirm the QUIC handshake fails.
- [ ] Point `--miner-auth-token-file` at an empty/unreadable path and confirm the node exits (does not mine locally).
- [ ] (Optional) `miner-stack`: bring up compose and confirm the miner reads token + fingerprint from `node-data/chains/<chain>/`.


n13 (changes requested):
**Reviewer model:** GPT Sol

Verdict: REQUEST_CHANGES

The pre-registration authentication flow and atomic `0600` token creation are directionally sound, but two blocking issues remain.

1. **[P1] Do not emit the miner credential to logs** (`node/src/miner_server.rs:83`). The token is the only credential required to register as a miner, yet it is written verbatim at info level on every startup. That persists the secret in stdout, journald/Docker logs, and any log aggregation, defeating the narrower access boundary provided by the `0600` file. Anyone with log access who can reach the QUIC port can authenticate, observe jobs, and submit/flood results. Log the token-file path only and require operators to read the protected file; do not log the secret itself.

2. **[P1] Fail external-miner startup instead of switching to local mining** (`node/src/service.rs:528`). `load_or_create_miner_auth_token` adds new startup failures for an empty/unreadable file, an invalid override path, or an unwritable default directory. `MinerServer::start` returns those errors, but this call converts every error to `None`; `mining_loop` then enters `handle_local_mining`. A validator configured with `--miner-listen-port` therefore burns CPU mining locally even though the CLI/docs say local mining is disabled in that mode. Propagate the startup error (or terminate the essential service) so invalid auth configuration fails early rather than changing mining mode.

Validation on exact head `96a2f5e5323a4e21e52df856c684671fb76ab365`:

- `git diff --check 803094f7...96a2f5e5` — passed
- `cargo test --locked -p quantus-node miner_server::tests -- --nocapture` — 6 passed
- `cargo test --locked -p quantus-miner-api` — passed (no unit/doc tests defined)
- `cargo clippy --locked -p quantus-miner-api --no-deps -- -D warnings` — passed
- `SKIP_WASM_BUILD=1 cargo clippy --locked -p quantus-node --no-deps` — passed with existing workspace warnings
- GitHub Fast Checks, Clippy/Doc, and miner-stack Compose config — passed; Linux/macOS test matrices were still running at review time

Local Compose rendering was unavailable because the Docker CLI is not installed in the review environment; the dedicated GitHub Compose check passed.

n13 (changes requested):
**Reviewer model:** GPT Sol

Verdict: REQUEST_CHANGES

The two earlier P1 findings are resolved on this head: the shared secret is no longer logged, and miner-server initialization failure now ends the essential mining task instead of falling back to local mining. One blocking documented-use compatibility issue remains.

1. **[P2] Keep the remote-miner Docker example usable after authentication becomes mandatory** (`miner-stack/README.md:364`). The command still supplies only `--node-addr`, worker count, and metrics port. The companion 0.3.0 miner requires both an auth token and a TLS certificate fingerprint, so this copy-paste path exits before connecting. The prose below names the new flags, but file paths also are not available inside this separate container unless they are mounted. Add read-only mounts plus `--auth-token-file` and `--tls-cert-sha256-file`, or show the two direct-value flags, in the command itself.

Validation on exact head `927b6e16ac3945e7e18392916b657a100cfec290`:

- Inspected the full `803094f7...927b6e16` diff, the three commits after the prior review, and the companion miner's matching `Ready { token }` / SHA-256 pinning implementation.
- `git diff --check 803094f7...927b6e16` — passed.
- `cargo +nightly fmt --all -- --check` and `taplo format --check --config taplo.toml` — passed.
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-node miner_server::tests -- --nocapture` — 7 passed.
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-miner-api` — passed (no unit/doc tests defined).
- `SKIP_WASM_BUILD=1 cargo clippy --locked -p quantus-node -p quantus-miner-api --no-deps` — passed with warnings; strict `-D warnings` still reports the new private-interface visibility warning plus two pre-existing node lints.
- `cargo clippy --locked -p quantus-miner-api --no-deps -- -D warnings` — passed.
- GitHub Fast Checks, Linux/macOS test matrices, Clippy/Doc, and miner-stack Compose config — all passed.

Local Compose rendering was unavailable because the Docker CLI is not installed; the dedicated GitHub Compose check passed.

n13 (changes requested):
**Reviewer model:** Fable 5

Verdict: REQUEST_CHANGES

Solid direction: authenticating miners before registration, persisting the TLS cert + fingerprint, and failing closed instead of falling back to local mining are all the right calls, and the token is correctly kept out of the logs. The auth match arms are ordered so `Ok(other)` can never bind a `Ready`, so the `{:?}` log never prints the token. But several issues in the bundled deployment and the fail-closed path are blocking.

## Blocking

**1. The bundled compose + docs drive an incompatible miner → zero mining** (`miner-stack/docker-compose.yml:52-53`, `miner-api/src/lib.rs:33`). This PR makes `Ready` a struct variant (breaking JSON wire change) and the compose command now passes `--auth-token-file` / `--tls-cert-sha256-file` to `ghcr.io/.../quantus-miner:${MINER_VERSION:-latest}`. I checked `quantus-miner` `origin/main`: it has neither flag, still sends the unit `MinerMessage::Ready` (`crates/quic-transport/src/lib.rs:36`), and pins `quantus-miner-api` to tag `v0.7.1-q-day-2`. So `:latest` either exits on the unknown argument or fails deserialization, and with `restart: unless-stopped` it crash-loops. Because `--miner-listen-port` also disables local mining, the documented stack produces **no blocks** until a matching miner ships. A coordinated breaking change is fine, but the default deployment shouldn't be broken on merge: pin `MINER_VERSION` to a compatible release, state a minimum miner version in the docs, and consider bumping the ALPN id (currently unversioned `quantus-miner`) or adding a version field so a mismatch fails at handshake with an actionable message instead of an opaque `Failed to read Ready from miner …`.

**2. The miner container is granted the node's private keys** (`miner-stack/docker-compose.yml:61`). The node's base path is `/var/lib/quantus` (`./node-data`), and the miner mounts the whole thing read-only (`./node-data:/node-data:ro`) just to read two files. That exposes `chains/<chain>/network/secret_ed25519`, the keystore, and the newly-persisted `miner-tls-key.der` to the miner process — which also runs untrusted GPU/CPU workloads. A compromised miner image can then impersonate the node's peer identity and its miner TLS endpoint. Mount only `miner-auth-token` and `miner-tls-cert-sha256`.

**3. `--miner-listen-port` is a silent no-op on non-validators** (`node/src/service.rs:815`). The entire miner path — token/cert creation, the listener, and the new fail-closed abort — is inside `if role.is_authority()`. Start a node with `--miner-listen-port 9833` (and `--miner-auth-token-file`) but without `--validator` and nothing binds, no token file is written, and there is no warning — directly contradicting `cli.rs:23-25` ("Startup fails if this path is empty/unreadable…") and MINING.md's "the node exits" claim. At minimum warn when the flags are set for a non-authority.

**4. A half-written or stale secret file bricks the whole node** (`node/src/miner_server.rs:248-296`, `node/src/service.rs:533`). The essential-task `return;` doesn't just stop mining — sc-service treats any essential-task completion as failure, so the whole process exits (P2P/sync/RPC included), and under `restart: unless-stopped` that's a permanent crash loop. Reachable states: cert written but key not (SIGKILL/ENOSPC between the two `create_new` writes) → `cert_exists != key_exists` error forever; a 0-byte token from an interrupted first write → the "is empty" error forever; a leftover process holding UDP/9833. The old code stayed up and mined locally. Write secrets via temp-file-plus-rename (atomic), or clean up the partial cert on key-write failure, so an interrupted start is recoverable. Relatedly, `miner-tls-cert-sha256` is overwritten from the cert bytes at `miner_server.rs:296` *before* the cert is validated by `with_single_cert`/`QuicServerConfig::try_from`, so a corrupt-but-nonempty cert destroys the fingerprint operators already distributed to miners.

**5. `${CHAIN}` is the `--chain` argument, not the chain-spec id** (`miner-stack/docker-compose.yml:52-53`, `MINING.md:~299`). The node writes under `base_path.config_dir(chain_spec.id())`, but `load_spec` (`node/src/command.rs:305-321`) maps e.g. `planck_live_spec → id "planck"` and `heisenberg_live_spec → id "heisenberg"`, and a spec-file path takes its id from the file. Default `CHAIN=planck` happens to match id `planck`, but any other value makes the miner read `chains/<arg>/…` while the node wrote `chains/<id>/…` → miner crash-loops on a missing file. Document that `<chain>` is the spec id, or derive both from the same value.

## Should fix

**6. Pre-auth DoS surface** (`node/src/miner_server.rs:438`, transport config at `:382`). The handshake read has no `tokio::time::timeout`, there's no cap on concurrent unauthenticated connections (`ServerConfig::concurrent_connections` left at the 100k default), and because the server sends keep-alives every 10s the 60s `max_idle_timeout` never fires — so a peer that connects and sends nothing pins a tokio task + connection forever. Per-connection stream limits are also at quinn defaults (100 bidi + 100 uni × ~1.25 MB), and the app only ever `accept_bi`s one stream, so unread stream data buffers in node memory pre-auth. MINING.md now advertises DoS resistance the code doesn't provide. Add a handshake timeout and an unauthenticated-connection cap.

**7. File permissions vs. the miner UID** (`node/src/miner_server.rs:335`, `miner-stack/README.md:~365`). Token/cert/key/fingerprint are all created 0600 owned by the node image's `USER 10001:10001`. In-compose this works only because the miner image also happens to run as 10001; the documented "copy the token/fingerprint to a remote miner host" path lands the files as a different owner and the miner gets EACCES. The fingerprint is public data and doesn't need 0600. Document the UID requirement (or relax the fingerprint mode), and note that "files don't exist yet" in the README can actually be a permission failure.

**8. Trailing-newline mismatch in the token** (`node/src/miner_server.rs:439`). The node writes the token file with a trailing `\n` and `trim()`s on read, but compares the wire-supplied token byte-exact. A miner that reads its `--auth-token-file` with `read_to_string` and sends it un-trimmed transmits `"<hex>\n"` and is rejected with only "invalid auth token". Trim the received token server-side (or write the file without the newline) to kill an entire class of silent cross-repo auth failures.

**9. Existing token perms not verified on load** (`node/src/miner_server.rs:187`). The 0600 guarantee only applies to files the node creates; a token restored from backup or dropped in at 0644 is read and trusted with no warning, while the CLI help still claims "mode 0600 on Unix". On non-Unix (`open_secret_file` at `:346`) the token and TLS key are created with default ACLs (world-readable) with no warning. Consider checking/repairing the mode on load, or at least warning.

**10. Hardening notes** (non-exhaustive): the token compare at `miner_server.rs:439` is a short-circuiting `==` (not constant-time) with no failed-auth rate limit/backoff and an unthrottled `warn!` per attempt — low risk for the 256-bit default token but worth a constant-time compare + a counter/close on failure. `MinerMessage` derives `Debug` while now carrying the secret (`miner-api/src/lib.rs:33`); no current path logs it, but since this is a published crate a redacting `Debug`/secret newtype would make the "never logged" property hold for all consumers. An operator-supplied token longer than `MAX_MESSAGE_SIZE` (1024) silently makes every miner fail auth with an opaque error.

**11. Docs: `node-data/` is still "safe to delete"** (`miner-stack/README.md:213`, reset recipe at `:342`). The `rm -rf node-data/` recipe now also wipes `miner-auth-token` and the TLS cert/key/fingerprint; the node regenerates them, but every external miner that pinned the old values silently stops with only "invalid auth token" on the node. Note that these must be re-copied after a reset.

## Cleanup (non-blocking)

- **DRY** (`node/src/miner_server.rs:221-346`): `write_secret_text_file`, `write_secret_text_file_overwrite`, and `write_secret_bytes_file` are three near-identical bodies differing only in a bool and a trailing newline, `write_miner_auth_token_file` is a one-line passthrough, and `open_secret_file` duplicates its whole body across `#[cfg(unix)]`/`#[cfg(not(unix))]` for one `.mode()` call. Collapse to one `write_secret_file(path, bytes, create_new)` and set the mode conditionally on a single `OpenOptions`. This repo also already vendors `sc_network::config::{get_secret, write_secret_file, open_secret_file}` (`client/network/src/config.rs:453-511`) implementing exactly this read-or-generate-0600-secret flow (and `get_secret` zeroizes the buffer, which the new code doesn't) — they're private, but exposing and reusing them would remove the whole hand-rolled block. This overlaps the user's "strictly DRY / keep code minimal" rules.
- **`private_interfaces` lint** (`node/src/miner_server.rs:186,237`): `pub fn load_or_create_miner_auth_token` / `load_or_create_miner_tls` return the private `MinerTlsMaterial` and are only called in-module — drop `pub` (this is the strict `-D warnings` lint prior CI flagged).
- **Reuse existing state** (`node/src/service.rs:773`, `node/Cargo.toml:84`): `Configuration.data_path` already holds `config_dir(chain_spec.id())`, so recomputing it risks drift from every other chain-scoped path; and the new `sha2` dependency is unnecessary since `sp_io::hashing::sha2_256` (already a node dep) computes the fingerprint.
- **Encode the invariant in a type** (`node/src/service.rs:461,529`): the three parallel `Option`s (`miner_listen_port`, `miner_auth_token_path`, `miner_tls_dir`) express one all-or-nothing invariant re-asserted 250 lines later by two `expect()`s. A single `Option<MinerListenConfig { port, auth_token_path, tls_dir }>` makes it a compile-time fact and removes both `expect()`s. Also: `auth_token: Arc<str>` is redundant (the server is always held as `Arc<Self>`), and `create_server_endpoint` clones both DER buffers from a struct that's dead after the call (leaving a second copy of the private key in memory).

Verdict: **REQUEST_CHANGES** — blocking issues in the bundled deployment (incompatible miner, private-key exposure via the mount, chain-id path mismatch) and the fail-closed path (silent no-op off-validator, unrecoverable half-written state taking down the whole node).

n13 (changes requested):
**Reviewer model:** Fable 5

Verdict: REQUEST_CHANGES

This is a solid revision. The prior blocking items are addressed: the token is kept out of logs (redacted `Debug` + path-only logging), miner-server startup now fails closed instead of falling back to local mining, the non-validator case at least warns, secrets are written atomically at 0600, the fingerprint is only published after the cert validates, and the incompatible bundled Compose stack was removed entirely. Auth is checked before registration, the match arms are ordered so `Ok(other)` can never bind a `Ready` (token never reaches the `{:?}` log), and the compare is constant-time. Compile/tests are clean (all 8 `miner_server` tests pass; CI clippy is warnings-only).

Two issues are still blocking, both with trivial fixes.

## Blocking

**1. Deterministic temp-file name + `create_new` can brick the node in a permanent crash loop** (`node/src/miner_server.rs:421-425`, trigger at `:341`, shutdown at `node/src/service.rs:542`). `atomic_write_bytes_file` names its temp file `.<filename>.<pid>.tmp` and opens it with `.create_new(true)` (`:458`/`:472`). The fingerprint file is rewritten through this path *unconditionally on every start* (`:341` runs even when cert/key already exist and the value is unchanged — the `if let Ok(existing)` block above only logs). If the node is SIGKILLed/OOM-killed/loses power in the window between the temp file being created (`:429`) and `fs::rename` (`:436`) — a window that includes `write_all` + `sync_all()`, and fsync is slow under exactly the I/O pressure that triggers OOM kills — a stale `.miner-tls-cert-sha256.<pid>.tmp` is left behind (cleanup only runs on graceful error paths, never on a hard kill). On the next start with the same PID, `create_new` fails with "File exists", the error propagates `load_or_create_miner_tls → MinerServer::start → return;` in the essential `qpow-mining` task, and Substrate shuts the whole node down. Under Docker with an exec-form entrypoint the node is PID 1 on every restart, so the temp name collides deterministically once a stale file exists on the persisted TLS volume — a self-perpetuating crash loop, with no error message telling the operator the fix is to delete a hidden dotfile. This is in the very code added to make writes crash-safe. Fix: use a randomized/unique temp suffix (or `O_TRUNC` instead of `create_new`), and/or only write the fingerprint when it is missing or actually changed.

**2. Pre-auth permit is held across the un-timed QUIC handshake, enabling a cheap miner lockout** (`node/src/miner_server.rs:531-540`). The `unauth_slots` permit is acquired at `:531` and then the task awaits `connecting.await` at `:540` — the QUIC address-validation + TLS handshake, which is **not** wrapped in any timeout. `AUTH_HANDSHAKE_TIMEOUT` (10s) is applied only *inside* `authenticate_miner_connection` (`:603`), i.e. after `connecting` already resolved. With `keep_alive_interval(None)` and `max_idle_timeout(60s)`, a peer that sends an Initial (source can be spoofed, no `retry()` address validation) and then stalls the handshake pins a permit for up to 60s. ~32 such half-open handshakes, refreshed at ~0.5/s (trivial cost), keep all `MAX_UNAUTHENTICATED_CONNECTIONS` slots exhausted, so every legitimate miner hits the `else { connecting.refuse() }` branch indefinitely. Because `--miner-listen-port` disables local mining with no fallback, a squatted validator produces no blocks. This is the gap between the DoS bound the PR advertises (a "10s handshake timeout") and what the code enforces (only the auth *read* is bounded; connection establishment is not). Fix: wrap `connecting.await` in a bounded timeout, or acquire the permit only after the handshake completes.

## Should fix (not independently blocking)

**3. `MAX_AUTH_TOKEN_LEN` validates raw length but the wire frame is JSON-escaped** (`node/src/miner_server.rs:235`, const at `miner-api/src/lib.rs:15`). The 512-byte limit is checked against `token.len()`, but the `Ready { token }` frame is `{"Ready":{"token":"…"}}` with escaping. A ≤512-byte operator token of escape-heavy characters (e.g. 512 `"` → 1024 bytes escaped, +21 overhead = 1045) passes node validation yet makes every miner's frame exceed `MAX_MESSAGE_SIZE` (1024), so all miners fail auth with the exact opaque "Message size exceeds maximum" error the const claims to prevent. Harmless for the 64-hex default token; only bites a pathological hand-written token. Validate `serde_json::to_vec(&Ready{token}).len()` instead.

**4. Server keep-alive removed, but job traffic is event-driven, so idle miners are dropped mid-job** (`node/src/miner_server.rs:504-510`). The comment "Authenticated miners stay alive via job traffic" is inaccurate: `handle_external_mining` broadcasts one `NewJob` per template and then waits silently; qpow rebuilds templates only on block-import/tx notifications. On a quiet or high-difficulty chain a mining round exceeds 60s with zero packets in either direction, and `max_idle_timeout(60s)` tears the connection down. The bundled quantus-miner is safe (5s client keep-alive), but MINING.md's protocol section documents only Ready/NewJob/JobResult — a third-party miner built to spec is silently dropped mid-job and loses the seal. Document the client keep-alive requirement (and fix the comment).

**5. TLS private-key file mode isn't verified on load, unlike the token** (`node/src/miner_server.rs:303-304` vs `:207`). `load_or_create_miner_auth_token` calls `ensure_secret_file_permissions` to warn-and-repair an over-permissive existing token file; `load_or_create_miner_tls` reads the equally-secret `miner-tls-key.der` with plain `fs::read` and never checks its mode. A key restored/copied at 0644 is trusted silently. Defense-in-depth only (requires operator-supplied bad perms + a local co-resident attacker), but the asymmetry is a one-line fix — call `ensure_secret_file_permissions` on the key too.

**6. `--miner-listen-port` on a non-validator warns and is ignored, contradicting the CLI help and repo precedent** (`node/src/service.rs:847`, `node/src/cli.rs:22-25`). The CLI help says "Startup fails if this path is empty/unreadable…", but off-validator the whole path (bind, token/TLS creation, fail-closed abort) is skipped with just a log line. The repo's own precedent for a flag-requires-validator case (`--rewards-inner-hash` in `command.rs`) returns `Err(sc_cli::Error::Input(...))`. `--miner-auth-token-file` without `--miner-listen-port` is dropped with no warning at all. Consider a hard startup error in the `command.rs` run closure where `config.role` is already available.

**7. Wire break with unversioned ALPN → opaque mismatch failure** (`node/src/miner_server.rs:493`, `miner-api/src/lib.rs:41`). `Ready` became a struct variant (breaking JSON change, crate 0.2.1→0.3.0), but ALPN stays `b"quantus-miner"` with no protocol-version signal. An old miner against a new node fails with "Failed to read Ready … invalid type" closed as "auth failed" — indistinguishable from a wrong token. The companion quantus-miner PR (#78, `--auth-token`/`--tls-cert-sha256`) exists and the PR body flags the dependency, so this is a coordinated change; still, an ALPN bump (e.g. `quantus-miner/2`) would turn future mismatches into a clean handshake failure instead of a token red-herring.

## Cleanup (non-blocking)

- **DRY / double work** (`miner_server.rs:318`, `:329`, `:483-497`): `validate_miner_tls` builds a full rustls+QUIC `ServerConfig` and discards it, is called twice on the generate path, and `create_server_endpoint` rebuilds the identical config a third time from the same bytes (ALPN set only in that copy — a real drift risk vs the "validate before publishing the pin" guarantee). Have `validate` return the config / parsed DER and thread it through.
- **Encode the invariant in a type** (`service.rs:460-462`, `:530`/`:532`): three parallel `Option`s (`miner_listen_port`, `miner_auth_token_path`, `miner_tls_dir`) threaded through two call sites and re-joined by two `expect()`s 70 lines later. A single `Option<MinerServerConfig { port, auth_token_path, tls_dir }>` makes it a compile-time fact and drops both `expect()`s.
- **Reuse existing state** (`service.rs:772-773`, `node/Cargo.toml:84`): `config.data_path` already holds `config_dir(chain_spec.id())` — recomputing risks drift from every other chain-scoped path. The new `sha2` dep is avoidable: `sp_io::hashing::sha2_256` is already available to the node crate for the single digest at `miner_server.rs:331`.
- **Unconditional fingerprint rewrite** (`miner_server.rs:341`): full atomic write + fsync every boot even when unchanged (also the recurring trigger for blocking item #1). Write only when missing or changed.
- **`private_interfaces` warning** (`miner_server.rs:277`): `pub fn load_or_create_miner_tls` returns the private `MinerTlsMaterial`; it's only used in-module. Drop `pub` (and on `load_or_create_miner_auth_token`) to silence the warning that fires on every build.
- **Minor**: `auth_token: Arc<str>` inside a type always held as `Arc<Self>` (a plain `String` suffices); `create_server_endpoint` clones both DER buffers from a `tls` that's dead after the call (leaving an extra private-key copy — pass by value and move); the `write_miner_auth_token_file → write_secret_text_file → atomic_write_text_file` passthrough chain; `serve_authenticated_miner`'s unused `_connection` (streams keep the connection alive); the `generated` bool threaded through a tuple to gate one redundant log line.
- **Silent-failure nits vs `~/.claude/CLAUDE.md` ("never allow a silent failure; all failures must produce error logging")**: `let _ = fs::remove_file(...)` at `:372`/`:431`/`:437` and `if let Ok(existing) = fs::read_to_string(&fingerprint_path)` at `:332` discard errors with no logging — a stranded 0600 secret temp file or an EACCES on the pin file passes unnoticed.

Verdict: **REQUEST_CHANGES** — blocking on the deterministic-temp-name crash loop (can permanently down a containerized validator, inside the crash-safety code itself) and the pre-auth permit held across the un-timed handshake (cheap sustained miner lockout that stalls block production). Both are small, localized fixes.

n13 (approved):
**Reviewer model:** K3 Max

## Review: miner QUIC auth + TLS pinning

Solid, well-scoped hardening of the external-miner channel. The design is sound and the details are handled carefully. No blocking issues.

### What I verified

**Protocol (`miner-api/src/lib.rs`)**
- `Ready { token }` is a clean breaking change; the manual `Debug` impl redacts the token (lib.rs:54-62), so it can't leak via logs.
- `MAX_AUTH_TOKEN_LEN` (512) plus the serialized-frame validation at node startup (miner_server.rs:247-274) correctly handle the JSON-escaping inflation case — a 512-byte token of quotes would otherwise blow the 1 KB `MAX_MESSAGE_SIZE` and fail every miner with an opaque framing error. Tested.

**Auth & TLS (`node/src/miner_server.rs`)**
- Token compare is constant-time with trim-on-wire semantics matching the trimmed on-disk value (miner_server.rs:277-288).
- Secrets are written atomically (unique temp suffix avoids the stale-temp-file deadlock under Docker PID-1 restarts), created 0600, and over-permissive restored files are repaired on load.
- Fingerprint is only (re)published after the cert/key pair builds into the actual serving config — a corrupt cert can't clobber an already-distributed pin (tested: `corrupt_cert_does_not_overwrite_fingerprint`).
- Versioned ALPN `quantus-miner/2` (miner_server.rs:62) makes old 0.2.x miners fail cleanly at the TLS handshake instead of an ambiguous auth error.
- Pre-auth DoS bounds (semaphore of 32 + `max_incoming` + 10s handshake/auth timeout) and the non-blocking result forwarder with flood disconnect (miner_server.rs:747) are well reasoned; the comments about parked handlers match the actual `tokio::select!` behavior.

**Service (`node/src/service.rs`)**
- The version-based supersede checks fix a real race: `version()` is read before `metadata()` (service.rs:164), re-checked after a blocking recv (service.rs:117), and re-checked after a failed submit (service.rs:238) so a rebuilt template interrupts instead of blaming the miner's seal. I confirmed `MiningHandle::version()` semantics in `client/consensus/qpow/src/worker.rs:142-148` — unchanged version guarantees unchanged metadata, and `submit()` re-verifies under its own lock, so a stale seal can never be imported.
- Miner-server startup failure now aborts the essential mining task instead of silently falling back to local mining — correct fail-closed behavior.

**CLI (`node/src/command.rs:671-684`)** — both flag-misuse combinations are rejected at startup. Good.

**Tests** — ran locally: all 9 new `miner_server` tests pass; `cargo clippy -p quantus-miner-api -p quantus-node` exits clean (the 2 node warnings are pre-existing in `cli.rs:89` / `command.rs:56`, untouched by this PR).

### Non-blocking notes

1. **PR body vs. actual change on miner-stack.** The body says "update … miner-stack Compose to share token + fingerprint files with the miner", but the PR actually *deletes* the entire `miner-stack/` directory (20 files, ~4000 lines) and its CI workflow. The deletion is internally consistent (no dangling references in `MINING.md`, `docs/`, or `scripts/`), but please update the description to say the stack was removed and where it lives now — reviewers shouldn't have to discover a 4000-line deletion from the commit log.

2. **Behavior change worth a release note:** `--miner-listen-port` on a non-`--validator` node changed from silently ignored to a startup error (command.rs:671). That's the right fail-fast, but any operator running a full node with the flag set will be surprised at upgrade time.

3. **Lockstep upgrade required:** this is a hard protocol break — miners must run the companion quantus-miner build (new ALPN + token + pin) before the node upgrade, or they drop at the handshake. The PR body mentions the companion PR; just make sure release comms say the same.

4. Cert expiry checked and dismissed: rcgen 0.13's `generate_simple_self_signed` defaults to 1975→4096 validity, so persisting the cert introduces no expiry cliff.

Verdict: **APPROVE** — no blocking findings; notes 1–3 are process/docs, not code.

PR #663, opened 2026-08-12
zk-tree perf

illuzen:
# Batch ZK-tree root computation once per block

## Summary

Reworks how the ZK tree handles insertions: transfers now only **append** their leaf, and the Merkle root is recomputed **once per block** in the zk-tree pallet's `on_finalize`, folding all of the block's leaves into the tree in a single bottom-up pass.

Previously every transfer ran a full leaf-to-root path update: `depth` Poseidon hashes, `3·depth` sibling reads, and `depth` node writes — repeated per insert even though all inserts in a block share most of their paths. The batched pass hashes each dirty node exactly once, so a block with `n` transfers costs roughly `n/4 + n/16 + … + depth` node hashes instead of `n·depth`.

## Why this is safe

Nothing on-chain needs the root mid-block:

- Wormhole exit proofs verify against **historical block headers** (bound by block hash), never against the live `Root`.
- Callers of `record_transfer` only consume the leaf index, which is still assigned at insert time.
- The only consumer of the fresh root is the header stamp in `on_finalize` — exactly where the computation now lives.

During block execution `Root` is simply the root as of the end of the previous block. A new `UnprocessedLeaves` counter tracks the block's pending range; it defaults to 0, so **no storage migration is needed**.

Hook ordering is safe: `on_finalize` runs in pallet declaration order, and `ZkTree` (index 21) is declared after every leaf inserter — including `MiningRewards` (index 6), which inserts the block reward leaf from its own `on_finalize`.

## Weight repricing

Insert pricing moves from "flat full path at the circuit depth ceiling" to a **marginal + per-block base** split, keeping `price × n` sound for any multi-insert call:

| Constant | Before | After |
|---|---|---|
| `INSERT_LEAF_DB_OPS` (reads, writes) | (52, 21) | (3, 4) |
| `INSERT_LEAF_POSEIDON_EVALS` | 18 | 3 |
| `FINALIZE_BASE_DB_OPS` (new, once per block) | — | (56, 22) |
| `FINALIZE_BASE_POSEIDON_EVALS` (new, once per block) | — | 19 |

The depth-dependent tail (boundary siblings, path above the batch, growth bookkeeping) is reserved unconditionally by the zk-tree pallet's `on_initialize`. Per-transfer PoV drops from ~135 KB (52 tree keys) to ~10 KB, which is the real throughput unlock on top of the earlier Poseidon `ref_time` correction.

Downstream weight formulas (wormhole batch exits, mining rewards, vesting, reversible transfers, the wormhole proof-recorder extension) scale off the shared constants and needed no structural change. Vesting's payout weight now clamps at its benchmarked base, since the marginal append is cheaper than the benchmark-time ops it substitutes.

## Behavioral changes

- `insert_leaf` returns just the leaf index (the root is no longer known at insert time).
- `LeafInserted` event no longer carries `new_root`; the authoritative root for the block is in the header.
- `TreeGrew` is emitted at finalize (once per settlement, with the final depth) instead of per insert.
- `get_merkle_proof` rejects leaves not yet folded into the root; at any block boundary all leaves are folded, so RPC behavior is unchanged.
- Root values are **hash-for-hash identical** to the old per-insert algorithm for any leaf sequence.

## Tests

- New equivalence suite: batched settlement == per-insert settlement == independent in-memory reference tree, across sizes spanning the depth-1/2 capacity boundaries and multi-level growth inside one batch.
- Block-grouping invariance: the same 21-leaf sequence split into two blocks at every possible point produces the same root, with all leaves provable.
- Growth-boundary cases: settlement starting exactly at `4^depth` (relies on parking the old root) and spanning the boundary (overwrites the parked value).
- Existing zk-tree, vesting, mining-rewards, and reversible-transfers suites updated/passing (30/27/51/58).

## Notes for reviewers

- The batch algorithm lives in `tree::update_range`; `grow_tree` now only parks the old root at `(old_depth, 0)` — everything above is recomputed by the same pass.
- Benchmarks (`weights_generated`) were produced against the old per-insert path; regenerating them will lower the vesting/reversible bases further. Until then the clamp keeps them conservative.

### DIFF
--- a/docs/zk-trie-architecture.md
+++ b/docs/zk-trie-architecture.md
@@ -190,9 +190,8 @@ Every balance transfer on the chain is automatically captured and recorded into 
    │ ZkTrie::insert_leaf()                         │
    │                                               │
    │  1. Store leaf data at next index             │
-   │  2. Compute leaf_hash = poseidon(leaf_data)   │
-   │  3. Mark the leaf as pending                  │
-   │  4. Emit LeafInserted event                   │
+   │  2. Mark the leaf as pending                  │
+   │  3. Emit LeafInserted event                   │
    └───────────────────────┬───────────────────────┘
                            │  (once per block)
                            ▼
--- a/docs/zk-trie-architecture.md
+++ b/docs/zk-trie-architecture.md
@@ -191,11 +191,29 @@ Every balance transfer on the chain is automatically captured and recorded into 
    │                                               │
    │  1. Store leaf data at next index             │
    │  2. Compute leaf_hash = poseidon(leaf_data)   │
-   │  3. Update path from leaf to root             │
-   │  4. Store new root                            │
-   │  5. Emit LeafInserted event                   │
+   │  3. Mark the leaf as pending                  │
+   │  4. Emit LeafInserted event                   │
+   └───────────────────────┬───────────────────────┘
+                           │  (once per block)
+                           ▼
+   ┌──────────────────────────────────────────────┐
+   │ ZkTrie on_finalize                            │
+   │                                               │
+   │  1. Fold ALL leaves appended this block into  │
+   │     the tree in one bottom-up pass (shared    │
+   │     parents hashed once, not once per leaf)   │
+   │  2. Grow the tree if capacity was crossed     │
+   │  3. Store new root                            │
+   │  4. Publish root to the block header          │
    └──────────────────────────────────────────────┘
 ```
+
+Root recomputation is deliberately **batched per block**: a per-insert path update
+costs `depth` Poseidon hashes and `3·depth` sibling reads for every transfer, while
+the batched pass costs roughly `n/4 + n/16 + … + depth` hashes for a block with `n`
+transfers. Nothing on-chain needs the root mid-block — exit proofs verify against
+historical block headers — so during block execution `Root` is simply the root as of
+the end of the previous block.
 
 ### 2. Generating a ZK Proof
 
--- a/pallets/mining-rewards/src/weights.rs
+++ b/pallets/mining-rewards/src/weights.rs
@@ -79,8 +79,10 @@ const BASE_KEY_POV: u64 = 2700;
 /// Weights for `pallet_mining_rewards` using the Substrate node and recommended hardware.
 pub struct SubstrateWeight<T>(PhantomData<T>);
 impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
-	/// Reserved in `on_initialize` before extrinsics run. Leaf inserts are flat-
-	/// priced at [`pallet_zk_tree::CIRCUIT_MAX_TREE_DEPTH`].
+	/// Reserved in `on_initialize` before extrinsics run. Leaf appends are priced
+	/// at the flat marginal per-insert share ([`pallet_zk_tree::INSERT_LEAF_DB_OPS`]);
+	/// the batched root recomputation's depth-dependent tail is reserved by the
+	/// zk-tree pallet's own `on_initialize`.
 	fn on_finalize_rewarded_miner() -> Weight {
 		// Minimum execution time: 150_000_000 picoseconds.
 		let (tree_reads, tree_writes) = pallet_zk_tree::INSERT_LEAF_DB_OPS;
@@ -103,7 +105,7 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 
 // For backwards compatibility and tests.
 impl WeightInfo for () {
-	/// Same flat circuit-depth pricing as `SubstrateWeight`, with `RocksDbWeight`
+	/// Same marginal per-insert pricing as `SubstrateWeight`, with `RocksDbWeight`
 	/// in place of the runtime's configured `DbWeight`.
 	fn on_finalize_rewarded_miner() -> Weight {
 		// Minimum execution time: 150_000_000 picoseconds.
--- a/pallets/reversible-transfers/src/weights.rs
+++ b/pallets/reversible-transfers/src/weights.rs
@@ -64,10 +64,11 @@ const EXECUTE_TRANSFER_BASE_READS: u64 = 5;
 const EXECUTE_TRANSFER_BASE_WRITES: u64 = 5;
 
 /// `execute_transfer`'s weight: the benchmarked base (compute + non-tree storage)
-/// plus the depth-dependent ZK-tree leaf insert performed by the wormhole proof
-/// recorder. `insert_leaf` walks the tree leaf-to-root, so DB ops and PoV scale
-/// with `tree_ops` via [`pallet_zk_tree::TREE_KEY_POV`], and the path update also
-/// computes one Poseidon hash per level (`tree_hash_time`).
+/// plus the ZK-tree leaf append performed by the wormhole proof recorder, priced at
+/// the flat marginal per-insert share of the block's batched root recomputation
+/// (`tree_ops` DB ops, PoV via [`pallet_zk_tree::TREE_KEY_POV`], Poseidon compute
+/// via `tree_hash_time`); the depth-dependent tail is reserved per block by the
+/// zk-tree pallet's `on_initialize`.
 fn execute_transfer_weight(
 	db: RuntimeDbWeight,
 	(tree_reads, tree_writes): (u64, u64),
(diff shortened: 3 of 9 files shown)


n13 (commented):
**Reviewer model:** Opus 5

I traced the batch algorithm against the storage layout, re-derived the weight model by hand, checked the hook-ordering claim in the generated runtime, and ran the suites locally: zk-tree 30/30, vesting 27/27, mining-rewards 51/51, reversible-transfers 58/58, `cargo check --workspace --all-targets` clean. **I could not find a correctness defect.** Everything below is non-blocking.

## What I verified independently

**`update_range`'s boundary handling is sound.** Because the appended leaves are contiguous, at every level only `parent_lo` can own children left of `lo`, and — in an append-only tree — nothing right of `hi` has ever been written, so the "skip the storage read" shortcut at `pallets/zk-tree/src/tree.rs:242` is safe. Multi-level growth works out too: the previous settlement's invariant `capacity_at_depth(Depth) >= LeafCount` gives `start <= 4^old_depth`, so for every new level `k >= 1` the node `(old_depth+k, 0)` falls inside the dirty range and is recomputed, while `(old_depth, 0)` is either read back as the parked root (when `start == 4^old_depth`) or overwritten by the same pass.

**The marginal/base weight split closes exactly.**
- Per leaf: 2 insert-phase reads (`LeafCount`, `UnprocessedLeaves::mutate`) + 1 finalize `Leaves` read = 3 ✓; 3 insert-phase writes + 1 amortized node write = 4 ✓.
- Internal-node writes over a batch are `Σ_{l=1..depth-1} (⌊(end-1)/4^l⌋ - ⌊start/4^l⌋ + 1) <= (n-1)/3 + depth - 1`. The per-leaf `n` covers the `(n-1)/3` term; `FINALIZE_BASE_DB_OPS.1 = d+6` covers `depth-1` plus `UnprocessedLeaves`/`Depth`/parked node/`Root`/header root — leaving exactly 2 spare, which is what the `TreeGrew` deposit's `Events`+`EventCount` writes need.
- Poseidon: actual `2n` leaf hashes + `<= (n-1)/3 + depth` node hashes + `<= 3` boundary leaf hashes, vs charged `3n + (d+3)` ✓.
- Reads: `3·depth + 5 <= 3d + 8` ✓.

**Hook ordering holds.** `construct_runtime` emits `AllPalletsWithSystem` in declaration order (`frame/support-procedural/src/construct_runtime/mod.rs:621`) and the `OnFinalize` tuple impl runs in that order, so `ZkTree` (21) folds after `MiningRewards` (6) — the only hook-based inserter (`pallets/mining-rewards/src/lib.rs:232`). `Vesting` (22) is declared after `ZkTree` but has no hooks. And the design degrades gracefully if that ever changes: a leaf appended after `ZkTree::on_finalize` just settles at the next fold, with `Root == first (LeafCount - UnprocessedLeaves) leaves` still true throughout.

**No stale-`Root` consumer.** Exit bundles bind to `frame_system::block_hash(n)` (`pallets/wormhole/src/lib.rs:1104`), never the live `Root`; the only in-runtime readers are the header stamp and `ZkTreeApi`, which always executes against a post-`on_finalize` state.

**No upgrade discontinuity.** `UnprocessedLeaves` defaults to 0 and the pass never recomputes already-settled subtrees, so existing `Nodes`/`Root` carry over verbatim. (The old `update_path`'s `depth == 0` early return that yielded a bare leaf hash as root was unreachable — `capacity_at_depth(0) == 0` always forced the growth branch first — so no legacy state can disagree with the new positional reference.)

## Findings (all non-blocking)

**1. `pallets/vesting/src/weights.rs:153` — the under-charge guard is now tautological.**
Adding `.max(base)` *inside* `payout_weight` (line 56) makes `payout_weight_never_undercharges_the_benchmarked_base` unfalsifiable: `Weight::max` is componentwise, so `augmented.all_gte(base)` is true by construction for any inputs. The doc comment still claims the test "pins that the clamp keeps the augmented weight from ever falling below it," but it would now pass even if `INSERT_LEAF_DB_OPS` and `TREE_KEY_POV` were both zeroed. Either assert on the pre-clamp value, or invert it into something that still bites — e.g. assert the clamp *is* currently active, so that when the benchmarks are regenerated and the substitution flips direction, someone is told.

**2. Nothing exercises the new cross-pallet `on_finalize` ordering.**
The PR's central new assumption — every leaf inserter's `on_finalize` runs before `ZkTree`'s — is recorded only in a comment (`pallets/zk-tree/src/lib.rs:290-293`). `pallets/mining-rewards/src/mock.rs:129` wires a `MockProofRecorder`, no `runtime/tests/*` file references `ZkTree`, and the new zk-tree tests call `process_pending_leaves()` directly rather than driving a block. A runtime-level test that runs a full block including a mining reward and asserts `UnprocessedLeaves == 0` plus a header root covering the reward leaf would pin an invariant that today rests on pallet-index luck.

**3. `pallets/zk-tree/src/lib.rs:327` — every leaf is Poseidon-hashed twice.**
`insert_leaf` computes `hash_leaf` solely to populate `LeafInserted`, then `update_range` re-reads `Leaves` and recomputes the identical hash (`tree.rs:228`). That is 2 of the 3 charged evals per leaf, in a PR whose whole point is cutting Poseidon work. Since the event's shape is already changing, dropping `leaf_hash` from it (the hash is derivable from `Leaves` via RPC) would cut per-leaf hashing by a third.

**4. `pallets/zk-tree/src/lib.rs:75-79` — the 5× `POSEIDON_EVAL_REF_TIME_PS` cut rests on a measurement that never runs.**
The cited `measure_hash_node_time` is `#[ignore]`d (`tree.rs:385`) and native, while the runtime executes wasm32, where Goldilocks' `u64×u64→u128` products are emulated rather than a single `mulq`. The repo's own generated benchmarks do bound it in the right direction (vesting `claim` 82µs vs `create_schedule` 39µs — a 43µs delta covering an entire vested transfer *plus* the insert's ~2 evals), so 10µs looks defensible; but nothing in CI pins it, and the real margin over the wasm-measured cost is thinner than the comment's "~3×" suggests. A `#[test]` upper-bound assertion, or a note that 3.3µs is a native lower bound, would make the reasoning auditable.

**5. `pallets/zk-tree/src/lib.rs:302-304` (and six `where AccountIdOf<T>: AsRef<[u8]>` clauses in `tree.rs`) are now redundant.**
`Config` requires `frame_system::Config<AccountId: AsRef<[u8]>>` (line 212) — which is precisely why that bound was added — so the per-impl `where` clauses duplicate what the supertrait already guarantees.

**6. `runtime/src/transaction_extensions.rs:164-167` — stale justification.**
"(`TreeGrew` is deliberately not modeled: … the flat depth-ceiling insert pricing already carries margin for young trees.)" The pricing is no longer flat-at-depth-ceiling, and `TreeGrew` now fires from `on_finalize`, so it can never appear in the extension's post-dispatch event scan at all. The conclusion still holds; the reason given for it no longer does.

**7. `pallets/zk-tree/src/lib.rs:390-391` — a pending leaf is reported as `LeafIndexOutOfBounds`,**
conflating "no such leaf" with "appended this block, provable next block." Not reachable through the RPC (which always resolves to a block hash where `UnprocessedLeaves == 0`), so this is cosmetic — but a distinct error variant would read better for any future in-runtime caller.

---

**Verdict: comment** — the algorithm, the weight model, and the ordering assumption all hold up under scrutiny; the items above are quality and coverage suggestions, none of them blocking.

n13 (approved):
**Reviewer model:** GPT Sol

APPROVE — I found no blocking correctness, weight-soundness, upgrade, or runtime-integration defect at `4278880d962be64ecdfef71f10fafec950c125bc`.

I traced the append-only settlement path through `insert_leaf`, `process_pending_leaves`, `update_range`, tree growth, proof generation, the real runtime hook tuple, and every leaf-producing call path. The contiguous dirty-range algorithm preserves existing siblings, overwrites or reuses the parked old root correctly across capacity boundaries, and converges to the same root regardless of block grouping. `UnprocessedLeaves` defaults to zero, so existing settled state remains valid across upgrade. The marginal/base split also covers the per-leaf reads, writes, and hashes plus the depth-dependent boundary and growth tail; a standalone wasm build of the same 16-felt Poseidon operation measured roughly 6.0–6.1 µs/hash on this host, below the new 10 µs charge.

Validation:

- `git diff --check 803094f7d9f836bd6b489c74acfa26f652d78a78...4278880d962be64ecdfef71f10fafec950c125bc` — clean.
- `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-zk-tree -p pallet-vesting -p pallet-mining-rewards -p pallet-reversible-transfers` — zk-tree 30 passed, vesting 58 passed, mining-rewards 27 passed, reversible-transfers 51 passed; no failures.
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime transaction_extensions::tests::full_block_finalize_settles_all_leaves_including_mining_reward -- --exact --nocapture` — passed, pinning mining-reward/ZkTree finalize ordering and header-root publication.
- GitHub format and Clippy/doc checks pass; the two full check/test matrix jobs were still running when this review was posted.

Non-blocking cleanup: `pallets/vesting/src/weights.rs:153` leaves the old `payout_bases()` helper unused (and produces a dead-code warning), while line 27 still links to the removed `payout_weight_never_undercharges_the_benchmarked_base` test name. Removing the helper and updating the link would keep the follow-up commit tidy.

No blocking findings.

PR #664, opened 2026-08-12
ci: Miner API version bump to quantus-miner-api-v0.3.0

illuzen:
## Miner API Release Proposal

This PR proposes releasing **quantus-miner-api** version `0.3.0`.

### Changes
- Updated `miner-api/Cargo.toml` version to `0.3.0`
- Updated `Cargo.toml` workspace dependency version to `0.3.0`
- Updated `Cargo.lock`

### DIFF
--- a/.github/workflows/miner-api-publish.yml
+++ b/.github/workflows/miner-api-publish.yml
@@ -9,63 +9,108 @@ on:
     types: [closed]
     branches:
       - main
+  # Recovery path: tag + publish what's already on a ref without a version-bump PR
+  # (e.g. main carries 0.3.0 but create-miner-api-tag never ran after a failed check).
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to publish (quantus-miner-api-vX.Y.Z or X.Y.Z)'
+        required: true
+        type: string
+      ref:
+        description: 'Git ref containing that version (branch, tag, or SHA)'
+        required: false
+        type: string
+        default: 'main'
+      is_draft:
+        description: 'Draft release (tag only; skip crates.io)?'
+        required: true
+        type: boolean
+        default: false
 
 jobs:
   prepare-miner-api-release:
     name: Prepare Miner API Release
-    if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'miner-api-release-proposal')
+    if: |
+      (github.event_name == 'pull_request' &&
+       github.event.pull_request.merged == true &&
+       contains(github.event.pull_request.labels.*.name, 'miner-api-release-proposal')) ||
+      github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     outputs:
-      version: ${{ steps.extract_version.outputs.version }}
-      is_draft: ${{ steps.extract_version.outputs.is_draft }}
-      commit_sha: ${{ github.event.pull_request.merge_commit_sha }}
+      version: ${{ steps.resolve.outputs.version }}
+      is_draft: ${{ steps.resolve.outputs.is_draft }}
+      commit_sha: ${{ steps.resolve.outputs.commit_sha }}
     steps:
-      - name: Checkout merge commit
+      - name: Checkout release commit
         uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.pull_request.merge_commit_sha }}
+          ref: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.ref || github.event.pull_request.merge_commit_sha }}
 
-      - name: Extract version from PR title
-        id: extract_version
+      - name: Resolve version and commit
+        id: resolve
         run: |
           set -euo pipefail
-          # Extract version from PR title (format: "ci: Miner API version bump to quantus-miner-api-vX.Y.Z").
-          # Take only the first match so a title with two version strings cannot
-          # write a multi-line value into GITHUB_OUTPUT.
-          VERSION=$(echo "${{ github.event.pull_request.title }}" \
-            | grep -o 'quantus-miner-api-v[0-9]\+\.[0-9]\+\.[0-9]\+' \
-            | head -n1 || true)
-          if [ -z "$VERSION" ]; then
-            echo "Error: Could not extract miner-api version from PR title: ${{ github.event.pull_request.title }}"
-            exit 1
+
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            RAW="${{ github.event.inputs.version }}"
+            case "$RAW" in
+              quantus-miner-api-v*) VERSION="$RAW" ;;
+              *) VERSION="quantus-miner-api-v$RAW" ;;
+            esac
+            if [[ ! "$VERSION" =~ ^quantus-miner-api-v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+              echo "Error: invalid version '$RAW' (want X.Y.Z or quantus-miner-api-vX.Y.Z)"
+              exit 1
+            fi
+            if [ "${{ github.event.inputs.is_draft }}" = "true" ]; then
+              echo "is_draft=true" >> "$GITHUB_OUTPUT"
+            else
+              echo "is_draft=false" >> "$GITHUB_OUTPUT"
+            fi
+          else
+            # Take only the first match so a title with two version strings cannot
+            # write a multi-line value into GITHUB_OUTPUT.
+            VERSION=$(echo "${{ github.event.pull_request.title }}" \
+              | grep -o 'quantus-miner-api-v[0-9]\+\.[0-9]\+\.[0-9]\+' \
+              | head -n1 || true)
+            if [ -z "$VERSION" ]; then
+              echo "Error: Could not extract miner-api version from PR title: ${{ github.event.pull_request.title }}"
+              exit 1
+            fi
+            if [[ "${{ contains(github.event.pull_request.labels.*.name, 'draft-release') }}" == "true" ]]; then
+              echo "is_draft=true" >> "$GITHUB_OUTPUT"
+            else
+              echo "is_draft=false" >> "$GITHUB_OUTPUT"
+            fi
           fi
 
-          # The tag/crates.io name comes from the title; cargo publish uses
-          # miner-api/Cargo.toml. Refuse to proceed if they disagree.
           CARGO_VERSION=${VERSION#quantus-miner-api-v}
+
+          # Tag/crates.io name must match what cargo publish will upload.
           if ! grep -q "^version = \"$CARGO_VERSION\"" miner-api/Cargo.toml; then
-            echo "Error: PR title version $VERSION does not match miner-api/Cargo.toml:"
+            echo "Error: release version $VERSION does not match miner-api/Cargo.toml:"
             grep '^version = ' miner-api/Cargo.toml || true
             exit 1
           fi
 
-          echo "version=$VERSION" >> $GITHUB_OUTPUT
-
-          # Check if this is a draft release
-          if [[ "${{ contains(github.event.pull_request.labels.*.name, 'draft-release') }}" == "true" ]]; then
-            echo "is_draft=true" >> $GITHUB_OUTPUT
-          else
-            echo "is_draft=false" >> $GITHUB_OUTPUT
+          # README ships in the .crate tarball; keep the install snippet honest.
+          if ! grep -q "^quantus-miner-api = \"$CARGO_VERSION\"" miner-api/README.md; then
+            echo "Error: miner-api/README.md install line does not match $CARGO_VERSION:"
+            grep -E '^quantus-miner-api = "' miner-api/README.md || true
+            exit 1
           fi
-          echo "Extracted miner-api version: $VERSION (matches Cargo.toml)"
-          echo "Merge commit: ${{ github.event.pull_request.merge_commit_sha }}"
+
+          COMMIT_SHA=$(git rev-parse HEAD)
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "commit_sha=$COMMIT_SHA" >> "$GITHUB_OUTPUT"
+          echo "Resolved miner-api version: $VERSION at $COMMIT_SHA"
 
   format-checks:
     name: 🏁 Format Checks (Miner API)
     needs: prepare-miner-api-release
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout merge commit
+      - name: Checkout release commit
         uses: actions/checkout@v4
         with:
           ref: ${{ needs.prepare-miner-api-release.outputs.commit_sha }}
@@ -86,7 +131,7 @@ jobs:
     needs: [prepare-miner-api-release, format-checks]
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout merge commit
+      - name: Checkout release commit
         uses: actions/checkout@v4
         with:
           ref: ${{ needs.prepare-miner-api-release.outputs.commit_sha }}
@@ -113,7 +158,7 @@ jobs:
     permissions:
       contents: write
     steps:
-      - name: Checkout merge commit
+      - name: Checkout release commit
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
@@ -142,7 +187,7 @@ jobs:
           fi
 
           git tag -a "$VERSION" -m "Miner API release $VERSION" "$SHA"
-          git push origin "$VERSION"
+          git push origin "refs/tags/$VERSION"
 
   publish-miner-api-to-crates:
     name: 📦 Publish to crates.io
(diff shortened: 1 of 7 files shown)


n13 (approved):
**Reviewer model:** Opus 5

## Verdict: approve — the bump itself is correct and complete

The three-line diff is mechanically right, and `0.3.0` is the semver-correct choice.

**What I verified**

- All three version strings agree with each other and with the branch/title: `miner-api/Cargo.toml:15`, root `Cargo.toml:236`, and the `quantus-miner-api` entry in `Cargo.lock` — all `0.3.0`, matching `release/quantus-miner-api-v0.3.0` and the PR title.
- **The minor bump is correct, not conservative.** Since `quantus-miner-api-v0.2.1`, `miner-api/src/lib.rs` gained a genuinely breaking change from #662: `MinerMessage::Ready` went from a unit variant to `Ready { token: String }`, which changes the JSON wire form from `"Ready"` to `{"Ready":{"token":"…"}}`. Under 0.x semver `0.2.x → 0.3.0` is the right signal; a patch bump would have silently broken external miners on `cargo update`. The ALPN was bumped in lockstep (`MINER_ALPN = b"quantus-miner/2"`, `node/src/miner_server.rs:62`), so a stale 0.2.1 miner fails cleanly at the TLS handshake rather than with a confusing auth error. Nicely done.
- `0.3.0` is free on crates.io (`max_version` is `0.2.1`) and the tag `quantus-miner-api-v0.3.0` does not exist yet, so neither publish step collides.
- `cargo package --locked --package quantus-miner-api` succeeds standalone (6 files, 16.4 KiB); all three workspace-inherited deps resolve to publishable version reqs, and the crate metadata (description/license/repository/keywords/categories) is complete.
- `Cargo.lock` is internally consistent — the only other reference is `quantus-node`'s bare-name dep entry, no version pin. `cargo metadata --locked` passes.
- The publish workflow's gates all match this PR: base `main`, `miner-api-release-proposal` label present, no `draft-release` label, and the title regex at `.github/workflows/miner-api-publish.yml:33` yields exactly one match.
- File set matches the prior release PRs (#657, #583) exactly.

Nothing below blocks the merge. The first item is the only one I'd consider folding in before merging, since crates.io versions are immutable.

---

### 1. `miner-api/README.md:30` — stale install version ships inside the 0.3.0 tarball

```toml
# quantus-miner-api = "0.1.0"
```

This README is one of the 6 files in the `.crate` tarball, so it is what renders on crates.io and docs.rs for 0.3.0. `0.1.0` predates *two* wire breaks: the `Ready { token }` change above, and the `distance_threshold` → `difficulty` rename in 0.2.0. A third-party miner author who copies that line builds against a protocol no 0.3.0 node speaks. Neither release workflow touches the README (`miner-api-release-proposal.yml:181` stages only the two `Cargo.toml`s and `Cargo.lock`), so it will keep drifting every release unless it is updated by hand or added to the sed step.

Relatedly, `miner-api/README.md:28` gives external miners a path dependency (`{ path = "../miner-api" }`) as the primary form, with the registry form commented out — but external miners are the reason this crate is published, and they do not have `../miner-api` on disk.

### 2. `MINING.md:635` — protocol spec documents a field name that no longer exists

```
| `distance_threshold` | String | Difficulty (U512 as decimal string) |
```

The actual field has been `pub difficulty: String` since 0.2.0 (`miner-api/src/lib.rs:111`), with no serde rename. Anyone implementing an external miner from `MINING.md` rather than from the crate will send `distance_threshold` and fail every `NewJob` with `missing field difficulty`. Pre-existing, but this is the PR that ships the contract the doc describes, and every other `MinerMessage` reference in `MINING.md` was updated for auth in #662 — this row was missed.

### 3. `node/src/miner_server.rs:62` — the ALPN is part of the wire contract but lives outside the contract crate

`MINER_ALPN = b"quantus-miner/2"` is explicitly versioned with the wire protocol (`MINING.md:796`: "versioned with the wire…"), yet it is defined in the node crate. External miners depending only on the published `quantus-miner-api` must hardcode the string and have nothing to recompile against when it changes. Since 0.3.0 is precisely the release where it changed, re-exporting it from `miner-api` alongside `MinerMessage` would put the whole handshake contract in one place. Not for this PR — but worth a follow-up.

### 4. `.github/workflows/miner-api-publish.yml:52` — the tag is pushed before anything is validated

`create-miner-api-tag` runs `git tag -a` / `git push origin` at lines 52-53, and both `format-checks` (line 57) and `build-and-test-miner-api` (line 78) declare `needs: create-miner-api-tag`. So any fmt drift, `clippy -D warnings` failure, or flaky test aborts the release *after* the tag is already on the remote — leaving `quantus-miner-api-v0.3.0` pointing at nothing published. Re-running then dies immediately at `git tag -a` ("tag already exists"), so recovery means deleting a pushed tag by hand. Moving the tag push to after `build-and-test-miner-api` would make the job idempotent.

### 5. `.github/workflows/miner-api-publish.yml:33` — the released version comes only from the PR title

`VERSION=$(echo "…title…" | grep -o 'quantus-miner-api-v[0-9]\+\.[0-9]\+\.[0-9]\+')` is never cross-checked against `miner-api/Cargo.toml`'s actual `version`. They agree here (both 0.3.0, verified), but if someone edits a title before merge or hand-applies the `miner-api-release-proposal` label to an unrelated PR, a tag is pushed claiming a version the tree does not contain, and `cargo publish` uploads a different one — permanently desyncing git history from crates.io. A one-line `grep -q "^version = \"${VERSION#quantus-miner-api-v}\"" miner-api/Cargo.toml` before tagging would close it. (Minor: `grep -o` emits one line per match, so a title with two version strings writes a multi-line value into `$GITHUB_OUTPUT`.)

### 6. Smaller workflow nits

- `miner-api-publish.yml:121` — `cargo publish --token … --allow-dirty` omits `--locked` even though lines 90/93/96/99 all use it, and `--allow-dirty` disables the clean-tree safety net on what is a fresh tag checkout anyway. The token is also passed both as an argv flag and via `env:`; the `env:` form alone is enough.
- `miner-api-publish.yml:69` — `cargo install taplo-cli --locked` is unpinned, while `ci.yml:39` pins taplo `0.9.3`. The PR-time gate and the release-time gate can therefore run different formatters over the same `.toml` files this PR edits — and per item 4, a taplo-default change would fail *after* the tag is pushed.
- There is no CHANGELOG anywhere in the repo, so for a release that breaks the wire protocol the GitHub release notes are the only migration signal external miner authors get. Worth saying "0.3.0 requires a node at #662 or later" explicitly.

---

**Verdict: approved.** The version bump is correct, complete, and publishable as-is; every item above is a pre-existing docs/automation gap this release surfaces rather than a defect introduced by `a87ab754`. Only the `README.md:30` version string is worth fixing inside this PR, since it is baked into the immutable 0.3.0 tarball.

n13 (changes requested):
**Reviewer model:** Opus 5

The workflow hardening here is going in the right direction, and several parts check out cleanly (see "Verified good" below). But moving tag creation behind CI introduces a new reachable state — *merged with a bumped manifest, no tag* — and the release-proposal workflow is not prepared for it. That combination can silently publish 0.3.0's code to crates.io labelled `0.2.2`.

### Blocking

**1. Version downgrade: tagging is now CI-gated, but the next version is still computed from tags.** `.github/workflows/miner-api-publish.yml:111` / `.github/workflows/miner-api-release-proposal.yml:58`

`create-miner-api-tag` now has `needs: [prepare-miner-api-release, build-and-test-miner-api]` and no `if:`, so a post-merge failure skips it. `main` then carries `miner-api/Cargo.toml` at `0.3.0` with the newest tag still `quantus-miner-api-v0.2.1`. The proposal workflow's only version input is line 58 — `git tag -l ... | sort -V | tail -n 1` — never the manifest. The next `patch` run computes `0.2.2` and the unconditional sed at line 156 rewrites the manifest *downward*. I ran that sed against the real file: `version = "0.3.0"` → `version = "0.2.2"`, and the workflow's own verify grep passes.

Every guard fails to catch it: the existing-tag check at `:145` passes (v0.2.2 is free), and this PR's new cross-check at `miner-api-publish.yml:46` passes *tautologically* — title and manifest are both the same wrong number derived from the same stale tag. `cargo publish` does no ordering check; crates.io accepts any version below `max_version` that doesn't already exist. Net result: 0.3.0's breaking `Ready { token }` / `difficulty` changes ship as a patch release, and 0.3.0 is never published.

This state was unreachable before the PR — the old `create-miner-api-tag` was the first job with no `needs:`, so a merge always produced a tag. Worth noting `main` has **no required status checks** (only 1 approving review), so a release PR can merge without CI ever having gone green.

Fix: derive the base version from `miner-api/Cargo.toml` (`cargo metadata` / `cargo pkgid`) rather than from tags, or at minimum refuse a computed version that is not strictly greater than the manifest's.

**2. No recovery path out of the state #1 creates.** `.github/workflows/miner-api-release-proposal.yml:190`

The correct recovery is "tag and publish what's already on main", and it's unreachable. Re-proposing the same version makes all four seds no-ops, so `git add` stages nothing and `git commit` at line 190 exits 1 (`nothing to commit`) under `set -ex`. And `miner-api-publish.yml` has only a `pull_request: closed` trigger — no `workflow_dispatch`, no `push: tags` — so a maintainer cannot start a publish by hand, and pushing the tag manually fires nothing. That leaves the downgrade in #1 as the path of least resistance. Please add a manual trigger, or detect and report the untagged-merge state explicitly.

**3. The publish job is the one job left pinned to a mutable ref.** `.github/workflows/miner-api-publish.yml:159`

The PR converted `format-checks` (`:71`), `build-and-test` (`:92`) and `create-miner-api-tag` (`:121`) to `ref: ...outputs.commit_sha`, but publish still uses `...outputs.version` — the tag *name*. `actions/checkout@v4` resolves an unqualified ref by checking `refs/remotes/origin/<ref>` **before** `refs/tags/<ref>` (`ref-helper.ts` `getCheckoutInfo`), and there is no tag-protection ruleset on this repo. So a branch named exactly `quantus-miner-api-v0.3.0` shadows the tag and gets packaged and uploaded under the org's `CARGO_REGISTRY_TOKEN` — content that was never format-checked, built, or version-cross-checked. Tagging raises no ambiguity error, so nothing upstream notices.

`commit_sha` is already in `needs` here. Switching this one token makes the pipeline SHA-pinned end to end, which is exactly the property the rest of the diff establishes.

### Should fix

**4. `cargo update` failure is still swallowed.** `miner-api-release-proposal.yml:175` — `|| echo "cargo update tried, proceeding."` is the one deliberately silenced command in a `set -ex` block where every other edit gets a `grep -q` + `exit 1` (including the README block this PR adds directly below it). A failure here commits a stale `Cargo.lock`; `cargo build --locked` then fails *after* merge, which lands you squarely in #1. This line is pre-existing, but the PR makes its cost much higher.

**5. The new cross-check doesn't cover the file this PR started stamping.** `miner-api-publish.yml:46` validates only `miner-api/Cargo.toml`. `miner-api/README.md` is now version-stamped by the proposal workflow and ships in the `.crate` tarball (confirmed: `cargo package --list` emits `README.md`), yet nothing validates it. A hand-authored release PR — the path this very PR took — publishes a crate whose README tells users to depend on the previous version. One line mirroring `:46` closes it.

**6. Tagging was made idempotent; publishing was not.** `miner-api-publish.yml:167` — the comment at `:132` says re-running after a flaked publish must work, but `cargo publish` hard-errors with `crate version X is already uploaded` if the upload landed and the job didn't (runner eviction, dropped connection on the final response). The tag guard short-circuits correctly, then publish fails permanently and the run can never be made green.

**7. Generated PR body under-reports its own diff.** `miner-api-release-proposal.yml:189` adds `miner-api/README.md` to `git add`, but the `### Changes` list at `:208-211` still names only three files. Add the bullet.

### Nits

- **`:150`** — the three `needs.*.result == 'success'` clauses are dead. With no status function in the expression, GitHub already applies the implicit all-needs-succeeded gate; only the `is_draft` clause does work. This PR added a third no-op clause.
- **`:46`** — `grep -q` uses `$CARGO_VERSION` as a regex (the `.`s are wildcards) and scans the whole manifest rather than the `[package]` table. Unreachable today (the version is digits-and-dots only, and the manifest has one `^version` line), but `grep -qF` is exact for free.
- **`:21`** — `commit_sha` is never checked non-empty before four jobs feed it to `ref:`. `actions/checkout@v4` treats `ref: ''` as falsy and silently falls back to the triggering ref rather than erroring. A null `merge_commit_sha` on a merged `closed` event is rare and the failure is loud (`git tag` can't resolve `""`), but a one-line `[ -n "$SHA" ]` matches the fail-early style of the rest of the block.
- **`:120`** — `fetch-tags: true` is a no-op alongside `fetch-depth: 0`, which already fetches all tags. Related: this job full-clones a Substrate monorepo to push one ref; `git ls-remote --tags` (or `git push origin "$SHA:refs/tags/$VERSION"`, which is idempotent server-side) needs no history.
- **`:35`/`:39`** — the PR title is still interpolated into `run:` rather than passed via `env:`. To be fair: this is pre-existing and the PR *reduced* the blast radius by moving extraction into a job with no `contents: write` (repo default token permission is `read`). The specific escalation via injected `$GITHUB_OUTPUT` writes doesn't work as it might appear — `exit 0` inside `$( )` only kills the subshell, leaving `VERSION` empty and tripping the `exit 1`. Still worth `env: PR_TITLE:` while you're in here; `${{ github.actor }}` at `:129-130` is fine (logins are alphanumeric).
- No `concurrency:` group on this workflow (`ci.yml:18` has one). The race window between the existence check at `:134` and the push at `:145` is now minutes long rather than seconds, because tagging happens after the build.
- `create-miner-api-tag` has no `is_draft` guard while publish does, so a draft release pushes a real tag that permanently burns that version number. Pre-existing, not introduced here — but the PR rewrote this job and the fix is one line.
- `cargo install taplo-cli --locked` at `:77` compiled taplo from source for ~2m of the 5m v0.2.1 run; `ci.yml:38-41` already downloads a prebuilt binary. That cost is now on the blocking path to the tag.
- `.github/workflows/dilithium-publish.yml` is a near-verbatim twin and still tags before any build runs — the exact bug this PR fixes. Worth a follow-up, ideally as a `workflow_call` reusable workflow rather than a third copy.

### Verified good

`taplo format --check` passes; `cargo metadata --locked --offline` exits 0, so the `Cargo.lock` bump is sufficient and `--locked` builds won't break. `0.2.1 → 0.3.0` is the right bump for the breaking `Ready { token }` change under 0.x semver. The `distance_threshold` → `difficulty` rename in MINING.md matches `miner-api/src/lib.rs:90`, with zero occurrences of the old name left repo-wide. The README sed's anchored regex matches exactly one line and fails loudly otherwise, and the claim that the README ships in the tarball is accurate. The tag idempotency guard's happy path is sound — `merge_commit_sha` is stable across re-runs (the payload is frozen) and `fetch-depth: 0` does make the remote tags visible to `git rev-list`. Workspace field inheritance resolves correctly for `cargo publish` from the tag checkout. The job rename breaks no required status check.

**Verdict: request changes** — finding #1 is a regression introduced by this PR that can publish 0.3.0's code to crates.io as 0.2.2, and #2 removes the clean way out of the state that causes it.

n13 (approved):
**Reviewer model:** GPT Sol

## Verdict: approve — the release blockers are resolved at `715226f`

No blocking findings remain.

The two commits since the prior change request close the concrete failure modes:

- `.github/workflows/miner-api-release-proposal.yml:54-159` now compares the manifest with the newest tag, refuses the untagged-higher-manifest state, bases automatic bumps on the higher version, and rejects equal/downward custom or computed versions. The formerly reachable `0.3.0` manifest → `0.2.2` downgrade is no longer possible.
- `.github/workflows/miner-api-publish.yml:14-106` adds a manual recovery entry point and validates both `miner-api/Cargo.toml` and the README install line before releasing.
- `.github/workflows/miner-api-publish.yml:108-190` pins checks and tagging to the resolved commit SHA and creates the tag only after format/build/test succeed. The tag guard is idempotent when the existing tag already resolves to that SHA.
- `.github/workflows/miner-api-publish.yml:192-228` also SHA-pins the crates.io checkout and treats Cargo's duplicate-upload response as success, so a dropped response after a successful upload can be retried.
- The version bump itself remains correct: `0.2.1 → 0.3.0` appropriately signals the breaking `MinerMessage::Ready { token }` wire change since the previous tag, and the manifest, workspace dependency, lockfile, PR title, and packaged README all agree on `0.3.0`. The protocol table now agrees with the actual `MiningRequest.difficulty` field.

Validation performed at exact head `715226f55dbf838b406f8f6b1eedce2dde26f2e4`:

- `actionlint` on both changed workflows: pass
- `taplo format --check --config taplo.toml`: pass
- `cargo +nightly fmt --all -- --check`: pass
- `cargo metadata --locked --no-deps --format-version 1`: pass
- `cargo test --locked --release --package quantus-miner-api`: pass
- `cargo clippy --locked --package quantus-miner-api --no-deps -- -D warnings`: pass
- `cargo doc --locked --no-deps --package quantus-miner-api`: pass
- `cargo package --locked --package quantus-miner-api --allow-dirty`: pass; six-file 0.3.0 crate packages and verifies successfully
- GitHub Fast Checks and Clippy/Doc: green; the two broad workspace test matrices were still running at review time
- crates.io latest remains `0.2.1`, and no `quantus-miner-api-v0.3.0` remote tag exists

Non-blocking follow-ups: add the README update to the generated PR body's Changes list, pass PR-title data through `env:` instead of interpolating it directly into shell, pin Taplo to the same version as CI, and use `cargo publish --locked`.

**Verdict: approved.**

PR #665, opened 2026-08-13
fix: grandfather RuntimeEnvironmentUpdated in the QPoW digest window

illuzen:
0.9.0 rejected 111-byte upgrade headers that 0.8.x already finalized, so upgraded nodes could not import the chain or replay from genesis.

### DIFF
--- a/client/consensus/qpow/src/lib.rs
+++ b/client/consensus/qpow/src/lib.rs
@@ -775,9 +775,7 @@ mod tests {
 
 	#[test]
 	fn verifier_rejects_undersized_digest() {
-		let digest = Digest {
-			logs: vec![DigestItem::Seal(POW_ENGINE_ID, vec![2u8; 64])],
-		};
+		let digest = Digest { logs: vec![DigestItem::Seal(POW_ENGINE_ID, vec![2u8; 64])] };
 		assert!(digest.encode().len() < DIGEST_LOGS_SIZE);
 
 		let params = BlockImportParams::<TestBlock>::new(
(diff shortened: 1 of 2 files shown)


n13 (approved):
## Review (Grok 4.6)

0.9.0’s exact 110-byte digest window made already-finalized 0.8.x `set_code` headers unimportable: those blocks deposit a 1-byte `RuntimeEnvironmentUpdated` between the pre-runtime item and the seal, so the sealed digest encodes to 111 bytes. `Header::hash()` still copies `min(len, 110)` — same truncation 0.8.x used — so accepting that one leftover restores sync / genesis replay without changing historical hashes.

The helper is in the right place (`qp_header::check_digest_commitment_window`) and both import sites use it: `extract_pow_seal` (before the first `hash()` on a network header) and `import_block` (on `post_header()`, i.e. after the seal is put back). The 110-byte path is unchanged; 111 is allowed only when there is exactly one `RuntimeEnvironmentUpdated`; two REUs, `Other`, and short digests still fail. Tests cover the historical layout plus those rejects. CI is green.

This matches the vendored frame-system change: current WASM no longer deposits REU, so honest 0.9.0+ blocks stay 110 bytes.

**Verdict: LGTM — approve.** Ship this; it is the sync fix.

### Non-blocking

1. **`primitives/header/src/lib.rs:77`** — The carve-out is “any 111-byte digest with one REU”, not specifically `[PreRuntime, REU, Seal]`. Combined with `extract_pow_seal`’s `pop()`, last must still be a seal to get through the verifier, so the historical layout works. Tightening the check to that exact three-item shape would match the comment and shrink the exception. Optional further tightening: only allow it below a known `spec_version` / height so a modified miner cannot mint *new* 111-byte headers.

2. **`client/consensus/qpow/src/lib.rs:419`** — Error text still says “expected 110-byte commitment window” after 111 is a legal length.

3. Worth a one-liner test that a `[PreRuntime, REU, Seal]` header hashes equal to the same bytes truncated to 110. That locks the 0.8.x hash identity this PR depends on.

— Grok 4.6

PR #667, opened 2026-08-13
fix: treat the QPoW digest as a fixed preimage+seal

n13:
## What

The sealed QPoW digest is now a **fixed value** (`QpowDigest`: 32-byte rewards preimage + 64-byte seal), not “whatever list encodes to 110 bytes.”

`Header.digest` remains a Substrate `Digest` on the wire. Changing that type would break decode of every existing block. `sp_runtime::traits::Header` also requires `&Digest` / `&mut Digest`.

## Compatibility

- Same SCALE header encoding. `poseidon_header_hash_is_stable` still passes.
- Canonical `[PreRuntime(32), Seal(64)]` still imports.
- Grandfathered 0.8.x `[PreRuntime, RuntimeEnvironmentUpdated, Seal]` (111 bytes) still imports (#665).
- New: a 110-byte digest that is **not** that shape is rejected (previously accepted).
- After `propose`, a header that is not a single 32-byte preimage is discarded so we do not mine a brick.

## Why

A byte budget plus “please don’t call `deposit_log`” is brittle. Checking **shape** at import and at authoring is the hardening that does not fork the live chain.

A later protocol change can replace the list on the header entirely (or commit `Poseidon(digest)`). That is a circuit + consensus fork and is not this PR.

## Test plan

- [x] `cargo test -p qp-header -p sc-consensus-qpow`
- [ ] Fresh-node sync of Heisenberg (preflight) before any release that includes this

### DIFF
--- a/client/consensus/qpow/src/lib.rs
+++ b/client/consensus/qpow/src/lib.rs
@@ -13,8 +13,7 @@ use sp_consensus_qpow::{QPoWApi, Seal as RawSeal};
 use sp_runtime::traits::Block as BlockT;
 use std::{marker::PhantomData, sync::Arc, time::Duration};
 
-use codec::Encode;
-use qp_header::{check_digest_commitment_window, DIGEST_LOGS_SIZE};
+use qp_header::{check_digest_commitment_window, QpowDigest, DIGEST_LOGS_SIZE};
 
 use crate::worker::UntilImportedOrTransaction;
 pub use crate::worker::{MiningBuild, MiningHandle, MiningMetadata, RebuildTrigger};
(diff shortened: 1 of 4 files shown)


n13 (commented):
Verdict: **APPROVE (opinion)** — no blocking findings at `63958623adfa68d7fe220a647150015de1168194`.

The length-only gap is closed cleanly. `QpowDigest::try_from_sealed` now requires the exact canonical item order, matching engine IDs, and 32-byte preimage / 64-byte seal payloads, while preserving the specific historical `[PreRuntime, RuntimeEnvironmentUpdated, Seal]` encoding. The network verifier applies that gate before its first header hash, `PowBlockImport` repeats it against the reconstructed post-seal header, and the authoring path discards proposals whose runtime added or replaced digest items. The on-wire `Digest` field and stable Poseidon header-hash fixture remain unchanged.

No blocking findings.

Non-blocking compatibility debt: `primitives/header/src/digest.rs:46` accepts the historical REU shape without a height/hash boundary, so a modified miner can still author a new 111-byte legacy-shaped header. That behavior is inherited from the replay fix and is necessary for existing 0.8.x upgrade blocks; a future hard fork could narrow it to known historical blocks if practical.

Validation:

- `cargo test --locked -p qp-header -p sc-consensus-qpow` — passed (14 + 5 tests; 0 failures).
- `cargo +nightly fmt --all -- --check` — passed.
- `git diff --check b9c8849064164824d494b39bf6556960fe8daed6...63958623adfa68d7fe220a647150015de1168194` — passed.
- GitHub CI: format and Analysis (Clippy & Doc) passed; Linux/macOS test matrices were still running at review time.
- `cargo check --locked -p qp-header --no-default-features` could not reach `qp-header`: the unchanged dependency graph fails in `sp-state-machine` because std-only `sp_core::storage::{Storage, StorageMap}` types are unavailable in that standalone feature combination.

GitHub does not permit a formal approval on my own PR, so this is posted as an explicit opinion verdict.

PR #672, opened 2026-08-28
feat: add dependency cooldown

dewabisma:
Basically adding the dependency cooldown workflow. Enforcing in repo rules will be done later because doesn't have permission.

### DIFF
--- a/.github/workflows/dependency-cooldown-audit.yml
+++ b/.github/workflows/dependency-cooldown-audit.yml
@@ -1,0 +1,13 @@
+name: Dependency cooldown audit
+
+on:
+  schedule:
+    - cron: '0 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  audit:
+    uses: Quantus-Network/shared-workflows/.github/workflows/dependency-cooldown-audit.yml@v1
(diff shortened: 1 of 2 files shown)


n13 (approved):
**Reviewer model:** GPT Sol

APPROVE — The PR correctly wires the dependency-cooldown pull-request gate and scheduled/manual audit into the shared `v1` workflows. No blocking findings.

The callers match the shared workflow interfaces and documented event set. The pull-request workflow re-runs for dependency updates plus bypass-label/reason changes, the audit retains both weekly and manual entry points, and both workflows keep the token at `contents: read` without passing secrets.

Validation:

- Reviewed the complete `94c4ca7a...918e21ad` diff and the shared `v1` implementation resolved at `32aabd71d49df3482e4721eeb3f3d6fb5918850f`.
- `git diff --check 94c4ca7a...918e21ad` passed.
- `actionlint` v1.7.12 passed for both added workflow files.
- The live `dependency-cooldown / Dependency cooldown` run passed at this head with only `contents: read`, discovering both `Cargo.lock` and `scripts/js/package-lock.json` and reporting zero newly introduced dependency versions.

As already noted in the PR description, selecting this check in the repository ruleset remains the separate operational step that turns the reporting workflow into an enforced merge gate.

PR #671, opened 2026-08-28
revert: remove pubkey cache and sig-only transactions

n13:
## Summary

- Restore self-contained full ML-DSA signatures as the runtime signature format.
- Remove `pallet-pubkey`, sig-only signature variants, cache runtime API/storage, reap accounting, cache-specific transaction extension wiring, and cache tests/docs.
- Keep the independently useful self-describing `Multisig::execute` interface, canonical stored-call binding, weight regressions, transaction version 4, and wormhole matcher recursion.

## Scope

This rolls back the public-key caching feature before release. The multisig edge-case fix from the same original commit is deliberately retained; the multisig pallet and its governance call sites are unchanged from `main` in this PR.

## Testing

- `cargo +nightly fmt --all -- --check`
- `cargo test --locked -p qp-dilithium-crypto -p pallet-multisig` (77 passed)
- `cargo test --locked -p quantus-runtime --lib multisig_execute_counts_its_resubmitted_call`
- `cargo test --locked -p quantus-runtime --test mod --features fast-governance` (41 passed, 1 ignored)
- `cargo check --locked --workspace --features runtime-benchmarks,try-runtime`

PR #35, opened 2025-05-27
Feature/installer script

n13:
Installer script

## All in one install script

run like this 
```sh
./install-quantus-node.sh
```

## Functions

- Installs latest binary release for the operating system (Mac OS and Linux)
- Generates node identity if needed
- Generates rewards address if needed
- Prints command to run miner

## Notes
A lot of the code here was already merged to main, so this PR is only scripts!

PR #22, opened 2025-05-21
feat: vest airdrop rewards

dastansam:
- replace `vesting` fork with Substrate official
- vest airdrop rewards
- general cleanup and refactoring of the pallet
    - unified storages into one about airdrop info. airdrop funders can now specify the vesting period and delay
    - remove unnecessary checks, etc.

Note: benchmarks are not updated yet, will be a part of a bigger general benchmarks/weight cleanup PR later

Issue #19, opened 2025-05-21
Change CLI


n13:
Quantus node CLI commands update

Remove: generate, inspect, insert
Keep: generate-node-key, inspect-node-key, quantus, help

**Reasoning**
generate is used to create EC keys (schnorr...) - this happens outside our code, in the sc-cli crate
inspect is complex and for EC keys
insert presumably also for EC keys

generate-node-key creates node identity key for libp2p - we still need this
inspect node key - inspects such keys
quantus - creates our keys

```
> quantus-node key --help
Commands:
  generate-node-key  Generate a random node key, write it to a file or stdout and write the corresponding peer-id to stderr
  generate           Generate a random account
  inspect            Gets a public key and a SS58 address from the provided Secret URI
  inspect-node-key   Load a node key from a file or stdin and print the corresponding peer-id
  insert             Insert a key to the keystore of a node
  quantus            Generate a quantus address
  help               Print this message or the help of the given subcommand(s)

```

n13:
Done

PR #675, opened 2026-08-29
Merge chain-private main into chain

n13:
## Purpose

Merge `Quantus-Network/chain-private` `main` (`011b2f58`) into public `chain` `main` (`308ba838`).

They split at `803094f7` (`ci: Miner API version bump to quantus-miner-api-v0.2.1`). This branch starts from public `main` (all public commits since the split) and merges private `main` on top. Non-overlapping changes from both sides are kept. Conflicts were resolved per the decisions below; chain-private wins except where this PR explicitly chose otherwise.

This is a new unification PR. It does not continue chain-private #41.

## What public unique work is kept (no conflict)

- QUIC-authenticated external-miner protocol, miner API v0.3.0, auth token + TLS pin files (`#662`, `#664`)
- zk-tree batched root recomputation (`#663`)
- dependency-cooldown CI (`#672`)
- release preflight docs, v0.10.0 bump
- pubkey-cache / sig-only txs were added and then reverted on public (`#661`, `#671`); they stay reverted

## Conflict resolutions

Nine files conflicted (20 hunks), grouped as originally presented. Choices are recorded in full.

---

### Group A — QPoW digest window

**Question:** Public (`#665`, after `#667` was reverted by `#670`) accepts exact 110 bytes, or 111 with exactly one `RuntimeEnvironmentUpdated`, at any height, and rejects short encodings (they pad to a colliding hash). Private (`#35`) uses a length cap only: blocks `<= 1_000_000` may be 111 bytes (any extra byte); after that, 110. It does not reject short encodings.

**Choice:** Allow both 110 and 111 digests. The block cutoff is OK (keeps testnet history importing). **Private API:** `max_encoded_digest_size` + `LEGACY_DIGEST_CUTOFF = 1_000_000`. Private verifier tests (invalid PoW / wrong-length seal, report 88219) kept.

---

### Group B — mining loop (`node/src/service.rs`)

**Question:** Public `#669` pauses on `is_major_syncing()`, clears the stored miner job, and pauses immediately when offline (no 30s grace), so external miners are not handed a stale job they grind for the whole sync. Private `#35` never reads network sync state (forged peer height, report 88224). Bitcoin-style: stale-tip latch + 30s no-peer grace. `--force-authoring` bypasses.

**Choice:** Split two different things:

- **When the node is allowed to author** — private. Bitcoin-style tip freshness + 30s no-peer grace. Dismiss public's `is_major_syncing()` pause (forged peer height, report 88224).
- **How the node talks to external miners** — public. That is the miner protocol (see Stitches). Private did not change it after the split.

---

### Group C — `Multisig::execute` ABI + wormhole weight walker

**Question:** Public `#661`, kept by `#671`: `execute(origin, address, id, call)`. Executor resubmits the call; byte-equal bind to storage. Hardware wallets can see the inner call. Propose also requires canonical encoding (no trailing bytes). Private `#11` + `#22`: `execute(origin, address, id)` only. Decode stored bytes with `decode_with_depth_limit` (depth-bounded, not all-consuming). Walker must read storage to price wormhole leaves. Recovery and extra Utility wrappers are already gone in this merge (private-only).

**Choice:** Keep public's call-carrying `execute` (clearsigning / self-describing weight / edge-case fix). Stitch:

- `propose` still depth-limits the decode (private Immunefi #88969) **and** requires canonical `decoded.encode() == stored bytes` so execute's byte-bind cannot strand a proposal
- wormhole `count_transfers` recurses into `execute { call, .. }` (public) and only walks `batch_all` (private utility shrink)
- storage-walk reservation and `count_stored_execute` removed; they existed only for the opaque-id ABI

---

### Group D — runtime versions

**Question:** Public: `spec_version: 145`, `transaction_version: 4` (comment: execute carries the call). Private: `spec_version: 147`, `transaction_version: 6` (recovery removal, FIPS context, fees, etc.).

**Choice:** Highest from private: **spec 147, transaction 6**. Public version numbers dismissed.

Note: this tree's `Multisig::execute` encoding is the public call-carrying one. Clients that spoke private's no-call execute need the new argument. `transaction_version` is left at 6 as decided.

---

### Group E — `docs/RUNTIME_SURFACE.md` extension list

**Question:** Public: HS whitelist only; execute carries inner call; still mentions `Recovery::close_recovery`. Private: adds `HighSecurityTxQuota` (16 txs / 24h); execute “walks the stored proposal”; ChargeTransactionPayment note about `HighSecurityFungibleAdapter` zero-tip.

**Choice:** Private text. The execute/walker sentence was updated to “recurses into the resubmitted inner call” so it matches Group C. Quota + tip adapter were already auto-merged.

---

### Group F — `pallets/vesting/src/weights.rs` (comment)

**Question:** Public (`#663` zk-tree batching): `end_schedule` / `retarget_schedule` (= 4) … “flat **marginal** insert cost”. Private (`#19` retarget is a same-owner swap, not a payout): `end_schedule` (= 4) only … “flat **circuit-depth** insert cost”.

**Choice:** Private comment / retarget-has-no-leaf. Flag: auto-merged public zk-tree made `INSERT_LEAF_*` the cheap *marginal* price. Private `payout_weight` then undercharges the benchmarked base (`payout_weight_never_undercharges_the_benchmarked_base` failed: 1.275s vs 1.305s ref_time). A clamp to the measured base is applied so we do not ship an undercharge; regenerate vesting benchmarks as follow-up. Public `payout_weight_unclamped` helper is not carried.

---

### Group G — `pallets/wormhole/src/weights.rs` (comment)

**Question:** Public: one leaf per exit; **marginal** insert; root batched in `on_initialize`. Private: exit leaf **plus** miner volume-fee leaf (and aggregator rebate on public batch); priced at `CIRCUIT_MAX_TREE_DEPTH`.

**Choice:** Private comment and extra fee-leaf accounting. Flag: with public batched `INSERT_LEAF_*`, the per-exit tail is no longer ~10× the one-exit settled weight (ZK verify dominates). Two tests that required `settled(1) * 10 < declared` were relaxed to `settled(1) < declared`; they still prove unused-exit refunds.

---

### External miner protocol is public (nothing was deferred)

The miner **wire protocol** and the node’s **session with an external miner** are the public ones. After the split, private never touched `miner-api/` or `node/src/miner_server.rs`; those files in this merge are public `#662` / `#664` / `#669`. There is no private miner-API patch to land later.

What that means in this tree:

- **Protocol:** ALPN `quantus-miner/2`, authenticated `Ready { token }`, TLS cert + SHA-256 fingerprint pin, 1 KiB frame limit (`miner-api` v0.3.0).
- **Server:** `MinerServer::start(MinerServerConfig)` with auth-token path and TLS dir. If `--miner-listen-port` is set and the server cannot start, the node **fail-closes** (no silent fallback to local mining).
- **Job lifecycle:** public `handle_external_mining` — snapshot `worker_handle.version()` before broadcasting, treat a version change as supersession, do not blame the miner for a seal that arrived after a template rebuild. `clear_current_job` exists on the public server so a connecting miner is not handed a stored job while authoring is paused.
- **Authoring policy (not protocol):** private `mining_loop` still decides *whether* to mine (tip freshness + 30s offline grace, never `is_major_syncing()`). That is Group B, not the miner API.

Operators must upgrade external miners and provision token + cert pin files before deploying this.

### Other stitches

- HS `propose` benchmark no longer pads trailing bytes (canonical-encode check from C would reject them).
- `RUNTIME_SURFACE.md` execute walker wording aligned with C.

## Follow-up (not in this PR)

- Regenerate vesting (and wormhole) weights against the batched zk-tree so the vesting clamp can go away. Not a miner issue.

## Verification (local)

- `cargo +nightly fmt --all`
- `SKIP_WASM_BUILD=1 cargo test --locked -p qp-header --lib` — 6 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p sc-consensus-qpow --lib` — 9 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-multisig --lib` — 60 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-vesting --lib payout_weight_never_undercharges` — passed after clamp
- `SKIP_WASM_BUILD=1 cargo test --locked -p pallet-wormhole --lib` — 85 passed, 3 ignored
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime --lib` — 68 passed
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime --test mod --features fast-governance` — 56 passed, 1 ignored
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-node` — 59 passed (includes miner auth/TLS)

## Operational note

Same as the miner-protocol stitch above: this is public `quantus-miner/2` with auth + TLS pinning. External miners that still speak the pre-`#662` protocol will not connect.


n13:
Veriifed spec 147 / tx 6 is new vs the last release - good

n13 (commented):
**Reviewer model:** GPT Sol

I reviewed exact head `4bbb553f9ca709e19d62f88765675c918a42ccc4` against base `308ba838d585a8dc8787e18d55ee15e5a4c41d76`. The merge has the following blockers:

1. **[P1] Restore exact QPoW digest-shape validation.** `client/consensus/qpow/src/lib.rs:419-428` and `:242-248` now reject only encodings larger than the height-dependent cap. `Header::hash()` pads every shorter digest into the fixed 110-byte commitment window (`primitives/header/src/lib.rs:297-314`), so a seal-only or otherwise undersized/noncanonical digest reaches hashing and PoW verification even though a canonical sealed header is exactly 110 bytes (with only the defined historical 111-byte shape allowed). Reinstate the exact allowed-shape check before every network-header hash/import and restore a regression test for an undersized digest.

2. **[P1] Reserve an outer decode level for `Multisig::execute`.** `propose` accepts an inner call at `MAX_EXTRINSIC_DEPTH` (`pallets/multisig/src/lib.rs:52-68,608-612`), but `execute` carries that call inside a boxed outer `RuntimeCall` (`:1098-1103`). FRAME decodes the complete extrinsic with the same maximum depth, so a proposal accepted at the boundary cannot later be encoded as a valid `execute` extrinsic. Reduce the stored-call ceiling by the wrapper depth and add exact boundary tests proving every accepted proposal is executable.

3. **[P1] Charge the two recursive call walks on multisig execution.** `WormholeProofRecorderExtension::weight()` and `validate()` both call `count_transfers` (`runtime/src/transaction_extensions.rs:657-690`), but a no-transfer `Multisig::execute` receives zero extension weight; the regression test at `:1618-1662` explicitly preserves that result. An unauthorised/nonexistent execute then refunds the pallet weight to one DB read (`pallets/multisig/src/lib.rs:1106-1117`). Repeated failing executions carrying broad/deep no-transfer call trees can therefore consume two caller-controlled traversals each while returning that work to block capacity. Reserve both worst-case walks for every execute and retain the reservation on early errors.

4. **[P1] Clear the external-miner job whenever authoring is paused.** The stale-tip, timestamp-error, and offline branches disable the worker (`node/src/service.rs:409-446,451-469`) but never call `MinerServer::clear_current_job`. `broadcast_job` retains the last request and hands it to newly connecting miners (`node/src/miner_server.rs:149-156,211-213`); `clear_current_job` exists only in tests and has no production caller. A miner connecting during a pause is therefore sent stale work and grinds it until a later broadcast. Clear the server-side job on every enabled-to-disabled transition (without reintroducing peer-height-controlled authoring).

5. **[P1] Migrate recovery state and reserved deposits before removing the pallet.** Runtime index 16 is removed (`runtime/src/lib.rs:293-294`), but `Migrations` contains only wormhole and treasury migrations (`:206-214`). Existing `Recoverable` configurations reserve `ConfigDepositBase + FriendDepositFactor`, and `ActiveRecoveries` reserve `RecoveryDeposit`; after this upgrade there is no call that can remove those records or release/repatriate those balances. Current public RPC state was empty when checked, but that is only a snapshot and this patch has no upgrade-time zero-state gate. Add a migration that cleans the storage and resolves every reserve, or enforce an atomic pre-upgrade invariant that makes removal safe.

6. **[P2] Repair the required release gate after removing recovery.** `docs/RUNTIME_UPDATE.md:218-220` says the recovery exercise phase is gone, while `docs/RELEASE_PREFLIGHT.md:136-142` still requires it and forbids skipping it. The current `quantus-cli` main (`2049564c`) also includes `Phase::Recovery` in its default set (`src/cli/exercise/mod.rs:127`). Consequently the documented hard-gate command cannot pass against this runtime. Land the matching CLI/runtime metadata update and make the two release documents agree before this becomes a release candidate.

Validation performed:

- `git diff --check`, `cargo +nightly fmt --all -- --check`, and Taplo format check passed.
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-node`: 59 passed.
- Affected crate/runtime suite: 727 passed, 6 ignored, 0 failed.
- Dependency-cooldown checks passed. The exact-head Fast Checks job remains queued with no steps started.
- Targeted `cargo clippy --all-targets -- -D warnings` was blocked by five existing lints in vendored `frame/support-procedural-tools/src/lib.rs`; none is in this PR's changed lines.

n13:
Stacked follow-up for the execute decode-depth issue (not present on either `main`; it is created by combining private's inner depth cap with public's call-carrying `execute`): #676.

`MAX_MULTISIG_CALL_DEPTH` drops to `MAX_EXTRINSIC_DEPTH - 1` so every accepted proposal still encodes as a valid `execute` extrinsic. Left off this merge PR on purpose.

n13:
Stacked follow-up for execute inner-call walk charging (not on either `main` after call-carrying execute): #677.

A no-transfer `Multisig::execute` reserved zero extension weight while `weight()`/`validate()` still walked the inner call; failing executes refunded the pallet to one read. #677 reserves two worst-case walks for every execute and keeps them on early errors. Left off this merge PR on purpose.

n13:
1. Addressed the miner job clearing - adopting strategy from main where the new miner protocol came from

2. Recovery pallet removal (index 16) will not get a storage/deposit migration.

Testnet never exposed recovery to clients, so there are no live `Recoverable` / `ActiveRecoveries` records or reserved deposits to unwind. Mainnet will not ship the pallet at all, so there is no prior state to migrate. Leftover empty prefixes, if any, stay vacant like the old assets/sudo indices.

n13:
No recovery migration will be done! Nobody used this feature on testnet, it's not exposed in any client

On mainnet, we won't have it at all.

No migration needed

We updated the docs to no longer refer to recovery

n13:
All issues have been addressed, some with new PRs some with merge fixes in here.

Making a new review.

n13 (commented):
**Reviewer model:** GPT Sol

I re-reviewed exact head `e2d1f1b8ac88d6c0f15774fe29ba418e5ea89cca` against base `308ba838d585a8dc8787e18d55ee15e5a4c41d76`, including all three commits added since the prior review. The QPoW digest-window validation and authoring-pause miner-job clearing now address those findings. Given the stated deployment facts and the earlier live empty-state check, I accept removal of Recovery without a migration as an operational choice; rechecking that zero-state precondition immediately before the testnet upgrade remains important.

~Three~ Zero blockers remain:

~1. **[P1] Reserve the outer codec level required by `Multisig::execute`.** `MAX_MULTISIG_CALL_DEPTH` still equals FRAME's full `MAX_EXTRINSIC_DEPTH` (`pallets/multisig/src/lib.rs:52-68`), and `propose` accepts calls at that boundary (`:600-612`). The same call must later sit inside `execute`'s `Box<RuntimeCall>` (`:1098-1103`), which consumes another decode level, so an accepted boundary proposal cannot be submitted as a valid execute extrinsic. #676 contains the right correction and a boundary regression test, but that change is not in this PR's head. Include the one-level reservation before merging #675.~

**=> This is not working in either chain or chain-private, therefore we made a separate PR #676**


~3. **[P1] Charge both recursive inner-call walks on every multisig execute.** `weight()` and `validate()` each call `count_transfers` (`runtime/src/transaction_extensions.rs:657-690`), which recursively walks `Multisig::execute`'s caller-controlled inner tree (`:476-515`). A no-transfer execute reserves zero extension weight, and the current regression test explicitly confirms that a failed non-signer call keeps a zero reservation (`:1618-1662`) while the pallet reports only its early storage-read work. Repeating broad/deep, no-transfer failing executes therefore consumes two uncharged traversals and returns that capacity to the block. #677 reserves two worst-case walks and retains them on early errors, but it is also absent from #675's head. Bring that fix into the merge.~

**=> This is not working in either chain or chain-private, therefore we made a separate PR #677**


~4. **[P2] Make the required release exercise command compatible with the removed pallet.** `docs/RELEASE_PREFLIGHT.md:131-143` now removes `recovery` from the documented default phases, but live `Quantus-Network/quantus-cli` `main` at `96ec0c9b` still includes `Phase::Recovery` in `default_set()` ([source](https://github.com/Quantus-Network/quantus-cli/blob/96ec0c9b49df3b214b588b6533c3199e42c67baa/src/cli/exercise/mod.rs#L103-L134)); there is no open matching CLI PR. Consequently the mandated `quantus exercise --fail-fast` gate still attempts a phase for runtime metadata that #675 removes. Land the coordinated CLI update and make the preflight use that compatible version.~

**=> CLI and other clients must be updated later!** 


Validation performed:

- `git diff --check`, `cargo +nightly fmt --all -- --check`, and `taplo format --check` passed.
- `SKIP_WASM_BUILD=1 cargo test --locked -p qp-header --lib`: 12 passed.
- `SKIP_WASM_BUILD=1 cargo test --locked -p sc-consensus-qpow --lib`: 10 passed.
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-node`: 59 passed.
- Both dependency-cooldown checks are green. The exact-head Fast Checks job is still queued with no runner or steps.

n13:
All these blockers are moot - they're addressed in additional PRs. 

Ready for final sign-off @illuzen

illuzen (approved):

PR #678, opened 2026-08-30
feat: staging-mainnet chain spec, fast-upgrade governance track

n13:
Ports the staging-mainnet work from `chain-private` PR #18 onto the public repo. This is the last piece of the chain-private move.

Stack: #675 (the merge) → #676 → #677 → this PR. It targets `n13/execute-walk-weight`, so review it after #677.

## What this does

**1. Fast-upgrade governance track.** A runtime WASM is hundreds of KiB, so it can never be a tech-referenda proposal: `MaxReferendaProposalSize` caps a `Lookup` preimage at 64 KiB. Upgrades have to go `authorize_upgrade(hash)` first, then the permissionless `apply_authorized_upgrade(wasm)`. Root can already do that, but only on the 1-day track, which is too slow for an emergency.

So there is now a second track:

- `pallet_custom_origins` at runtime index 23 contributes a single `Origin::FastUpgrade` to `RuntimeOrigin`. It has no calls, no storage, no events and no config items; its whole body is one `#[pallet::origin] enum`. A pallet is the only way to add a `RuntimeOrigin` variant, and a distinct non-Root origin is required because `TracksInfo::track_for` maps proposal origin to track, so two tracks cannot both be Root. Reusing `Signed(_)` is the impersonation hole `track_for` already rejects (#91247/#91270).
- Track 1 (`fast_upgrade`): 80%/80% constant curves, which is 8-of-10 with a ten-member collective, and 10-minute prepare, confirm and enactment windows. Track 0 keeps 61%/60% and its 1-day windows for everything else.
- `frame-system` gains `Config::AuthorizeUpgradeOrigin`, defaulting to Root. The runtime sets it to `Root or FastUpgrade`. `set_code` and `authorize_upgrade_without_checks` stay Root-only, so the short windows can publish an upgrade hash and nothing else.

`quantus-cli` already ships this flow (`tech_referenda` FastUpgrade submission, `system.apply_authorized_upgrade`), generated against a runtime that has the `Origins` pallet. The public runtime does not have it yet, so the CLI's upgrade path does not currently work against public builds.

**2. Cheaper governance.** Tech decision deposit 1000 → 10 UNIT, referendum submission deposit 100 → 10 UNIT, both scaled by `FEE_SCALE`. At the old numbers each launch signer would need over 1100 QUAN of liquid balance at genesis just to carry one referendum end to end. The maximum preimage deposit for a 64 KiB proposal is about 6.65 UNIT at the current scale, still under the 10 UNIT submission bond, so noted bytes stay collateralized after `unnote`.

**3. `staging_mainnet` genesis preset.** The mainnet dress rehearsal. Genesis is built by `mainnet_config_genesis`, the same function the eventual mainnet preset will call. The only difference is the treasury multisig nonce, 1 for staging and 0 for mainnet, which gives the two chains distinct treasury accounts and therefore distinct genesis hashes. Everything else is 1:1, so whatever is rehearsed is what launches.

- Treasury: 6-of-10 multisig of `MAINNET_TREASURY_SIGNERS_SS58`, which is also seeded as the tech collective.
- Signer endowments are derived, not written down. `governance_treasury_signer_seed` sums the treasury multisig bootstrap, the referenda submission and decision bonds, a maximum-size preimage deposit, and a small scaled fee headroom. It is deliberately the minimum that lets one treasurer bootstrap the multisig and carry one referendum end to end. Changing `FEE_SCALE` or any deposit re-derives it instead of stranding governance, and a genesis test pins every signer balance to the formula. Read the actual figures out of the generated spec's `balances` section.
- `mainnet_vesting_schedules` is a DUMMY scaffold behind `MAINNET_VESTING_FINALIZED = false`. While that flag is false only staging-mainnet builds; a mainnet preset panics. It carries stand-in team, backer and ecosystem entries plus 20 HD rehearsal accounts (`staging-0`..`staging-19`) so those keys can exercise claim.
- `utc_midnight_ms` replaces hand-computed epoch constants, so every vesting date is written as the date it documents. Tests pin all three.

**4. Docs and `--force-authoring`.** New runbooks `docs/STAGING_MAINNET_LAUNCH.md` and `docs/CHAINSPEC_CREATION.md`. The upgrade docs are rewritten for the authorize-then-apply flow. `--force-authoring` gets clearer help text: it is required to start mining on a new chain from a single validator, because at genesis there are no peers and the best block is always older than `--max-tip-age`. `--dev` already implies it.

## Runtime compatibility

The new origin variant appends at index 23, after `Vesting` at 22, so existing encoded `proposal_origin` values in referenda storage stay valid. No migration. No call indices move, so the call ABI is unchanged.

Runtime metadata does change, so hardware-wallet metadata needs regenerating for whatever release carries this.

No `spec_version` bump: `runtime/src/lib.rs` says not to increment it in feature PRs. No `transaction_version` bump either, since the signed extrinsic encoding is unchanged.

`pallet-referenda` stores each referendum's deposit amount, so the bond reduction does not affect refunds of already-placed deposits on live chains. Only new submissions see the new numbers.

## Differences from chain-private #18

- Dropped one rustfmt-only hunk in `node/src/command.rs` that rewrote `None =>` into `None => { … }`. It conflicted with the miner-flag block that landed publicly, carries no behaviour, and `cargo +nightly fmt --all -- --check` passes without it.
- Added pallet 23 to `docs/RUNTIME_SURFACE.md`. That file enumerates every pallet index in a table and gives each one a section; #18 added the pallet to the runtime but to neither list.

Everything else is the same change, split into four commits for review.

## Remaining before launch

1. Merge, cut a release tag, run `./scripts/genesis_generate_spec.sh <tag> staging_mainnet`.
2. Commit the JSON and the `"staging_mainnet"` `include_bytes!` arm in `load_spec`.
3. Boot the first server and add its bootnode. See `docs/STAGING_MAINNET_LAUNCH.md`.

### DIFF
--- a/client/cli/src/commands/run_cmd.rs
+++ b/client/cli/src/commands/run_cmd.rs
@@ -140,7 +140,11 @@ pub struct RunCmd {
 	#[arg(long, conflicts_with_all = &["alice", "bob", "charlie", "dave", "eve", "ferdie", "one"])]
 	pub two: bool,
 
-	/// Enable authoring even when offline.
+	/// Enable authoring even when offline (no connected peers).
+	///
+	/// Required to start mining on a new chain from a single validator,
+	/// for example at mainnet or staging-mainnet genesis. Without this flag,
+	/// mining pauses until at least one peer is connected.
 	#[arg(long)]
 	pub force_authoring: bool,
 
--- a/docs/CHAINSPEC_CREATION.md
+++ b/docs/CHAINSPEC_CREATION.md
@@ -1,0 +1,77 @@
+# Creating a Chain Spec (JSON)
+
+How to produce a committed, raw JSON chain spec for a network profile
+(`heisenberg`, `planck`, `staging_mainnet`). The genesis runtime in the JSON is
+the **published release artifact**, not a local build — anyone can reproduce
+and verify it.
+
+## Naming convention
+
+For a profile `<profile>` (e.g. `staging_mainnet`):
+
+| Name | Meaning |
+| --- | --- |
+| `<profile>_live_spec` | `--chain` id that builds genesis from the preset compiled into the node (used only during generation) |
+| `node/src/chain-specs/<profile-with-dashes>.json` | the committed raw spec |
+| `<profile>` | `--chain` id that loads the committed JSON (what operators use) |
+
+## Prerequisites
+
+- The runtime preset exists in `runtime/src/genesis_config_presets.rs` and is
+  wired up in `node/src/chain_spec.rs` (`<profile>_chain_spec()`) and in
+  `load_spec` (`node/src/command.rs`) under `"<profile>_live_spec"`.
+- Clean git working tree (the script refuses otherwise).
+- `jq`, `curl`, `xxd`, and a Rust toolchain that builds the node.
+
+## Steps
+
+1. **Cut a release tag.** CI publishes the runtime WASM
+   (`quantus-runtime-vNNN.compact.compressed.wasm`) as a release asset on
+   `Quantus-Network/chain`.
+
+2. **Generate the spec from the tag:**
+
+   ```sh
+   ./scripts/genesis_generate_spec.sh <release-tag> <profile>
+   ```
+
+   The script:
+   - creates branch `genesis/<profile>/<release-tag>` at the tag and builds
+     the node there, so genesis state comes from exactly the released code;
+   - runs `build-spec --chain <profile>_live_spec --raw
+     --disable-default-bootnode` into
+     `node/src/chain-specs/<profile-with-dashes>.json` (without
+     `--disable-default-bootnode`, a spec with no bootnodes gets a throwaway
+     `/ip4/127.0.0.1` bootnode injected);
+   - downloads the release's compressed runtime WASM and splices it into the
+     JSON's `:code` key (`.genesis.raw.top."0x3a636f6465"`), so the genesis
+     runtime is byte-identical to the published artifact.
+
+3. **Wire the JSON into the node.** Add a `"<profile>"` arm to `load_spec` in
+   `node/src/command.rs`:
+
+   ```rust
+   "staging_mainnet" => Box::new(chain_spec::ChainSpec::from_json_bytes(include_bytes!(
+       "chain-specs/staging-mainnet.json"
+   ))?) as Box<dyn sc_service::ChainSpec>,
+   ```
+
+4. **Record the genesis hash.** Run `quantus-node --chain <profile> --tmp` and
+   note the genesis hash it prints on startup.
+
+5. **Verify the genesis runtime against the release** (against any running
+   node of the new chain):
+
+   ```sh
+   ./scripts/genesis_verification.sh --release-tag <release-tag> --node-url http://localhost:9944
+   ```
+
+6. **Add bootnodes** once the infrastructure exists: edit the JSON's top-level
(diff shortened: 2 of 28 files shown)


n13 (commented):
**Reviewer model:** GPT Sol

Blocking finding:

1. **[P1] Update the mainnet runtime-upgrade security model for the new FastUpgrade track** (`docs/TECH_COLLECTIVE_GOVERNANCE_TUNING.md:3-27`, `:55-88`, `:126-137`). This document explicitly scopes itself to the runtime-upgrade track, but after this PR moves upgrades to track 1 it still publishes track 0's parameters: 2-hour prepare, 24-hour confirm/enactment, 61%/60% thresholds, and the five-member threat table with a full-day reaction window. The actual upgrade lane at `runtime/src/governance/definitions.rs:146-163` is 10-minute prepare/confirm/enactment with 80%/80% thresholds (8-of-10 at staging/mainnet genesis). Operators and reviewers following this mainnet tuning/incident-response document would therefore materially overestimate the time available to veto an upgrade and apply the wrong quorum model. Document both tracks, make the FastUpgrade section describe the supported runtime-upgrade flow, and update its timing, quorum, and security summary accordingly.

No other blocking findings were found in the runtime origin separation, staging genesis/accounting, or authorize/apply operator flow.

Validation at exact head `a90fd34c52cd7f75fce33970551cce04a9ec581f` against base `9811c6162e3ad70a12e08833e349ca59a152793d`:

- Complete 27-file diff and affected call paths inspected; the mid-review base merge changed only formatting and preserved the PR patch ID.
- Runtime tests: 73 library tests and 62 integration tests passed; 1 unrelated test is intentionally ignored.
- Force-authoring/node CLI tests: 3 passed.
- Runtime benchmark/try-runtime feature check, nightly format, `git diff --check`, and shell syntax check passed.
- Quantus CLI PR #143 is merged with the matching FastUpgrade authorization plus exact-WASM application flow.
- Hosted format, dependency-cooldown, and analysis checks pass; Linux/macOS test matrices were still running at review time.

n13:
Addressing the doc update

n13:
Addressed the review's P1 in `docs/TECH_COLLECTIVE_GOVERNANCE_TUNING.md`:

- Scope now covers both tracks and states the supported upgrade route: vote on the code hash on track 1 (`Origins::FastUpgrade` → `system.authorize_upgrade` only), then anyone applies the exact WASM via `apply_authorized_upgrade`. `set_code` is never the route (64 KiB proposal cap, hardware-wallet limits); a track-0 Root `authorize_upgrade` remains as the slow fallback.
- Timing table per track (track 1: 10 min prepare / confirm / enactment, ≈30 min end to end; track 0 unchanged), including that `apply_test_timing` covers both tracks.
- Threshold and verification tables redone for the 10-member mainnet collective on both tracks (track 1: 8-of-10, 8a/2n passes, 7 ayes fail; track 0: 6 ayes, 4 nays block), with the 5-member case kept as a one-liner.
- Reaction window and incident response rewritten: 10-minute confirm on the fast lane, Root cancel/kill unreachable within it, defense is 3 honest nays; `AuthorizedUpgrade` single-slot semantics.
- Security summary now has separate threat tables for track 0 and track 1 and the per-track design assumptions.
- Replaced the stale `file:line` references with symbol names, updated deposits (10 × `FEE_SCALE`) and `spec_version` (147), dropped the mention of a community referenda instance that no longer exists.

Every symbol cited was checked against the branch's source.

illuzen (approved):

PR #682, opened 2026-08-31
chore: clear all clippy and rustdoc warnings

n13:
Takes `cargo clippy --workspace` from 1,442 warnings to 0 and `cargo doc` from 38 to 0, and makes CI fail on any new warning.

## Why so many

Almost all of the warnings were in the crates copied from polkadot-sdk and litep2p. Upstream compiles those with a workspace lint policy that turns off clippy's style lints (846 of our warnings were "tabs in doc comments"). The copies lost that policy when they came into this repo. Four crates had already pasted it into their own `Cargo.toml`.

## What this does

- Upstream's lint policy is defined once under `[workspace.lints]`. **Only the vendored crates inherit it** (34 crates, `[lints] workspace = true`). The 21 first-party crates have no `[lints]` entry and keep clippy's defaults, exactly as on `main`, so the `-D warnings` checks on the published crates and in pm-quality-check keep their teeth. (The first push of this PR got that wrong and had every crate inherit; fixed in `build: keep first-party crates on clippy defaults`.)
- The `panic` / `unwrap_used` / `expect_used` denies move to a crate-level `#![deny(...)]` in the three crates where they were actually active: `quantus-runtime`, `sc-service`, `sc-transaction-pool`. Same rule, same crates. litep2p's own `warn` on those lints (68 permanent warnings) is removed; say if you want it back.
- CI's workspace `cargo clippy` and `cargo doc` steps now run with `-D warnings`, like the publish workflows already do. The separate published-crates clippy step was a subset of the workspace run and is removed.
- Real fixes for everything the policy doesn't cover: 12 clippy and 26 rustdoc warnings in the workspace run, 9 more first-party warnings once those crates were back on defaults, and 23 lints in test targets that made pm-quality-check red. Two small API changes: `record_high_security_tx` / `record_hs_quota` return `DispatchResult` with a new `HighSecurityTxQuotaExceeded` error instead of `Result<(), ()>`, and the node's `take_trimmed_secret` returns `Option`.

## Checked

- Workspace clippy 0, rustdoc 0, fmt and taplo clean, `cargo check --features runtime-benchmarks,try-runtime` clean.
- The exact CI gate commands pass locally: `cargo clippy -p qp-dilithium-crypto --no-deps -- -D warnings`, same for `quantus-miner-api`, and all 17 pm-quality-check crates with `--all-targets --all-features -- -D warnings`.
- Tests pass in every crate whose code changed (pallet-referenda, -wormhole, -reversible-transfers, -scheduler, -mining-rewards, -multisig, sc-transaction-pool, sc-consensus-qpow, quantus-node, litep2p) and the runtime integration tests that use the quota helper.


client/transaction-pool/src/lib.rs
```diff
@@ -21,6 +21,7 @@
 #![recursion_limit = "256"]
 #![warn(missing_docs)]
 #![warn(unused_extern_crates)]
+#![deny(clippy::panic, clippy::unwrap_used, clippy::expect_used)]
```
n13, line 25: This was just elsewhere before, we didn't invent this in this PR

"What changed in #682 is only where it's written, not what's enforced. On main the same three lints were deny in [workspace.lints.clippy] and only 5 crates opted in — of which two (sc-network, sc-network-transactions) immediately cancelled it with a crate-level #![allow(...)]. So the deny was effectively active in exactly quantus-runtime, sc-service, and sc-transaction-pool. Since every crate now inherits the workspace table and vendored code is full of unwrap(), the deny couldn't stay there without breaking the build — so it moved into those three crates as this attribute. Same three crates, same three lints, just declared in the crate instead of in Cargo config."

n13 (commented):
**Reviewer model:** GPT Sol

Blocking finding:

- **Keep the existing strict Clippy gates strict.** `Cargo.toml:80` sets `clippy::all = allow`, and the new `[lints] workspace = true` entries apply that permissive policy to first-party crates too, including `miner-api/Cargo.toml:22` and `primitives/dilithium-crypto/Cargo.toml:55`. The dedicated checks in `.github/workflows/ci.yml:104-105`, both publish workflows, and the 17-crate post-merge matrix rely on `-D warnings`; that flag does not re-enable Clippy lints explicitly allowed by the inherited manifest policy. I reproduced the precedence with the same lint table: a `clippy::let_and_return` violation passes `cargo clippy -- -D warnings` and fails only when `-W clippy::all` is added. As written, those checks silently stop enforcing the default Clippy set despite the CI comment that published crates remain strict. Please scope the permissive upstream policy to vendored crates, or explicitly re-enable `clippy::all` in every strict invocation (including release workflows).

I found no other blocking issue in the behavioral edits: the iterator, `Option`, modulo, `Copy`, import, and rustdoc rewrites are semantics-preserving.

Validation: reviewed the complete `4c4a8f2f...f5102b8d` diff and affected lint/test paths; confirmed all 53 workspace members inherit the new table and the prior production deny coverage is retained; `git diff --check`, `cargo metadata --locked --no-deps`, `taplo format --check --config taplo.toml`, and `scripts/fmt.sh --all -- --check` passed. Exact-head Linux check/test, format, analysis, Clippy, and rustdoc CI jobs are green; the macOS matrix job is still running.

n13 (commented):
Not reviewing again — no new commit or comment since the last review.

n13:
Addressed. The first push had every crate inherit the workspace table, which put `clippy::all = allow` under the `-D warnings` gates on the published crates and the pm-quality-check matrix — your reproduction was right.

Now only the 34 vendored crates inherit the policy; the 21 first-party crates have no `[lints]` entry and run clippy's defaults, as on `main`. Verified with the exact gate commands: `cargo clippy -p qp-dilithium-crypto --no-deps -- -D warnings` and `-p quantus-miner-api` pass, and all 17 pm-quality-check crates pass `--all-targets --all-features -- -D warnings`. Getting those to pass also meant fixing the 9 first-party warnings the policy had been hiding and 23 lints in test targets that pm-quality-check was already failing on before this PR.

n13 (commented):
**Reviewer model:** Opus 5

Reviewed `1691a644` against `4c4a8f2f`. I ran the checks rather than taking the description on trust:

- `SKIP_WASM_BUILD=1 cargo clippy --locked --workspace` → **0 warnings**.
- `SKIP_WASM_BUILD=1 cargo doc --locked --workspace --no-deps` → **0 warnings**.
- `cargo clippy -p <crate> --all-targets --all-features -- -D warnings` for `pallet-reversible-transfers`, `pallet-wormhole`, `pallet-scheduler`, `pallet-multisig`, `pallet-mining-rewards`, `pallet-treasury`, `pallet-qpow`, `sc-network`, `sc-consensus-qpow`, `qp-dilithium-crypto` → **all clean**.

The claims hold. I found no correctness bug: every mechanical rewrite is semantics-preserving (list at the bottom). The findings are about two misplaced doc comments and about lint coverage the diff drops without saying so.

---

## 1. `PreValidatedProof` took over `ExitBundle`'s doc comment

`pallets/wormhole/src/lib.rs:514-525`

The new alias was spliced between the doc block and the struct that block was written for. The five lines starting "Normalized on-chain view of an exit proof, independent of aggregation depth" now document a 4-tuple alias, and `pub struct ExitBundle` at line 525 has no docs at all.

That paragraph states why per-segment nullifier attribution is sound: one private-batch proof is one segment, one public-batch proof is one segment per inner batch. It is the reason a double-spent nullifier invalidates only its own segment. It belongs on `ExitBundle`. Move the alias above line 514, or below the struct.

## 2. The weights floor comment now sits on the wrong assert

`pallets/wormhole/src/weights.rs:88-93`

One doc block covered both floors while they were one test. Doc comments bind to the next item only, so "The public-batch floor tracks the cheaper post-turnstile-removal pre-validation path (measured ~1.83B ps on 2026-08-08)" is now attached to the **private**-batch assert, and the public-batch assert is undocumented. Someone regenerating weights reads 1.83B next to a 550M threshold.

The conversion itself is an improvement. `#[cfg(test)]` starts at line 360, so both `const _` items are always compiled, including in the wasm runtime build. That is stronger than the `#[test]` it replaces.

## 3. `all = "allow"` silences `clippy::suspicious` and `clippy::perf` on 25 crates that were on defaults

`Cargo.toml:82`

Twenty-five crates had no `[lints]` table on `main` and ran full clippy defaults. They now get `all = "allow"` with only `complexity` and `correctness` raised back. `suspicious`, `style` and `perf` stay off. I confirmed with a two-line probe crate:

```rust
impl Sub for B { fn sub(self, r: B) -> B { B(self.0 + r.0) } }        // suspicious_arithmetic_impl
impl PartialEq for P { fn eq(&self, o: &P) -> bool { self == o } }    // unconditional_recursion
```

Both warn under clippy defaults. Both are silent under the new table.

These are not pristine upstream copies. `pallets/balances` carries "reject below-ED transfers that would reap the sender (#5)" and "drop force and migration extrinsics from the vendored pallet (#12)". `frame/support` carries "charge Quantus ML-DSA-87 overhead for base extrinsic weight (#14)". `pallets/referenda` carries "V12 audit: fix referenda, scheduler, and frame-system findings (#640)".

**Concrete suggestion.** I measured the cost of keeping `suspicious`. Adding one line to the workspace table:

```toml
suspicious = { level = "warn", priority = 1 }
```

produces exactly **31 warnings** workspace-wide, all in vendored code, all trivial: `frame-support` 13 (`needless_maybe_sized` ×7, `multiple_bound_locations` ×4, `empty_line_after_doc_comments`, `unnecessary_result_map_or_else`), `pallet-referenda` 15 (`multiple_bound_locations`), `sc-transaction-pool` 1, `sc-network-sync` 1, `sp-wasm-interface` 1. That is 31 warnings to keep `unconditional_recursion`, `suspicious_arithmetic_impl`, `suspicious_op_assign_impl`, `await_holding_lock` and `misrefactored_assign_op` on for the whole workspace. The 846 tabs-in-doc-comments are `style`, so this does not bring them back.

## 4. `pallets/frame-system` had a deliberately smaller table and loses correctness lints

`pallets/frame-system/Cargo.toml:69`

The PR body says four crates "had already pasted it into their own `Cargo.toml`". True for `client/executor-wasmtime`, `client/network-sync` and `client/wasm-interface`: I diffed those three against the new workspace table and they are byte-identical, a pure no-op. Not true for `pallets/frame-system`. Its table was only `all = "allow"`, `complexity = warn`, `correctness = warn`, with no priority-2 carve-outs.

Inheriting the full table newly allows three **correctness** lints there: `eq_op`, `erasing_op`, `while_immutable_condition`, plus roughly fifteen complexity lints. I verified with a probe that `eq_op` and `erasing_op` fire under the old three-line table and are deny-by-default under plain clippy. `pallets/frame-system` has 15 Quantus commits including "Harden FRAME primitives against audit findings (#616)" and "V12 audit: fix referenda, scheduler, and frame-system findings (#640)". Suggest leaving its original table in place.

## 5. The panic deny narrows from package-wide to lib-only, and it covered five crates, not three

`runtime/Cargo.toml`, `runtime/src/lib.rs:2`, `client/service/src/lib.rs:24`, `client/transaction-pool/src/lib.rs:24`

Cargo `[lints]` applies to every target in a package. An inner `#![deny(...)]` in `src/lib.rs` applies only to the lib. I verified with a probe: with `[lints.clippy] unwrap_used = "deny"` an integration test fails to compile; with the inner attribute it does not.

So these lose the deny: `runtime/tests/` (~113 `unwrap()`/`expect(`/`panic!(` sites, no `#![allow]` anywhere in there), `runtime/build.rs`, and `client/transaction-pool/benches/basics.rs`.

Separately, five packages had `[lints] workspace = true` on `main`, not three. `client/network` and `client/network-transactions` also had it and get no replacement. Their libs had blanket `#![allow(...)]` so the lib was already exempt, but the non-lib targets were not. The proof is still in the tree: `client/network/build.rs:19` carries `#![allow(clippy::unwrap_used)]`, which only ever mattered because the deny reached `build.rs`. That allow is now dead, and `client/network/benches/{notifications_protocol,request_response_protocol}.rs` are uncovered.

Production code is still covered and no CI job runs `--all-targets` on those crates today, so nothing breaks. But "Same rule, same crates" in the description is not accurate, and it is worth a line saying the deny is now lib-only.

## 6. `client/litep2p` loses its no-panic warn with nothing in its place

`client/litep2p/Cargo.toml:82`

The deleted comment said "remote input paths must not panic" and "Using warn instead of deny to allow gradual cleanup". That was a Quantus addition, not upstream noise, on the crate that parses attacker-supplied wire data. `unwrap_used`/`expect_used`/`panic` are restriction-group lints, so the new workspace table does not cover them at any level, and there is no `#![warn]` in `client/litep2p/src`. The cleanup it was tracking is not finished.

You asked in the description. My vote is keep it. `[lints] workspace = true` cannot carry extra lints, but one line in `client/litep2p/src/lib.rs` restores it using the exact mechanism this PR already uses for the three deny crates:

```rust
#![warn(clippy::panic, clippy::unwrap_used, clippy::expect_used)]
```

## 7. Nothing locks in the win, and two comments are now wrong

`.github/workflows/ci.yml:99-107` and `Cargo.toml:79-80`

`cargo clippy --locked --workspace` is at 0 on this branch and `cargo doc --locked --workspace --no-deps` is at 0. Neither CI step passes `-D warnings` or `RUSTDOCFLAGS`, so nothing stops the count climbing back on the next merge. Given the state this PR reaches, `-- -D warnings` on line 99 and `RUSTDOCFLAGS: "-D warnings"` on the doc step are now cheap, and without them this work has to be redone.

Two stale claims:

- `ci.yml:100-101` still reads "Workspace-wide deny is not feasible yet (~1.6k existing lints)". That is exactly what this PR fixed.
- The new `Cargo.toml:79-80` comment says first-party crates "are gated with `-D warnings` in CI". Only `qp-dilithium-crypto` and `quantus-miner-api` are gated automatically. `pm-quality-check.yml` is `on: workflow_dispatch` only, so its 17 crates are checked when someone runs it by hand. The split is still the right call, but the comment should not claim a gate that does not run on PRs.

## 8. A working doc link was dropped instead of fixed

`client/network/src/litep2p/shim/request_response/mod.rs:136`

The base link was `(crate::request_response::ProtocolConfig)`. It broke on a missing `s`: `client/network/src/lib.rs:265` declares `pub mod request_responses`, and line 275 inside it declares `pub type ProtocolConfig = RequestResponseConfig`. So `crate::request_responses::ProtocolConfig` resolves and the link could have been kept. As written, the page now points at `RequestResponseConfiguration`, a name that exists nowhere in the crate.

The other downgrades are all correct. I checked each: `NetworkWorker` has no definition in `client/network/src`; `DEFAULT_BACKOFF`, `CollectedFees` (`pub(super)`), `do_schedule_transfer_inner`, `Self::per_transfer_weight` and the `genesis_config_presets` items are private; `TreasuryPortion` is deleted; `pallet_zk_tree::INSERT_LEAF_*` is a glob. And `sp-io` → `sp_io` in `frame/support/src/traits/hooks.rs:543` is a genuine repair.

## 9. Minor

- **`node/src/command.rs:56`.** Reshaping `take_trimmed_secret` to return `Option` moves the "no secret provided on stdin" message away from the condition that produces it. The file's own convention is the other one: `read_secret_from_stdin` at line 73 and `generate_quantus_key` at line 90 both carry `#[allow(clippy::result_large_err)]`. One more of those would have left the function alone. This is the one security-sensitive file the PR touches, so I would rather not churn it.
- **`frame/examples/basic/Cargo.toml`.** That directory contains only a `Cargo.toml`, no `src/`. It is not in `[workspace] members` and nothing depends on it. Adding `[lints] workspace = true` there has no effect.
- **`patches/frame-storage-access-test-runtime/Cargo.toml`.** Quantus-authored (no Parity header, description "Local stub - skips substrate-wasm-builder..."), pulled in as a path dependency, and it now inherits `clippy::all = allow`. Small, but it shows the "first-party crates do not inherit this" invariant is enforced by a comment and nothing else.
- **Not metadata-neutral.** Appending `HighSecurityTxQuotaExceeded` to `pallet_reversible_transfers::Error` changes the pallet error type and therefore the runtime metadata and the `CheckMetadataHash` digest. It is appended last with no `#[codec(index)]` anywhere, so existing error indices are stable, and `runtime/src/lib.rs:72` puts the `spec_version` bump on the release workflow. Nothing to do. Worth knowing that a PR titled "clear clippy warnings" is not a no-op for the runtime.

---

## Checked and clean, so no need to re-litigate

- `filter_map` → `map` in `routing_table.rs:187` and `transport_service.rs:542`: every path in both original closures returned `Some`, so it is a strict no-op.
- All four `% == 0` → `is_multiple_of()` sites: no zero divisor is reachable. Divisors are literals (60, 24) or non-zero consts (`SCALE_DOWN_FACTOR = 10_000_000_000`, `AMOUNT_SCALE_DOWN_FACTOR` behind `leaf_quantum()`). In `credit_native` the new form is actually safer: `is_multiple_of(0)` returns false there instead of panicking mid-block.
- `.as_ref() == [0u8; 32]`: `BytesDigest: AsRef<[u8]>`, so this is `impl PartialEq<[U; N]> for &[T]`, length-checked and identical.
- `if let (Some(expr), None) = ...` in `call.rs:387` and `error.rs:82`: identical control flow, and the tuple-of-refs borrow ends before the value is consumed.
- `Sc(Box<sc_cli::KeySubcommand>)`: `clap_builder` has `impl<T: Subcommand> Subcommand for Box<T>`, pure forwarding, and `sc_cmd.run(&cli)` auto-derefs.
- `thread_local! { ... = const { ... } }`: `RefCell::new`, `BTreeSet::new` and `Weight::from_parts` are const fn. No observable difference.
- `status.enactment` without `.clone()`: `DispatchTime` is `Copy`, no partial move.
- `WAITING_PROOF` / `HASH_READY`: genuinely dead. The call sites had already been rewritten to `let ... else` and `if let`; only comment references remained.
- `use codec::Encode` moved into `mod tests`, `MaxEncodedLen` dropped from `transaction_extensions.rs`: zero remaining uses outside `#[cfg(test)]` in both.
- `Result<(), ()>` → `DispatchResult`: `DispatchResult` is in scope via `frame_support::pallet_prelude::*`, `try_mutate`'s error type is unconstrained, and the only production caller (`runtime/src/transaction_extensions.rs:104`) discards the error with `map_err(|_| ...)`. On-wire behaviour is unchanged.
- `check-cfg`: only `cfg(ignore_flaky_test)` is used anywhere in the workspace, and `sc-network` both inherits the allowlist and emits its own `cargo::rustc-check-cfg`. No gate breaks.

---

Nothing here blocks. The verified result is real: 1,442 clippy warnings and 38 rustdoc warnings to zero, with the published-crate and pm-quality-check gates passing, and the DRY fix of hoisting one shared table is the right shape. The two doc comments in `pallets/wormhole` are worth fixing before merge since they are one-line moves. Items 3 through 7 are judgement calls I would make differently, but they are yours to make, and item 7 in particular is worth doing while the number is still zero.

**Verdict (advisory): Approve**

n13 (commented):
Verdict (advisory): **APPROVE**

No blocking findings at `ab6dfe92`.

The earlier lint-gate defect is fixed: only the 34 vendored manifests inherit the permissive upstream policy, while first-party crates remain on Clippy defaults, and the current CI commands now hard-fail both Clippy and rustdoc warnings. I also traced every non-manifest edit; the iterator, `Option`, `Copy`, modulo, import, test, and quota-error rewrites preserve the existing behavior. `HighSecurityTxQuotaExceeded` is appended to the pallet error enum, so existing error indices remain stable.

Two non-blocking documentation nits remain:

- `pallets/wormhole/src/lib.rs:514-525`: the existing `ExitBundle` explanation now attaches to `PreValidatedProof`, leaving `ExitBundle` undocumented. Moving the alias above the block (or its own comment below the struct) would restore the intended association.
- `pallets/wormhole/src/weights.rs:88-92`: the public-batch explanation attaches to the private-batch compile-time assertion. Splitting the comments between the assertions would avoid pairing the ~1.83B measurement with the 550M private threshold.

Validation:

- Reviewed the complete `4c4a8f2f...ab6dfe92` diff and affected lint, runtime, node CLI, networking, and test paths.
- `git diff --check`, `cargo metadata --locked --no-deps --format-version 1`, `taplo format --check --config taplo.toml`, and `scripts/fmt.sh --all -- --check` passed.
- `SKIP_WASM_BUILD=1 cargo clippy --locked --workspace -- -D warnings` passed.
- `SKIP_WASM_BUILD=1 RUSTDOCFLAGS="-D warnings" cargo doc --locked --workspace --no-deps` passed.
- All 17 post-merge matrix crates passed `cargo clippy --locked --package <crate> --all-targets --all-features -- -D warnings`.
- Focused high-security quota tests (normal and zero-capacity rejection paths) and the node secret-trimming test passed.
- All live exact-head GitHub checks are green on Linux and macOS.

GitHub does not permit authors to approve their own PR, so this comment records my approval opinion.

PR #683, opened 2026-08-31
Bump qp-rusty-crystals dilithium/hdwallet to 4.1.1

n13:
Bumps `qp-rusty-crystals-dilithium` and `qp-rusty-crystals-hdwallet` from 4.1.0 to 4.1.1 (`Cargo.toml` + lockfile; no other crates touched).

dilithium 4.1.1 fixes a signature-forgery vulnerability: the malicious-key guard only rejected an all-zero `t1`, but any key whose coefficients keep `HighBits(c·2^D·t1)` at zero accepted forged signatures (`z = 0`, empty hint) for every message with no secret key. 4.1.1 rejects the whole degenerate family; honest keys are unaffected. hdwallet 4.1.1 is the companion release raising its dilithium floor to `^4.1.1`.

Cooldown-bypass-reason: dilithium 4.1.1 fixes a signature-forgery vulnerability (degenerate t1 public keys accepted forged signatures); first-party crates published by our own CI today, waiting 30 days would leave the vulnerable 4.1.0 pinned.

Note for the cooldown gate: apply the `dependency-cooldown-bypass` label (the label needs to be created in this repo first) — the check re-runs on labeling and reads the reason line above.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7380,18 +7380,18 @@ checksum = "5872607e25ea4ee5fb37e64bf1462168e1a36a4e719cdc8a105533c708253918"
 
 [[package]]
 name = "qp-rusty-crystals-dilithium"
-version = "4.1.0"
+version = "4.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2facf5349ce149731e6d21dc338f847c78e4a3a65e92e086e2cee244e7cf0457"
+checksum = "789877c169226a35d2ea686bbd9d506becc693f7bb0ee91acc03f74491e80c0f"
 dependencies = [
  "zeroize",
 ]
 
 [[package]]
 name = "qp-rusty-crystals-hdwallet"
-version = "4.1.0"
+version = "4.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


n13 (commented):
**Reviewer model:** GPT Sol

Verdict (advisory): Approve

No blocking findings.

The manifest and lockfile resolve both `qp-rusty-crystals-dilithium` and `qp-rusty-crystals-hdwallet` to 4.1.1. I inspected the published source delta: dilithium applies the shared degenerate-`t1` rejection predicate in verification, public-key parsing, and secret-key import, with regressions for ML-DSA-44/65/87; hdwallet 4.1.1 raises its dilithium requirement to 4.1.1. A targeted Cargo 1.93 update from the recorded base reproduced this PR's lockfile exactly.

Validation:

- `git diff --check` and locked Cargo metadata passed.
- Chain's `qp-dilithium-crypto` all-feature suite passed (23 tests), as did strict all-target Clippy.
- The exact upstream release workspace passed the dilithium all-feature suite (164 tests), the hdwallet suite (62 unit tests plus integration/zeroization tests), and strict all-target Clippy.
- All current PR checks pass: format, Linux and macOS check/test matrices, Clippy/docs, and the dependency-cooldown gate.

The unpacked crates.io dilithium package cannot run its own unit tests in isolation because test-only vectors outside the crate were not included in the package. This does not affect downstream builds; the corresponding exact release-workspace tests above cover the shipped source.

PR #704, opened 2026-09-14
Replace Unimplemented Revert Aux-Hook with No-Op and Handle `current_dir` Resolution Errors

mozluk:
### Description
This pull request resolves medium- and low-severity CLI panic vectors in `chain` identified during the Quantus workspace security audit (**FM-13, FM-14**).

Previously, invoking the `revert` subcommand triggered an `unimplemented!()` panic when executing consensus auxiliary data revert logic (leftover from Grandpa's removal)[cite: 49]. Additionally, resolving a relative path passed via `--node-key-file` relied on an unhandled `std::env::current_dir().unwrap()`, which panicked if the working directory was invalid or deleted[cite: 49].

#### 1. Safe No-Op Consensus Revert Hook (FM-13 - `node/src/command.rs`)
* **Eliminated `unimplemented!()` Panic:** Replaced the unimplemented Grandpa consensus revert closure with a deliberate no-op closure returning `Ok(())`: ```rust let aux_revert = Box::new( |_client, _, _blocks| -> Result<(), sc_cli::Error> { Ok(()) }, );

Upstream Type Alignment: Conforms strictly to upstream sc_cli::AuxRevertHandler (Box<dyn FnOnce(...) -> Result<(), sc_cli::Error>>), allowing chain reverts to complete cleanly without crashing the node process[cite: 49].

#### 2. Graceful Relative Node Key Resolution 
 (FM-14 - node/src/command.rs)Eliminated current_dir().unwrap(): Replaced .unwrap() with explicit error mapping via .map_err(...).  Descriptive Input Error: Maps working directory lookup failures to sc_cli::Error::Input with an informative error message instead of an unhandled panic[cite: 49, 61].
#### How to Review
Inspect node/src/command.rs under Some(Subcommand::Revert(cmd)) to verify the no-op aux_revert closure.  Inspect the relative path resolution branch for --node-key-file to verify the error mapping around std::env::current_dir().


node/src/command.rs
```diff
@@ -495,9 +495,12 @@ pub fn run() -> sc_cli::Result<()> {
 			runner.async_run(|config| {
 				let PartialComponents { client, task_manager, backend, .. } =
 					service::new_partial(&config)?;
-				let aux_revert = Box::new(|_client, _, _blocks| {
-					unimplemented!("TODO - g*randpa was removed.");
-				});
+				// Grandpa was removed from this chain, so there is no consensus
```
n13, line 498: General policy is to fail early vs hide bugs

I think it's better to throw, which is why we're throwing.

Why did you make this change?

n13: If anything, a nice error message would be better, we don't want to ignore this

node/src/command.rs
```diff
@@ -578,8 +581,18 @@ pub fn run() -> sc_cli::Result<()> {
 						if path.is_absolute() {
 							path.to_path_buf()
 						} else {
-							// This is a valid assumption because the node is run from the shell
-							std::env::current_dir().unwrap().join(path)
+							// This is a valid assumption because the node is run from the shell;
```
n13, line 584: This change is good

n13 (changes requested):
Question on the first item

Second item is good, and agree with that

PR #706, opened 2026-09-19
fix: cut on-chain fees 10x via FEE_SCALE 1/10

n13:
Set `FEE_SCALE_DEN` from 1 to 10. Weight, length, deposits, and the high-security cap all go through this dial.

Live check on `--dev`, 1 QTC Alice to Bob (ML-DSA-87):

| | Before | After |
|---|---|---|
| `payment_queryInfo` | 0.010166025 QTC | 0.0010166025 QTC |
| `TransactionFeePaid` | 0.009991025 QTC | 0.0009991025 QTC |

Both numbers are exactly 10x smaller.

This is a runtime change. `spec_version` is not bumped here; the release workflow does that when the upgrade ships.


n13 (commented):

Verdict (advisory): Approve

No blocking findings.

The 1/10 ratio is consistently propagated through the fractional weight/base-fee coefficient, length pricing, scaled governance and multisig charges, bootstrap funding, and the high-security inclusion-fee cap. Scaling the cap with the charged fees preserves its relative admission headroom, and the updated padded-destination assertion continues to exercise the intended pre-fee rejection boundary. Leaving `spec_version` unchanged also matches the repository's documented release-proposal workflow.

Non-blocking: several nearby comments still quote the 1/1 absolute amounts, including the fee examples in `runtime/src/configs/mod.rs:370-379,443-458,511-512`, the high-security examples at `runtime/src/configs/mod.rs:559-564`, the multisig values at `runtime/src/configs/mod.rs:681-683`, and the padding-test comment at `runtime/tests/transactions/high_security_tip.rs:257`. With this PR those actual amounts are one tenth of the quoted values. Please align those comments in a follow-up so runtime documentation does not misstate the deployed economics.

Validation:

- `git diff --check 2d2791d7158dc2a3bddb12986c80a4eee253556c...ef54730a7dc6736433b42309cb1b1c781c57a791` passed.
- `SKIP_WASM_BUILD=1 cargo test --locked -p quantus-runtime` passed: 77 unit tests and 62 integration tests passed, 1 intentionally ignored, 0 failed.
- Exact-head GitHub checks are all successful: format; Linux and macOS workspace check/test matrices; clippy/doc analysis; dependency cooldown.

illuzen (approved):

